Issue #6, opened 2013-07-24
Support Web API routing attributes


JamesNK:
It would be nice if refit supported Web API attributes (coming in the next version of Web API) so the same routing can be shared on the server and client.

I think the best way to implement this would be to support overriding the underlying model directly (e.g. like IContractResolver in Json.NET) and then having a seperate Refit.WebApi package.

onovotny:
Closing due to age. Please re-open if this is still desired, though I'm not sure how this would work.

PR #18, opened 2013-09-08
Ability to support different attributes for routing.

mhbuck:
Being able to support other attributes for the routing I can see being potentially valuable for sharing of an interface on the client and server. Worked with this a bit and reading issue #6.

I have created this pull request to show my current thinking and get some feedback if I am on the right track with this. I have had a bit of a look at the IContractResolver mentioned by @jamesnk and gave this a shot.

Some of the things I am just wondering about:
- Naming - Is RestMethodResolver appropriate (Naming is hard :tm:)
- Is there anything I should watch out for being portable?
- Should `GetInterfaceRestMethodInfo` have the interface type passed as the parameter or instead the individual MethodInfo objects? (More I think about it leaning towards passing the MethodInfo objects.
- How should the files should be split out what approach would you like taken? For now they are just bundled into one.

Any other thoughts on the approach?


onovotny:
closing this due to age

PR #83, opened 2014-12-06
Support for custom serializers

### LINKED ISSUE #75: Support BodySerializationType.Xml
Why only Json and Forms? Xml is still needed for many legacy interfaces.

bennor:
@paulcbetts This is just a first-pass at it so we can talk about the approach. 

I've added `BodySerializationMethod.Custom`, because it saved having to change the signature of `RestMethodInfo.BodyParameterInfo` (and it makes a few things easier later), but it's not _required_ -- `RestMethodInfo.findBodyParameter()` treats any parameter with a type in `RefitSettings.CustomSerializers` as a body parameter.

Keen to see if this is like what you'd imagined us doing?

One thing I've done a bit differently to what you suggested was that I've made `ITypeSerializer.SerializeAsHttpContent()` synchronous. Making it async pushes a massive amount of changes through the codebase (mostly in tests, to be fair), but do we actually need to do it for serializing the _request_ body?

At least the following are still to do:
- [ ] Make `ITypeSerializer.SerializeAsHttpContent()` async (maybe?)
- [ ] Merge in #80 or add `RefitSettings` to `RestMethodInfo` and deal with the conflicts later.
- [ ] Add a basic XML serializer as a reference
- [ ] Write tests

This will resolve #75.

### DIFF
--- a/Refit/Attributes.cs
+++ b/Refit/Attributes.cs
@@ -70,7 +70,7 @@ namespace Refit
     }
 
     public enum BodySerializationMethod {
-        Json, UrlEncoded
+        Json, UrlEncoded, Custom
     }
 
     [AttributeUsage(AttributeTargets.Parameter)]
--- a/Refit/RefitSettings.cs
+++ b/Refit/RefitSettings.cs
@@ -1,4 +1,8 @@
-﻿using System.Reflection;
+﻿using System;
+using System.Collections.Generic;
+using System.Net.Http;
+using System.Reflection;
+using System.Threading.Tasks;
 
 namespace Refit
 {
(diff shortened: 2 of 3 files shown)


ashutoshraina:
:+1: for this ...I need it right now for dealing with legacy Api's ...when can we see this completed and merged ?

CSchick:
+1 I don't know the code well enough to intelligently make alternate suggestions but the intent and premise of the changes are needed and the approach as explained seems well thought out and professional. :)

Cheesebaron:
https://github.com/paulcbetts/refit/blob/master/Refit/RequestBuilderImplementation.cs#L112
https://github.com/paulcbetts/refit/blob/master/Refit/RequestBuilderImplementation.cs#L345

It is basically these two places it is used. So what do you propose we do here instead?

bennor:
This PR is almost a year old. It will probably need a lot of work before we can use it. Might even be worth throwing out and starting again. Unfortunately, I don't have the time for that at the moment.

Would be amazing if one of you wanted to pick it up.

onovotny:
closing this due to age

PR #104, opened 2014-12-19
Allow HttpContent body/return types for extra customizability. 

### LINKED ISSUE #78: Multipart requests
Hey Paul,

I saw on the roadmap that there are plans to handle Multipart requests. What was your idea on how to implement it as I need it for file uploads and maybe I can help out with implementing it?

Alternatively would it make sense to allow passing in a parameter of `HttpContent` and in such a case assign it directly to the `Content` property of the `HttpRequestMessage`? This is sort of similar to how you handle Streams....? This will also allow for a lot of flexibility to allow users to structure the request exactly the way they want to.

Thanks

onovotny:
This should fix #78


Refit-Tests/InterfaceStubGenerator.cs
```diff
@@ -57,9 +57,9 @@ public void HasRefitHttpMethodAttributeSmokeTest()
             var fixture = new InterfaceStubGenerator();
 
             var input = file.GetRoot().DescendantNodes()
-                .OfType<InterfaceDeclarationSyntax>()
-                .SelectMany(i => i.Members.OfType<MethodDeclarationSyntax>())
-                .ToList();
+                            .OfType<InterfaceDeclarationSyntax>()
```
bennor: We're clearly still missing the R# setting that stops it from indenting chained method calls like this. Would be good to clean that up so we don't get unnecessary changes like this.

Refit-Tests/InterfaceStubGenerator.cs
```diff
@@ -128,13 +128,13 @@ public interface IAmARefitInterfaceButNobodyUsesMe
         [Get("whatever")]
         Task RefitMethod();
 
-        [Refit.GetAttribute("something-else")]
+        [Get("something-else")]
```
bennor: Formatting the document has broken this test case - it needs the `Refit.` because that's what's being tested.

Refit-Tests/InterfaceStubGenerator.cs
```diff
@@ -128,13 +128,13 @@ public interface IAmARefitInterfaceButNobodyUsesMe
...
         Task AnotherRefitMethod();
 
         [Get(ThisIsDumbButMightHappen.PeopleDoWeirdStuff)]
         Task NoConstantsAllowed();
 
-        [Get  ("spaces-shouldnt-break-me")]
+        [Get("spaces-shouldnt-break-me")]
```
bennor: Same deal here - the space after `Get` was what was being tested.

Refit-Tests/InterfaceStubGenerator.cs
bennor: Looking at this on my machine, are you sure your local settings aren't somehow overriding the project settings? These indents shouldn't be happening - if I do _Tools > Cleanup Code..._ then select _Default: Reformat Code_, these indents are reset to what they were before.

bennor:
Do you reckon this will make #83 is unnecessary? With a few small changes to this PR, we could support any type that derives from `HttpContent` instead of just `HttpContent` itself, then custom serialization of any type could be achieved with a class that derives from `HttpContent`.

Would mean we wouldn't have to sacrifice readability/usability on the interfaces, because it would allow something like this: 

```
public interface IWhateverApi
{
    [Get("/some-xml-endpoint")]
    XmlContent GetSomeXml();
}
```

paulcbetts:
@onovotny I think if you burn the "Fix formatting" commit, this might be salvageable

PR #115, opened 2015-01-15
adds support for PATCH method out of the box

cristi-badila:


paulcbetts:
:+1: Thanks @cristi-badila!

bennor:
@onovotny Rolled his implementation of this into #102, so we'll need to clean that up before it gets merged.

cristi-badila:
@paulcbetts glad to help, but unfortunately there is a small issue in my pull request (I initially set the method for the attribute to "Patch" and then updated it to "PATCH" but I forgot to update the corresponding specs which now fail since they expect to find the method "Patch"). Should I open another pull request for this? Sorry for the inconvenience.

paulcbetts:
@cristi-badila Yep, no worries on the extra PRs, they're free of charge!

PR #117, opened 2015-01-15
fixing the missing assembly version information for .NET 45 and WP8

moswald:


paulcbetts:
:+1:, thanks!

promontis:
This is not pushed to NuGet yet, right? Currently, I am getting an exception when using Refit in WPA8.1 saying it could not resolve version 2.2.0.0. Seems to be related to this issue... could you perhaps push this to NuGet?

PR #118, opened 2015-01-16
Debug.writeline

detroitpro:
No matter what I can't get Refit-Tests/RequestBuilder.cs to not change line endings, I only added 1 test.

Small change, added IsDebug to allow for outputting to the console/test runner.


onovotny:
closing this due to age. also, the best way to add logging is via a Delegating Message Handler.

PR #124, opened 2015-01-26
Always use compression if the client supports it

skendrot:
This will greatly reduce the size of the response making requests faster

### DIFF
--- a/Refit/RestService.cs
+++ b/Refit/RestService.cs
@@ -2,6 +2,7 @@ using System;
 using System.Net.Http;
 using System.Collections.Generic;
 using System.Linq;
+using System.Net;
 using System.Threading.Tasks;
 using System.Threading;
 
@@ -38,7 +39,11 @@ namespace Refit
 
         public static T For<T>(string hostUrl, RefitSettings settings)
         {
-            var client = new HttpClient() { BaseAddress = new Uri(hostUrl) };
+            var handler = new HttpClientHandler();
+            if (handler.SupportsAutomaticDecompression) {
+                handler.AutomaticDecompression = DecompressionMethods.Deflate | DecompressionMethods.GZip;
+            }
+            var client = new HttpClient(handler) { BaseAddress = new Uri(hostUrl) };
             return RestService.For<T>(client, settings);
         }
 


skendrot:
While this post is old, it does clearly state how to use automatic decompression
http://blogs.msdn.com/b/dotnet/archive/2013/07/17/httpclient-2-2-is-now-stable.aspx
> If an implementation of **HttpClient** doesn’t support automatic decompression, it returns **false** from the **SupportsAutomaticDecompression** property and throws a **NotSupportedException** from the setter and getter for the AutomaticDecompression property.

This was created as a way to "future proof" the use of decompression. If an HttpClient is written for some other device like an IoT device and they are unable to decompress for whatever reason then it will throw an exception.

bennor:
Pretty sure that's there so classes deriving from the base `HttpClientHandler` can override to show they don't support automatic decompression and you're just newing up one that always will, but I guess it could be implemented differently on some future platform.

No harm in keeping the check anyway. :smile:

mburbea:
I disagree. Please do not automatically set this property. Automatic decompression may do some undesirable behavior. I use refit to proxy some calls and stream the response. What actually happens is that when automatic compression is set, `HttpClient` will decompress the stream then send the response down. This was adding about a 10-20ms overhead before the response would be send to the client depending on the size of the payload and creating more garbage. If your eventual goal is to send the result to a browser that will decompress it this is just wasteful. 
I solved my need by doing the following:
```csharp
    [Headers("Accept-Encoding:gzip,deflate,sdch")]
    public interface IPassThroughRoute
```
I know this is not the norm admittedly, but I like that if I need the option I can just send a httphandler with the compression fields set.

skendrot:
I can see your point on that. I saw Refit as a client library rather than a server tier library. Of course it could be used in either case and I would not want decompressing just to compress it again either. There work be a workaround for this by sending your own HttpClientHandler into the overridden For method. And the same can be said for using compression... All client development should be requesting compression so maybe make a change so that each platform has a different implementation? On .NET it would not request zip, while client (WP/Store) would.

paulcbetts:
Sorry, I think that if users want compression, they can provide their own HttpClient, I don't want to build this into Refit by default, I'd rather have what Microsoft chooses to provide as defaults for HttpClient

PR #128, opened 2015-02-05
Moved ApiException into the Portable project

### LINKED ISSUE #126: Cannot access to ApiException class.
I want to catch an ApiException thrown, but I cannot access to the ApiException class to specify it in the catch clausure.

bennor:
This should allow projects that use the Refit library directly (i.e. other PCLs) to catch `ApiException`.

This is super lazy, but putting `ApiException` in `RestService.cs` means I don't have to change 230958 project files.

Fixes #126.

### DIFF
--- a/Refit/RequestBuilderImplementation.cs
+++ b/Refit/RequestBuilderImplementation.cs
@@ -505,62 +505,6 @@ namespace Refit
         }
     }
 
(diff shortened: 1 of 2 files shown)


paulcbetts:
Into It. :sparkle:

Issue #136, opened 2015-03-14
Content-Type header attribute on method doesn't apply correctly


smoy:
```
[Post("/store")]
[Headers("Authorization: Bearer", "Content-Type: application/json; charset=UTF-8")]
Task<string> InsertItem([Body] string content);
```

The above snippet doesn't seem to apply `application/json` to the content header.

I have to resort to the following work around

```
[Post("/store")]
[Headers("Authorization: Bearer")]
Task<string> InsertItem([Body] string content, [Header("Content-Type")] string contentType);
```

Now, it seems whatever I pass into contentType is apply to the content header.

bennor:
Will fix this today or tomorrow.

muhammetsahin:
Hi,

I've tried

	[Post ("/route/execute")]
		[Headers("Authorization: Bearer","Content-Type: application/json; charset=UTF-8")]
		Task<PropertyDto> GetProperty ([Body(BodySerializationMethod.Json)] Dictionary<string, object> data,string accessToken);

and tried this,

	[Post ("/route/execute")]
		[Headers("Content-Type: application/json; charset=UTF-8")]
		Task<PropertyDto> GetProperty ([Body(BodySerializationMethod.Json)] Dictionary<string, object> data,[Header("Authorization: Bearer")] string accessToken);

data value is:

var data = new Dictionary<string, object> {
				{"Controller","Property" }, 
				{"Action", "List"},
				{"Method", "GET"},
			};

WEB API:


Url
http://appapi.xxxx.com/api
Content-Type
application/json
Authorization
Bearer e7LFtHMK3Ru84CZWzy9OIOnQMQPm7DIanNEdI85sacGO29QR4jqmVz5oQ1ckaJmr39KS2S96i4bX5iYnjnbSRRNddT0noRHP_1UI8Aq6zT__H0m2Ip7wirMzDBKa_A7zCZztRnVfj9llipOWWU7s9MHzxh_mO3hmaUOZHGGPMWcqfx5QUvcNVNz6OjCgAp8lFBEucENBDHa3endey7la0tEmPCiCZdpMJN5ol75oALGR1tuhJ05eYT2KLRlgzR41TlQZ6J-sn-qJcXIq552P5g
Raw
{"Controller":"Property","Action":"List","Method":"GET"}

But I've got error:

500 Internal Server Error 

Note: access token is not  correct. it is sample accesstoken.

bennor:
[...]

Issue #121, opened 2015-01-21
Sending duplicate header


pierskarsenbarg:
I have two methods in my interface:

`[Get("/Session/GetId")]
        Task<ApiInt> GetProfileId([Header("X-SecureCode")] string secureCode, [Header("X-SessionId")] string sessionId);`

and 

`[Post("/Order")]
        Task<ApiOrder> CreateOrder([Body]ApiOrder order, [Header("X-SessionId")] string sessionId);`

that are called one after another (need the id to create the order).

However, I've noticed that when the second method is being run, the x-sessionid header is being sent over twice so when the API received it, it's coming as a comma separated list.

I've run this through fiddler and can see header coming over.

Is there something I should be doing in Refit to stop this happening?

bennor:
Sounds like a bug - can you create a small repro project (or add a test)?

bennor:
Also, totally unrelated but as of [2.2.0](https://github.com/paulcbetts/refit/releases/tag/2.2.0) you shouldn't need the `ApiInt` wrapper class anymore.

pierskarsenbarg:
Ok, I've spent some time putting together something... (here: http://github.com/pierskarsenbarg/RefitTest)

However, (and annoyingly so) I can't reproduce the duplicated header.

It's not like it's coming through as two headers, the value is reproduced (so in my example, X-Header: foobar becomes X-Header: foobar, foobar

I had a look to see if there was anywhere that we were setting default headers and the only one is an accept header (setting it to application/json).

RTodorov:
:+1: here. It only happens on my Android app. Exact same scenario, my session token header is being duplicated.

I'll dig into it and if I find anything relevant I'll update this post. I also have a couple of default headers (X-User-Agent, X-Locale) but it worked fine until... a couple of days ago! Suddenly it started duplicating the X-Token header!

RTodorov:
Guys,

The problem is here:

            var s = value.ToString();
            request.Headers.TryAddWithoutValidation(name, s);

            if (request.Content != null) {
                request.Content.Headers.TryAddWithoutValidation(name, s);
            }

The header is being added twice only on POST calls, because request.Content is not null. Then, before the request is made, HttpClient joins the two headers into a single one, concatenated by a comma.

I don't know what triggered it since it started a couple of days ago. Before that, everything worked just fine, including POST requests.

The other question is: do we really need to add the header on both request.Headers and request.Content.Headers?

RTodorov:
[...]

PR #139, opened 2015-03-27
Header fixes

bennor:
This PR contains the fixes for both of the outstanding header related bugs (#121 and #136).

* Adding of all headers is now deferred until the body parameter has been added to the request. This means `Content` is there and we can add headers to it.
* We now only try to add to the content headers collection if adding to the request headers collection failed.

Okay


RTodorov:
Cool! Thanks a lot @bennor !

adrianolc:
@bennor do you know when this fix will be released?

bennor:
@adrianolc & @RTodorov: this is up on NuGet now (in v2.3.0).

RTodorov:
Thanks a lot @bennor

PR #141, opened 2015-04-15
Uri template

darrelmiller:
This PR createsa pluggable URLTemplateHandler property on RefitSettings so that RFC 6570 URI Templates can be used with a replacement URLTemplateHandler.  I did not include that new handler in this PR so as to not introduce a dependency on Tavis.UriTemplates in refit.  
If this PR is accepted then I would release a new Nuget Package that implements a handler that looks like this https://gist.github.com/darrelmiller/deff3525791be6470520

### DIFF
--- a/Refit/PortableRefitSettings.cs
+++ b/Refit/PortableRefitSettings.cs
@@ -1,0 +1,48 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Reflection;
+using System.Text.RegularExpressions;
+using Newtonsoft.Json;
+using Newtonsoft.Json.Serialization;
+
+namespace Refit
+{
+    public class RefitSettings
+    {
+        public RefitSettings()
+        {
+        }
+        public JsonSerializerSettings JsonSerializerSettings { get; set; }
+        public IUrlParameterFormatter UrlParameterFormatter { get; set; }
+        public IUriTemplateHandler UriTemplateHandler { get; set; }
+    }
+
+    public interface IUriTemplateHandler
+    {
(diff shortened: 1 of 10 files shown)


paulcbetts:
Can you give an example of what this would look like with RFC 6570 URIs? I feel like this PR basically means if you enable this, all of the documentation is completely wrong, it's basically a different library.

darrelmiller:
The only breaking change that I am aware of is that URI Templates require query parameters to be specified, so you would need to do,

[Get("/users/list{?sort}")]

to be able to have sort as a query parameter.  The rest of the URI template capabilities are a superset of what your documentation covers.

It might be possible to get around the query string parameter issue.  My URITemplate library does have ways of generating the template based on parameters that are available.  Whether that is what people would want is a different matter.  This wasn't my idea in the first place ;-)

jzeferino:
@paulcbetts This PR makes a lot of sense since it allows the developer to change the way the URL template are resolve. Using this PR along with Tavis.UriTemplate we could support the URI Template Spec RFC6570 level 4. Here are some examples of what can be achieved with this PR integrated with Tavis.UriTemplate: https://github.com/uri-templates/uritemplate-test/blob/master/spec-examples.json

Please take a look at scenarios implementing level 4, from line 59.

@darrelmiller I would like to help with this PR, if @paulcbetts make it possible.

onovotny:
Closing due to age and changes since then.

jzeferino:
@onovotny is the current version of refit supporting the URI Template Spec RFC6570 level 4?

onovotny:
Refit has its own scheme, it's not specifically compliant with any RFC's.

jzeferino:
Thank you.

PR #146, opened 2015-05-17
First cut at multipart upload support

mteper:
Not yet well tested, but feedback is welcome. :smile:


mteper:
Now tested via Runscope (thanks, Paul!).

Largely working, except that it is not currently possible (at all) to specify file name when using Stream, and when using byte array, the name of the method parameter is used as file name.

```csharp
// for example,
interface ISample {
  Task Upload(byte[] bytes);
}

// will be uploaded as attachment named "bytes"
```

That's less than ideal. I am thinking a better approach would be introducing a FileNameAttribute on parameter that can be used to set the file name:
```csharp
interface ISample {
  Task Upload([FileName("Sample.pdf")] byte[] bytes);
}
```

@paulcbetts What do you think?

mteper:
Went with `AttachmentNameAttribute` instead.

mteper:
Done, tested, and documented. Let's :shipit: :)

Refit-Tests/MultipartTests.cs
```diff
@@ -0,0 +1,124 @@
...
+        [Post("/")]
+        Task<HttpResponseMessage> UploadString(string someString);
+
+        [Multipart]
+        [Post("/")]
+        Task<HttpResponseMessage> UploadFileInfo(FileInfo fileInfo);
+    }
+
+    [TestFixture]
+    public class MultipartTests
```
paulcbetts: Can you [Skip] these tests? I still want them in here because they're useful but having them all fail by default is :crying_cat_face:

Refit/RequestBuilderImplementation.cs
```diff
@@ -167,6 +188,51 @@ static void setHeader(HttpRequestMessage request, string name, string value)
...
+                    FileName = fileInfoValue.Name
+                };
+                multiPartContent.Add(fileContent);
+                return;
+            }
+
+            if (byteArrayValue != null) {
+                var fileContent = new ByteArrayContent(byteArrayValue);
+                fileContent.Headers.ContentDisposition = new ContentDispositionHeaderValue("attachment")
+                {
```
paulcbetts: :arrow_up:

Refit/RequestBuilderImplementation.cs
```diff
@@ -167,6 +188,51 @@ static void setHeader(HttpRequestMessage request, string name, string value)
...
+        {
+            var streamValue = itemValue as Stream;
+            var stringValue = itemValue as String;
+            var fileInfoValue = itemValue as FileInfo;
+            var byteArrayValue = itemValue as byte[];
+
+            if (streamValue != null) {
+                var streamContent = new StreamContent(streamValue);
+                streamContent.Headers.ContentDisposition = new ContentDispositionHeaderValue("attachment")
+                {
```
paulcbetts: :arrow_up:

Refit/RequestBuilderImplementation.cs
```diff
@@ -167,6 +188,51 @@ static void setHeader(HttpRequestMessage request, string name, string value)
...
+             
+            if (stringValue != null) {
+                multiPartContent.Add(new StringContent(stringValue), itemName);
+                return;
+            }
+
+            if (fileInfoValue != null) {
+                var fileContent = new StreamContent(fileInfoValue.OpenRead());
+                fileContent.Headers.ContentDisposition = new ContentDispositionHeaderValue("attachment")
+                {
```
paulcbetts: :arrow_up:

paulcbetts:
This is mostly :gem:

mteper:
@paulcbetts sorry, what do :arrow_up: indicate? :blush:

mteper:
Oh, is this about braces?

paulcbetts:
:+1: from me, any thoughts @bennor?

bennor:
Looks great to me too :+1:

mikecousins:
Is there any way to specify the filename when using this in a PCL? FileInfo doesn't work in PCL and the other methods (string, stream, bytes) can't use dynamic names can they?

mteper:
See `AttachmentName` attribute. FYI, I am rethinking that one a bit and considering changing it to Attachment with name and mimeType parameters.

mikecousins:
Sure, but that's static, not dynamic isn't it? My use case is sending an "email". I need a list of files, each with a user specified filename.

mteper:
Got it. Yes, that's not a use case I worked towards. Might need a special Refit collection/type for that. @paulcbetts any other ideas?

mikecousins:
Maybe something using https://github.com/dsplaisted/pclstorage and make it accept an IFile or IEnumerable<IFile>?

mteper:
Wouldn't really solve this universally, would it? Still no help for streams or byte arrays.

mikecousins:
Good point. Maybe `IEnumerable<KeyValuePair<string, Stream>>`, `IEnumerable<KeyValuePair<string, byte[]>>`, `IEnumerable<KeyValuePair<string, string>>`, `IEnumerable<KeyValuePair<string, FileInfo>>`?

mteper:
Not a fan of KVP as it's not extensible nor clear (what does Key mean, what does Value mean?).

I'd favor an `IEnumerable<IAttachment>` with concrete implementations for `StreamAttachment`, `ByteArrayAttachment`, and `FileInfoAttachment`.

The interface would be along these lines:
```csharp
interface IAttachment
{
  public string FileName { get; }
  public string MimeType { get; }
  public HttpContent GetContent();
}
```

mikecousins:
Yeah, that would be ideal for sure.

PR #152, opened 2015-06-05
Add nunit test adapter to VS so VSIX isn't required

onovotny:
Adding the NUnit VS Test Adapter nuget to the test project removes the need to have the NUnit VSIX pre-installed in VS to run unit tests


paulcbetts:
Into it!

PR #154, opened 2015-06-20
Use NetCore45 TFM instead of WinRT45

onovotny:
The WinRT45 TFM is obsolete and should not be used. NetCore45 is the correct replacement.


bennor:
:+1:

onovotny:
No code changes; it's the same binary just using the correct TFM in NuGet

onovotny:
This now includes a fix to write the RefitStubs.g.cs file into the \obj dir. This is better as generated files should not be checked in anyway. This should also eliminate read-only issues as the file isn't checked in; it's created on-the-fly.

onovotny:
Addresses #159

mburbea:
:+1:

Issue #165, opened 2015-07-22
Alternate packages location not supported


batkuip:
I have  a nuget.config which sets the repositoryPath to somewhere outside the solution folder.  Unfortunately this breaks refit as the refit.targets is expecting packages to be in the default location

`<PreBuildEvent>mono "$(SolutionDir)packages/refit.2.3.0/tools/InterfaceStubGenerator.exe" "$(ProjectDir)RefitStubs.cs" "@(Compile)"</PreBuildEvent>`

It works if I change it to:

`<PreBuildEvent>mono "$(MSBuildThisFileDirectory)../tools/InterfaceStubGenerator.exe" "$(ProjectDir)RefitStubs.cs" "@(Compile)"</PreBuildEvent>`

However not sure what best practice is here.

paulcbetts:
I'd take that PR

onovotny:
This is addressed in #154 and #153

batkuip:
@onovotny path in 359a3d7d7faed0a61a80d067d21aed613ec6d17b looks wrong too me.  If targets is in the build folder then tools is it's sibling folder.  Am I missing something?

onovotny:
The path is right... The targets cannot be directly in a \build folder as targets must always be in a folder with a TFM to work with NuGet v3.

The new .nuspec has the targets in `build\portable-net45+netcore45+wp8+wpa81+monoandroid+xamarin.ios10\`

batkuip:
@onovotny ah ok - that's for your changes.  This PR is just a fix for this specific issue on the current build.

onovotny:
Sure, but that still leaves refit broken for UWP projects. I'd like to fix it all :)

paulcbetts:
I'm not going to support the new NuGet v3, sorry

onovotny:
Can I ask why, it's just a path change?

PR #153, opened 2015-06-20
Combined PR that addresses #88 and #89 and #148 in one

onovotny:
I've merged the two into a single PR as I'm using it in an unofficial build for now....just passing it along.

PR #166, opened 2015-07-22
Support alternate nuget repositoryPath

batkuip:
PR for issue #165


onovotny:
Just as a head's up, this is already addressed in #153 and #154

paulcbetts:
Gonna use #153 and #154 instead, but thanks for having a look @batkuip!

Issue #167, opened 2015-07-29
Way to keep generated classes internal?


ncruces:
Hi,

I came across this when trying to use some internal datatypes in my interfaces.

For example:
```
interface IRemoteApi
{
    Task<User> GetUser(string id);
}
````

Here both `IRemoteApi` and `User` are internal. User implements a public `IUser` and I wrap all calls to the interface in methods responsible for caching, etc. So I'd rather not export it.

Doing this gets me a `CS0050: Inconsistent accessibility` in `RefitStubs.cs`. Changing the mustache to make the autogenerated classes internal fixes it.

Would it be possible (and make sense) to make the class internal whenever the interface is internal?

I tried to have a go into it, but I'm having difficulty understanding the CodeAnalysis stuff, so before I go any further, I'd like to have a feel on whether this makes sense (and a PR would be accepted).

bennor:
:+1:

ncruces:
#168 was merged, this can be closed.

PR #168, opened 2015-07-29
Allow AutoGenerated classes to be internal.

ncruces:
Implements feature discussed in #167.


bennor:
This looks good to me! :+1:

onovotny:
Just one question. What if someone wants the interface to be internal but expose the data types?

ncruces:
We have this:
```C#
internal interface IRemoteApi
{
    Task<User> GetUser(string id);
}
```

Refit generates this:
```C#
public partial class AutoGeneratedIIRemoteApi : IRemoteApi
{
    ... 
}
```

Which is typically used as:
```C#
var service = RestService.For<IRemoteApi>(url);
```

Since `IRemoteApi` is internal, what exactly is the point of the implementing class being public? Is it common to reference `AutoGeneratedIIRemoteApi` directly from another assemblies?

I may very well be missing some use case (the class is explicitly partial and open for extension, after all). But I'm failing to see the case where it makes sense for the interface to be internal and the class public, on the default configuration.

onovotny:
Oh, I misunderstood. The AutoGenerated* class should def be internal then. I was thinking about the parameter types, but those wouldn't be generated :)

ncruces:
Right, where this fails for me is that if the parameters/return types are internal (mine are) the class *can't* be made public.

But making those internal was my design choice. I have internal, mutable DTOs, which I implement public, readonly-view interfaces for outside consumption.

onovotny:
Yup, totally get it :+1:

ncruces:
Great! :+1: You can close #167.

PR #181, opened 2015-08-31
Allow reserved words in param names

### LINKED ISSUE #180: Reserved Words as Method Argument Names Cause Problems In Generated Code?
So I'm trying to use Refit against an API that has geo-coordinates in it. Unfortunately the API authors decided to shorten the argument names, so they are "lat" and "long" (ugh). "long" is of course a reserved word. It turns out you can declare a parameter with the name "long" by prefixing it with @ i.e

Task DoSomething(long lat, long @long) 
{
}

However the stub generated by Refit doesn't include the @ on the parameter name in the generated code in the stub file, so it fails to compile.

Is there anyway to tell refit that's a reserved word and it should use the @? I was kind of hoping the @ itself would be enough. I guess I can probably work around this with the AliasAs attribute, but it would nice if it 'just worked'.

bennor:
Fixes #180 by passing through arguments exactly as specified.


paulcbetts:
:+1:

PR #189, opened 2015-09-28
Fixed logic bug where it was possible for code to discover (by infere…

mteper:
…nce) a body component in a multipart request.

Since it doesn't make sense for a multipart request to contain a body, downstream logic broke down. The check is now more rigorous and a body will never be found for a multipart request.


paulcbetts:
:+1:

PR #190, opened 2015-09-28
Add URL and HTTP method to the ApiException

ammachado:
Adding these properties makes it easy to debug failed calls.

### DIFF
--- a/Refit/RestService.cs
+++ b/Refit/RestService.cs
@@ -74,6 +74,8 @@ namespace Refit
         public HttpStatusCode StatusCode { get; private set; }
         public string ReasonPhrase { get; private set; }
         public HttpResponseHeaders Headers { get; private set; }
+        public HttpMethod HttpMethod { get; private set; }
+        public string Url { get; private set; }
 
         public HttpContentHeaders ContentHeaders { get; private set; }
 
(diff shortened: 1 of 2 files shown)


Refit/RestService.cs
```diff
@@ -101,9 +105,9 @@ public T GetContentAs<T>()
                 default(T);
         }
 
-        public static async Task<ApiException> Create(HttpResponseMessage response, RefitSettings refitSettings = null)
+        public static async Task<ApiException> Create(string url, HttpMethod httpMethod, HttpResponseMessage response, RefitSettings refitSettings = null)
```
bennor: Change the signature here to take an `HttpRequest` instead of `url` and `httpMethod`, then any future additions from the request won't have to touch the calling code.

Refit/RestService.cs
```diff
@@ -85,9 +87,11 @@ public bool HasContent
         }
         public RefitSettings RefitSettings { get; set; }
 
-        ApiException(HttpStatusCode statusCode, string reasonPhrase, HttpResponseHeaders headers, RefitSettings refitSettings = null) :
+        ApiException(string url, HttpMethod httpMethod, HttpStatusCode statusCode, string reasonPhrase, HttpResponseHeaders headers, RefitSettings refitSettings = null) :
```
bennor: But keep this as it is.

Refit/RestService.cs
```diff
@@ -74,6 +74,8 @@ public class ApiException : Exception
         public HttpStatusCode StatusCode { get; private set; }
         public string ReasonPhrase { get; private set; }
         public HttpResponseHeaders Headers { get; private set; }
+        public HttpMethod HttpMethod { get; private set; }
+        public string Url { get; private set; }
```
bennor: Change this to 

```csharp
public Uri Uri { get; private set; }
```

paulcbetts:
:+1:

Issue #200, opened 2015-11-04
Path delimiters are allowed in URL path parameter values


mteper:
Let's say I have an API interface defined thusly:

``` cs
public interface IMyApi
{
  [Get("/api/entity/{id}/foo")]
  Task GetEntityFoo(string id);
}
```

The following test will pass:

``` cs
public void TestFooThatCausesBadPath()
{
  var fixture = new RequestBuilderImplementation(typeof(IDummyHttpApi));
  var factory = fixture.BuildRequestFactoryForMethod("FetchSomeStuffWithVoidAndQueryAlias");
  var output = factory(new object[] { "6/6" });

  var uri = new Uri(new Uri("http://api"), output.RequestUri);
  Assert.AreEqual("/api/6/6", uri.PathAndQuery);
}
```

Shouldn't the path delimiter character in the middle of the parameter get URL encoded?

Before I submit a fix, I want to make sure I am not missing something.

mteper:
Added PR with a new setting, two out of the box implementations, and unit tests.
//cc @bennor @paulcbetts

MKuckert:
I think `PassthroughPathSegmentTransformer` should be the default setting in terms of backwards compatibility.

PR #201, opened 2015-11-04
Added IUrlPathSegmentTransformer and two implementations, passthrough…

mteper:
…, and encoding to address #200.


paulcbetts:
:+1:

Issue #160, opened 2015-07-15
Setting multipart stream field name


mikeyoon:
I'm trying out the new multipart feature, and I noticed there's no way to set the a file's field name or disposition type. The following change will probably work in my case when using a FileInfo:

``` C#
var fileInfoValue = itemValue as FileInfo;
if (fileInfoValue != null) {
    var fileContent = new StreamContent(fileInfoValue.OpenRead());
    fileContent.Headers.ContentDisposition = new ContentDispositionHeaderValue("form-data") { //changed disposition type to form-data
        FileName = fileInfoValue.Name,
        Name = itemName //Added itemName here
    };
    multiPartContent.Add(fileContent); 
    return;
}
```

But that does leave a gap for the Stream or byte[] and there's currently no attribute for setting the disposition type. I think AliasAs could be obeyed for multipart files to solve the name issue, but maybe the AttachmentName attribute needs to be expanded to support disposition type. I can add that and make a PR if that sounds good, but I'm open to other ideas.

TheTekton:
That would work for where multipart w/Refit isn't working for me, I think. (API expects name="image" in this particular case.) Was thinking that might be a good PR too.

TheTekton:
Okay, so while I'm not sure if what I've done locally to work around this is the best way, but made a gist to explain it nevertheless (checked out 2.3.0 as couldn't get latest commit to build locally [and had to disable monodroid due to System.Object reference error in 2.3.0 and latest Xamarin Studio stuff]):

https://gist.github.com/TheTekton/0eebf4600743564d7ba3

Created a Name attribute (wasn't sure of what to call it more specifically, in a hurry), then built it out like the AttachmentName attribute. When the Name attribute is empty, it's not used. Only testing byte[], here, but can't think of any reason why it might break the other supported types. Anyways, perhaps something for more discussion, but deferring PR for more feedback as well (even if what I have works, doubt the naming of things is exactly kosher).

kerr23:
I'm wondering if this is similar to my issue.
When trying to post to a service and my post looks like:
```POST http://places/service/v1/upload/ HTTP/1.1
Content-Type: multipart/form-data; boundary="----MyGreatBoundary"
Content-Length: 48636
Expect: 100-continue
Connection: Keep-Alive

------MyGreatBoundary
Content-Disposition: attachment; filename=stream
<binary data>
------MyGreatBoundary--```

What my service is expecting, and what I think is reasonable here is something more like:
(In the body)
```
[...]

PR #214, opened 2016-01-20
Modified addMultipartItem to use Content-Disposition "form-data" by default (#160)

adrienjeanjean:
Also mapped the itemName to the part name and the attachmentName attribute value to filename.
Added automatic attempt to serialize to JSON when itemValue is not stream/bytearray/string/fileInfo before throwing exception


adrienjeanjean:
I'm pulling my pull request back as I realise this is going to break previous implementations relying on the default Content-Disposition to be attachment. I'll have a go at another solution. Sorry for the inconvenience.

Issue #161, opened 2015-07-16
Allow sending custom classes in multipart


tamirdresher:
Currently you can only send String, Stream, FileInfo, and Byte array in a multipart request.
it would be useful to allow other custom classes to be sent by converting them to JSON

i think that all it takes is to add the following code at the end of the RequestBuilderImplementation.addMultipartItem(...) method

``` csharp
try 
{
    var stringContent = new StringContent(JsonConvert.SerializeObject(itemValue, settings.JsonSerializerSettings), Encoding.UTF8, "application/json");
    multiPartContent.Add(stringContent, itemName);
    return;
} 
catch (Exception ex)
{
    throw new ArgumentException(string.Format("Unexpected parameter type in a Multipart request. Parameter {0} is of type {1}, whereas allowed types are String, Stream, FileInfo, and Byte array or a type that can be converted to json", itemName, itemValue.GetType().Name), "itemValue",ex);

}
```

what do you think? i can create a pull request if it sounds good

deepforest:
Totally agree.
+1

egitlits:
It definitely useful!

yarons74:
Looks like a good solution

onovotny:
Fixed in #426. Please file a new issue if there's still things missing.

ianrathbone:
Hi guys thanks for making this available!
I've got stuck with this, just created a class with a stream and a string:
```
public class Upload
    {
        public Stream media { get; set; }
        public Configuration configuration { get; set; }
    }
```

Then using it on the interface

```
[Multipart]
[Post("/media")]
Task<Media> UploadMediaWithConfiguration([Header("Authorization")] string authorization,
            Upload upload);
```

But I get the error:

> {"Unexpected parameter type in a Multipart request. Parameter upload is of type Upload, whereas allowed types are String, Stream, FileInfo, Byte array and anything that's JSON serializable\r\nParameter name: itemValue"}

onovotny:
@ianrathbone that's not going to work like that since each parameter needs to be its own part in the overall content. You'd need to split those into separate parameters. The POCO case turns into a JSON string payload.

onovotny:
One other suggestion -- you can use the AuthorizationTokenGetter func on the Refit settings to plug in a token cache/value, etc. so you don't need to pass it on every api call. Just add `[Headers("Authorization: Bearer")]` as an attribute to the methods where you want it, and it'll call your func to get the value.

ianrathbone:
[...]

PR #223, opened 2016-03-15
Enhance and fix multipart data

erisco:
- Quote multipart `name` and `filename` values.
- Include `name` with `filename`.
- Allow any JSON-serializable object to be used.

Expecting this PR to need some cleanup.


jamsea:
This looks like it fixes issue #161. @paulcbetts can you provide feedback on this PR? We're currently using this in production. It'd be great to have this published as an official nuget package so we don't have to point to our fork

onovotny:
Can you please rebase this off of the latest master?

Also, the name parameter should not refer to the item name, but rather the parameter name. 

Thanks!

onovotny:
Closing this since it's old. Looking at the code, refit is currently setting the name and filename items now. As for quoting the values, can you share a repro where the existing logic isn't working right? I looked at how `ContentDispositionHeaderValue` and it appears to be quoting the values.

I'll address serializing arbitrary objects separately.

Thanks!

Issue #235, opened 2016-05-22
NETStandard / .NET Core support


onovotny:
Currently Refit does not work with .net core projects. With .NET Core moving to use msbuild, we should still be able to invoke the targets in the same way.

The easiest course of action is to just expose refit as netstandard and rely on mono to exec the tool as per today. The next option is to make the generator tool a netcore app itself or a dotnet command (dotnet-refit) and enable a pre-build script to invoke it.

onovotny:
Workaround for now:

VS 2015 Update 2 adds support for changing a regular CSProj PCL to support .NET Standard, so create a csproj pcl and set it to .net standard 1.4

then use this project.json for it:
```json
{
  "dependencies": {
    "NETStandard.Library": "1.5.0-rc2-24027",
    "refit": "2.4.1",
    "Newtonsoft.Json": "8.0.3"
  },
  "frameworks": {
    "netstandard1.4": {
      "imports": ["dotnet", "portable-net45+win8"]
    }
  }
}
```

the imports are needed because refit doesn't explicitly support netstandard yet

Because it's a csproj, the targets work as expected. You can then add a reference to it from your asp.net 1.0 website or from another xproj like this:

```json
"MyPclLibrary": {
      "target": "project"
    }
```

PR #236, opened 2016-05-23
.NET Standard support

onovotny:
Addresses #235

This PR adds a `netstandard1.3` impl but relies on the existing targets to run. A future PR/feature would enable the interface generator to be a net core app


bilal-fazlani:
Waiting for this library to be Dotnet core compatible

PR #238, opened 2016-06-03
Added comment to prevent code analysis

mariusGundersen:
This will prevent the StyleCopAnalyzer from analyzing the generated file


mariusGundersen:
BTW, the mustache file in the tools folder of the nuget isn't actually read by the exe, since it's marked as `<EmbeddedResource>` in the csproj file, so it's actually inlined into the exe (open the exe file in a text editor and you will find the mustache template). This makes the comment about editing the mustache template in the generated file not true.

accidentaldeveloper:
@mariusGundersen Thanks for clarifying about the mustache file. I had tried to edit it as well and couldn't understand why my changes weren't reflected.

onovotny:
Done in #284

PR #241, opened 2016-06-08
Include <auto-generated/> header in mustache template

accidentaldeveloper:
As a general rule, it's unnecessary for automated code analysis tools to analyze generated files. For this reason some of these tools, including StyleCop, are configured to ignore files with an `<auto-generated/>` file header. It's impossible for a library to generate files that match the style guidelines for every project and this header will avoid errors resulting from style discrepancies.


csciallano:
love it :)

onovotny:
Done in #284

PR #242, opened 2016-06-14
Allow setting of the content type and field name for multipart fields

panetta-net-au:
Allows setting of the content type for multipart fields

Allow setting of the file name separate to the field name for multipart fields. The parameter name or AliasAs attribute now define the field name. This has the potential to break compatibility with applications which depend on the AttachmentName to set the field name. The alternative is to ignore the parameter name and only rely on AliasAs however this would result in inconsistent API design

Relates to: Setting multipart stream field name #160

### DIFF
--- a/README.md
+++ b/README.md
@@ -411,7 +411,7 @@ At this time, multipart methods support the following parameter types:
 
 The parameter name will be used as the name of the field in the multipart data. This can be overridden with the `AliasAs` attribute.
 
-For byte array, Stream and string parameters, use `AttachmentName` parameter attribute to specify the
+For byte array and Stream parameters, use `AttachmentName` parameter attribute to specify the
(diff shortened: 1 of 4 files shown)


panetta-net-au:
Ah, there's a pretty bad bug. I'll fix it first

panetta-net-au:
@paulcbetts I realised I tried to merge this to master instead of a feature branch.
Is it worth me fixing this and resolving conflicts or should I just close the pull request?

onovotny:
@panetta-net-au can you please rebase this on dev?

panetta-net-au:
Sure.

panetta-net-au:
New pull request open here
https://github.com/paulcbetts/refit/pull/288

PR #85, opened 2014-12-10
Add overload support by mangling method names

nekresh:
Refit now support method overloading.
The methods dictionary is built using mangled method name instead of using the method name only.

I'm not quite sure about the Mangle method position (currently in RequestBuilder type).

Closes #82.

Issue #82, opened 2014-12-05
Overloaded method support


nekresh:
As of now, refit doesn't handle method overload.
The methods dictionary can't be built because of DuplicateKeyException.

Would it be interesting to support WebApi's style method overload ?

``` C#
IEnumerable<X> Get();
X Get(int id);
```

PR #247, opened 2016-07-05
Support for dynamic query parameters and method overloads

jlucansky:


bennor:
A couple of things: 

1. Can you give an example of the proposed dynamic query parameters in action, or better, add some docs to the readme for it.
2. Why use an attribute for distinguishing overloads when you could just use the parameter types to determine uniqueness. Seems like an extra layer of complexity we could do without. Ideally overloads would _just work_.

onovotny:
Can you please rebase and ensure it builds correctly? Some of the added files seem to be missing from some of the projects.

Thanks!

Refit/RequestBuilderImplementation.cs
```diff
@@ -446,6 +463,13 @@ public RestMethodInfo(Type targetInterface, MethodInfo methodInfo, RefitSettings
             Name = methodInfo.Name;
             MethodInfo = methodInfo;
 
+            var uma = methodInfo.GetCustomAttributes(true)
+                .OfType<UniqueNameAttribute>()
+                .FirstOrDefault();
+
+            if (uma != null)
+                Name = uma.MethodName;
```
ahmedalejo: `UniqueNameAttribute` should not be compulsory for supporting overloads
name should default to method mangling #85 or [#82 Comment](https://github.com/paulcbetts/refit/issues/82#issuecomment-66140879)

jlucansky: Yes you are right and I wish there was another way, but it is almost impossible to get the same method signature in runtime as during stub generation.

onovotny:
Closing due to age. Refit also now supports method overloads

bennor:
Both of these features exist. Overloads should _just work_. Dynamic query parameters are supported via the `[Query]` attribute.

Issue #267, opened 2016-09-23
Multi-part form post to controller in asp.net core fails


mstrong64:
This is a great library so thanks very much for all the effort! 

I've just tracked down an issue whereby posting an image to an MVC endpoint fails after upgrading to asp.net core 1.0 (running Kestrel). I switched to using the Refit Multipart attribute but the post still wouldn't work. The IFormFile in the following controller was always null:

```
[Route("api/uploadImage")]
[HttpPost]
public async Task<SomeResponse> UploadImage(IFormFile image)

```

I eventually got this working with the following code using HttpClient:

```
async Task UploadImage(string fileName, string token)
{ 
  var documentToSend = System.IO.File.ReadAllBytes(fileName);
  var client = new HttpClient(new NativeMessageHandler());
  using (var multipartFormDataContent = new MultipartFormDataContent())
  {
    multipartFormDataContent.Add(new ByteArrayContent(documentToSend),
      '"' + "image" + '"',
      '"' + "image.jpg" + '"');
    client.DefaultRequestHeaders.Add("Authorization", "Bearer " + token);
    var result = await client.PostAsync($"{Constants.BaseApiUrl}/uploadImage", multipartFormDataContent);
  }
}

```
Note that I am using Kestrel and this may not be an issue with IIS, but the problem seems to be the `content-disposition` field for the file in post.

Refit version (doesn't work):

`content-disposition: attachment;filename=image.jpg`

HttpClient version (works as expected):

`content-disposition: form-data;name="profile";filename="image.jpg";filename*=utf-8''%22image.jpg%22`

Note also that the `name` attribute must match the parameter name in the controller action!

onovotny:
Yeah, definitely a bug here that I also discovered recently. I was able to work around it and still use Refit here:
https://github.com/onovotny/SignService/blob/master/src/SignClient/Program.cs#L120-L133

But this should be fixed internally.

onovotny:
Fixed

PR #274, opened 2016-10-03
Multipart fix

onovotny:
This fixes several multipart request issues in #267


onovotny:
@bennor fixed directly in master. GitLink was throwing with netstandard libs for some reason

PR #284, opened 2016-12-22
Add auto-generated attribute and extend pragma for stubs

rangp:
In my setup StyleCop Analyzers are having issues with the generated refit stubs. The changes below solve my problem. I think they won't cause any harm, since the file is auto-generated.


bennor:
Looks good to me 👍

onovotny:
Thanks!

PR #288, opened 2017-01-17
 Allow setting of the content type and field name for multipart fields

panetta-net-au:
Rebase and branch-change for https://github.com/paulcbetts/refit/pull/242

Allows setting of the content type for multipart fields

Allow setting of the file name separate to the field name for multipart fields. The parameter name or AliasAs attribute now define the field name. This has the potential to break compatibility with applications which depend on the AttachmentName to set the field name. The alternative is to ignore the parameter name and only rely on AliasAs however this would result in inconsistent API design

Relates to: Setting multipart stream field name #160


panetta-net-au:
@onovotny Anything you'd like me to change/update?

onovotny:
Haven't had a chance to review yet as I was travelling all week. I should have time in the next few days.

panetta-net-au:
Okay, thanks.

onovotny:
One question/concern about having the content type set as an attribute -- that makes it impossible to specify at runtime. In the examples, including the unit tests (thanks!), the content type had to be determined at compile-time.

While it's unreasonable to expect a file-extension->mime type to work in Refit, I do think that the mechanism should support setting the value at runtime.

In many ways, it feels like we're trying to set too much Metadata in parameters of a Multipart method. Can you consider using some sort of new "special" type, perhaps "PartInfo" that has all of the relevant data on it as members? Then, you could have either a single or sequence of PartInfo's as input to a Multipart method.

panetta-net-au:
Yeah, I was struggling with that while writing the tests. From memory, it's the same problem for the filename extension.

Are you suggesting that the method signature include a PartInfo parameter? Eg:

Task<HttpResponseMessage> UploadStream([PartInfo("streamInfo")] Stream stream, PartInfo streamInfo = new PartInfo() { AttachmentName = "test.pdf", ContentType = "application/pdf"} );

Or are you suggesting a PartInfo class that contains the stream + information. Eg:
Task<HttpResponseMessage> UploadStream(StreamPart stream);

and called like
UploadStream(new StreamPart(stream, contentType:"application/json"))

onovotny:
I'm suggesting the second option -- that you have a PartInfo and a StreamPart, FileInfoPart, etc that contain all of the data that was in the attributes. I'm thinking that might make it easier to call?

panetta-net-au:
Thanks, I'll take look at doing that this week.
Is this something we'd want in addition to the attributes to allow for simpler/static implementations like with the filename?

panetta-net-au:
@onovotny I've added the wrapper as you requested.

A couple of thoughts:

- We could get away with using System.Net.Http.ByteArrayContent and StreamContent instead of using the wrappers if we didn't want to support dynamic file names. I think having dynamic file names are worth the wrapper though (otherwise what's the point of dynamic content type support)
- Should I deprecate the existing AttachmentNameAttribute and remove the newly added AttachmentContentTypeAttribute? They provide a more refit-esque experience but they duplicate functionality. Keen to hear your thoughts.

I haven't documented yet - will add documentation when the implementation is nailed down.

Refit/MultipartItem.cs
```diff
@@ -0,0 +1,105 @@
...
+    public abstract class MultipartItem
+    {
+        public MultipartItem()
+        {
+
+        }
+
+        public MultipartItem(string fileName, string contentType)
+        {
+            this.FileName = fileName;
```
onovotny: should be null or empty checking

Refit/MultipartItem.cs
```diff
@@ -0,0 +1,105 @@
...
+            this.ContentType = contentType;
```
onovotny: Same here, null or empty checking.

Refit/MultipartItem.cs
```diff
@@ -0,0 +1,105 @@
...
+        }
+
+        public virtual string FileName { get; set; }
```
onovotny: FileName and ContentType don't seem to have a reason to be `virtual`, is there?

Refit/MultipartItem.cs
```diff
@@ -0,0 +1,105 @@
...
+        public virtual string ContentType { get; set; }
+
+        public virtual HttpContent ToContent()
+        {
+            var content = CreateContent();
+            if (!string.IsNullOrEmpty(ContentType))
```
onovotny: Seems like ContentType should always be required? if not, I'm ok with this and skipping the null or empty checks above

Refit/MultipartItem.cs
```diff
@@ -0,0 +1,105 @@
...
+        public MultipartItem(string fileName, string contentType)
+        {
+            this.FileName = fileName;
+            this.ContentType = contentType;
+        }
+
+        public virtual string FileName { get; set; }
+        public virtual string ContentType { get; set; }
+
+        public virtual HttpContent ToContent()
```
onovotny: Is there any reason for this to be virtual?

Refit/MultipartItem.cs
```diff
@@ -0,0 +1,105 @@
...
+
+        protected abstract HttpContent CreateContent();
+    }
+
+    public class StreamPart : MultipartItem
+    {
+        public StreamPart(Stream value, string fileName, string contentType) :
+            base(fileName, contentType)
+        {
+            Value = value;
```
onovotny: Check value for null

Refit/MultipartItem.cs
```diff
@@ -0,0 +1,105 @@
...
+        }
+
+        public StreamPart(Stream value)
+        {
+            Value = value;
+        }
+
+        public Stream Value { get; set; }
```
onovotny: only a public get... set should be private

Refit/MultipartItem.cs
onovotny: Set should be private for FileName and ContentType

Refit/MultipartItem.cs
```diff
@@ -0,0 +1,105 @@
...
+            return new StreamContent(Value);
+        }
+    }
+
+    public class ByteArrayPart : MultipartItem
+    {
+        public ByteArrayPart(byte[] value, string fileName, string contentType) :
+            base(fileName, contentType)
+        {
+            Value = value;
```
onovotny: Check for null

Refit/MultipartItem.cs
```diff
@@ -0,0 +1,105 @@
...
+        }
+
+        public ByteArrayPart(byte[] value)
+        {
+            Value = value;
+        }
```
onovotny: Check for null

Refit/MultipartItem.cs
```diff
@@ -0,0 +1,105 @@
...
+
+        public byte[] Value;
```
onovotny: Property with private setter.

Refit/MultipartItem.cs
```diff
@@ -0,0 +1,105 @@
...
+        }
+    }
+
+#if !NETFX_CORE
+    public class FileInfoPart : MultipartItem
+    {
+        public FileInfoPart(FileInfo value, string fileName, string contentType) :
+            base(fileName, contentType)
+        {
+            Value = value;
```
onovotny: Check for null

Refit/MultipartItem.cs
```diff
@@ -0,0 +1,105 @@
...
+        }
+
+        public FileInfoPart(FileInfo value)
+        {
+            Value = value;
```
onovotny: Check for null

Refit/MultipartItem.cs
```diff
@@ -0,0 +1,105 @@
...
+        }
+
+        public FileInfo Value { get; set; }
```
onovotny: private setter

Refit/Attributes.cs
```diff
@@ -148,4 +148,14 @@ public AuthorizeAttribute(string scheme = "Bearer")
         {
         }
     }
+
+    [AttributeUsage(AttributeTargets.Parameter)]
+    public class AttachmentContentTypeAttribute : Attribute
```
onovotny: Nuke this attribute and lets just use the wrapper

onovotny:
@panetta-net-au put a few comments in the review, small things but otherwise looks good. I agree that the ContentTypeAttribute should be nuked and I'm good with deprecating the AttachmentNameAttribute with a message that points people to the wrapper. Just make sure it's a soft deprecate not a hard fail one.

Refit/MultipartItem.cs
panetta-net-au: Given that ContentType wasn't required previously, I imagine there are web services out there that don't require it.

Refit/MultipartItem.cs
panetta-net-au: I might've been a little too "open" with my API design. I can remove the virtuals and make public sets private

Refit/MultipartItem.cs
panetta-net-au: FileName is technically not required as it can fallback to the parameter name (or the to-be deprecated attribute)

panetta-net-au:
@onovotny I think I've actioned all requested changes with the exception of the FileName/ContentType null checks. 

I also made the FileInfoPart filename fall back to FileInfo.Name before falling back to the parameter name to be consistent with the existing behaviour.

Refit/MultipartItem.cs
onovotny: I think that for the purpose of using the MultiPartItem special type, it's probably better to have FileName required. The parameter name inference is for "low friction" things, but since there's a wrapper type already involved, it may as well be explicit.

panetta-net-au: Okay, would you like me to remove the fallback for FileInfoPart's file name to FileInfo.Name as well? I think that would be cleaner.

panetta-net-au:
@onovotny Any further changes? I might have an hour today to do them if needed - otherwise can find some time another day.

panetta-net-au:
Hi @onovotny Just following up on this since it's been a little while. Anything further you need from me on it?

onovotny:
hey, sorry, got swamped... will review soon and merge.

PR #304, opened 2017-04-05
Allow post-processing of HttpResponseMessages

enscope:
Sometimes, it is beneficiary to be able to process the `HttpResponseMessage` even when the return type of API call defines more specific class, e.g. because an API you are interfacing with is defining some important run-time state in response header and there was no other way to access such information than by explicitly returning `HttpResponseMessage` and then taking care of `DeserializeObject` manually. 

That brings two nasty issues:
- the interface is not as clear, as it could be - the result value is replaced by `HttpResponseMessage`
- it is error prone and more complicated to do so 

My proposed solution introduces `IHttpResponseMessagePostProcessor` with only a single method void `PostProcessHttpResponseMessage(HttpResponseMessage message)`, that in style of Spring/NET `IContextAware` allows implementing entity to work with the message. There is no cascading to related entities, only a deserialized object is checked, so the performance is not much affected and it is called post-processor to avoid clients thinking they should store the message in their structure somewhere.

I hope you will not consider it edge case and will merge it to the main branch.

### DIFF
--- a/Refit/RequestBuilderImplementation.cs
+++ b/Refit/RequestBuilderImplementation.cs
@@ -377,7 +377,16 @@ namespace Refit
                     return (T)(object)content; 
                 }
 
-                return JsonConvert.DeserializeObject<T>(content, settings.JsonSerializerSettings);
+                var deserializedObject = JsonConvert.DeserializeObject<T>(content, settings.JsonSerializerSettings);
+
+                // enable message post-processor functionality on resulting object if implements interface,
+                // so it is possible to access http response message even when resulting type
+                // of the call is something more specific (e.g. to allow processing of response headers)
+                if (deserializedObject is IHttpResponseMessagePostProcessor) {
+                    (deserializedObject as IHttpResponseMessagePostProcessor).PostProcessHttpResponseMessage(resp);
+                }
+
+                return deserializedObject;
             };
         }
 
--- a/Refit/RestService.cs
+++ b/Refit/RestService.cs
@@ -15,6 +15,11 @@ namespace Refit
         T For<T>(HttpClient client);
     }
 
+    public interface IHttpResponseMessagePostProcessor
+    {
+        void PostProcessHttpResponseMessage(HttpResponseMessage message);
+    }
+
     public static class RestService
     {
         public static T For<T>(HttpClient client, RefitSettings settings)


bennor:
.NET already has a standard way to intercept requests ([DelegatingHandler](https://msdn.microsoft.com/en-us/library/system.net.http.delegatinghandler(v=vs.118).aspx)).

I'm not sure what this provides that DelegatingHandler doesn't?

enscope:
@bennor: Yes, but `DelegatingHandler` works with `HttpResponseMessage` directly, before the JSON deserialization, or am I mistaken? The point of my patch is able to let Refit deserialize the JSON content and then apply some additional post-processing to deserialized instance, so you don't have to handle deserialization yourself.

The other option would be to know the actual type of the content of Message and somehow inject the values there from `DelegatingHandler`, so it is there when deserializing.

To give an example usage:

	public class APIResponse
		: IHttpResponseMessagePostProcessor
	{
		private const string X_SERVER_TIMESTAMP_HEADER = "X-Server-Timestamp";
		private const string X_TOTAL_COUNT_HEADER = "X-Total-Count";

		[JsonProperty(PropertyName = "error")]
		public ErrorDataObject Error { get; set; }
		public int ServerTimestamp { get; set; }
		public int TotalCount { get; set; }

		public void ThrowExceptionOnError()
		{
			if (Error != null)
			{
				throw RemoteAPICallException.FromDataObject(Error);
			}
		}

		public void PostProcessHttpResponseMessage(HttpResponseMessage message)
		{
			if (message.Headers.Contains(X_SERVER_TIMESTAMP_HEADER))
			{
				ServerTimestamp = int.Parse(message.Headers.GetValues(X_SERVER_TIMESTAMP_HEADER).First());
			}
			if (message.Headers.Contains(X_TOTAL_COUNT_HEADER))
			{
				TotalCount = int.Parse(message.Headers.GetValues(X_TOTAL_COUNT_HEADER).First());
			}
		}
	}

bennor:
Interesting. I can see the value, but I wonder if some of those specific examples could be better served by materialising properties on the response directly? e.g. with a header attribute to indicate where the value should come from.

bennor:
I mean it would be nice if you could just do:

```csharp
[Header("Date")]
public DateTimeOffset ServerTimestamp { get; set; }
```

And have the property value set for you by Refit automatically.

enscope:
Yes, I was thinking about that as well.
 
I went with this simpler approach because of few things:
- there is no real overhead when deserializing the response, if the interface serves as a marker and only then it is processed (contrary to - i suppose - some kind of reflection, as currently the only thing that is accessing the resulting instance is JSON deserializer); but I have to say, I did not have time to dive into Refit too deeply, so there can possibly be some other magic during the auto-generation phase, but from what I saw, it's only JSON deserializer called on Content and handling internal structure of result;
- implementing it this way gives you much higher amount of flexibility with other custom processing of the message if such needed;
- time constraints... as mentioned before, I am under some time constraints and needed a working solution for the API I am interfacing with (and as I like Refit, I did not wanted to change the library)

Maybe the most complete solution would be the combination of two, if the performance hit is not that big with automatic discovery.

enscope:
Is there something new regarding this pull request?

Cheesebaron:
@enscope The argument "I went for a simpler approach" seems to contradict with what @bennor suggests, or it depends on the point of view. From the developers perspective it would be easier just to annotate properties, with the header they want it to be populated from. While the other class allows for more post processing.

This is a personal opinion, but I prefer the annotations compared to the custom postprocessor class.

wekempf:
I think both can be useful. This approach pretty much allows you to do anything, while the attribute approach limits you somewhat. Being able to use a custom attribute (derived from a base attribute or implementing a specific interface) with a similar method to PostProcessHttpResponseMessage but intended to only parse the single value would be as flexible as this approach, but in some cases this approach might still be preferable. For instance, this approach would allow you to handle scenarios where the body isn't JSON fairly easily. Add in a class level attribute that let you control the serialization completely and you'd get a completely flexible system that could handle any payload.

```csharp
[XmlContent]
public class APIResponse
{
	private const string X_SERVER_TIMESTAMP_HEADER = "X-Server-Timestamp";
	private const string X_TOTAL_COUNT_HEADER = "X-Total-Count";

	public ErrorDataObject Error { get; set; }

        [Header(X_SERVER_TIMESTAMP_HEADER)]
	public int ServerTimestamp { get; set; }

        [XmlContentIgnore]
	public int TotalCount { get; set; }

	public void ThrowExceptionOnError()
	{
		if (Error != null)
		{
			throw RemoteAPICallException.FromDataObject(Error);
		}
	}

	public void PostProcessHttpResponseMessage(HttpResponseMessage message)
	{
		if (message.Headers.Contains(X_TOTAL_COUNT_HEADER))
		{
			TotalCount = int.Parse(message.Headers.GetValues(X_TOTAL_COUNT_HEADER).First());
		}
	}
}
```
Throw in the ability to use a base class with these concepts to handle metadata included in every response and you'd further simplify some cases.

onovotny:
Closing due to age. Also, Refit now has `ApiResponse<T>`, which should let you do what you need with the response headers and the content.

PR #307, opened 2017-04-27
Fixed a path separator error on a non-Windows platform.

snowfoxzx:
In the `Linux` or `Mac` platform, when using the `dotnet cli` to build the project, it will cause the runtime to throw `System.InvalidOperationException: {name} doesn't look like a Refit interface. Make sure it has at least one method with a Refit HTTP method attribute and Refit is installed in the project.`

This PR can fix this problem.

### DIFF
--- a/Refit/targets/refit.targets
+++ b/Refit/targets/refit.targets
@@ -27,7 +27,7 @@
     </PropertyGroup>
 
     <PropertyGroup Condition="'$(OS)' != 'Windows_NT'">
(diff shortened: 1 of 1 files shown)


Tanhe:
If this is burning anyone, I have a temporary workaround. Just add this to your csproj to fix the filename between generation and the compile:

` <Target Name="FixRefitStubOnLinux" Condition="'$(OS)' != 'Windows_NT'" AfterTargets="GenerateRefitStubs">
    <Message Text="FixRefitStubOnLinux: Moving '$(IntermediateOutputPath)\RefitStubs.g.cs' to '$(IntermediateOutputPath)/RefitStubs.g.cs'" />
    <Move SourceFiles="$(IntermediateOutputPath)%5CRefitStubs.g.cs" DestinationFiles="$(IntermediateOutputPath)/RefitStubs.g.cs" />
  </Target>`

PR #309, opened 2017-05-01
support for a second GetToken callback that enables refreshing the token in unexpected cases

devmil:
Whenever there is an unexpected 401 because of an invalid token the second callback (if provided) gets called so that the Refit client is able to refresh the token on demand before the overall call results in an Unauthorized exception.
Normally the expiration of a token is tracked on the client side and a refreshed token gets fetched when needed.
This new functionality allows to wire up a fallback logic that refreshes the token whenever the expiration in reality doesn't match the calculated one (e.g. the user changes the system time)


onovotny:
I'm not sure I agree here. The client should be checking the token's expiration date and refreshing it before it expires.

onovotny:
I'm going to close this as this behavior really ought to be up to the caller of the API to get right. If the issue is around clocks being wrong, then the user can wire up that logic in the authentication callback. Also, Polly can be used to handle exceptions and retry in other ways.

PR #317, opened 2017-06-01
make BaseAddress optional

MaxDeg:
Usefull in environment where the BaseAddress is retreived dynamically (service discovery, local load balancing) and still keeping only one HttpClient


bennor:
Seems okay to me. I assume all tests are still passing?

MaxDeg:
Yes indeed

onovotny:
@MaxDeg One last thing before we merge it - can you please add a unit test or two around this property?

MaxDeg:
I think this pull request must be cancelled.
When trying to write some tests I faced an issue with HttpClient. The SendAsync version that refit use is not overridable. This methods perform check or HttpRequestMessage  to ensure the request uri is Absolute or that a BaseAddress is defined on HttpClient.

onovotny:
Is there another approach that would meet your objectives?

MaxDeg:
Well to be honest I don't see any except setting a fake BaseAddress to override.

onovotny:
Closing for now

PR #326, opened 2017-06-21
Minimize memory pressure by not always copying json payload to string

Cheesebaron:
On very large JSON payloads, it maybe be useful not to copy the entire thing into a string before trying to deserialize it. So this uses streams and JSON reader to deserialize.

### DIFF
--- a/Refit/PushStreamContent.cs
+++ b/Refit/PushStreamContent.cs
@@ -1,0 +1,312 @@
+﻿// Copyright(c) Microsoft Open Technologies, Inc.All rights reserved.
+// Microsoft Open Technologies would like to thank its contributors, a list
+// of whom are at http://aspnetwebstack.codeplex.com/wikipage?title=Contributors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License"); you
+// may not use this file except in compliance with the License. You may
+// obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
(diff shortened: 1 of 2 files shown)


Cheesebaron:
@bennor could you please review this? I talked to @onovotny he doesn't know the regular review process in this repo, maybe you could help here 😄

Cheesebaron:
Let me try the PushStreamContent to see how that works. Didn't know of that one.

bennor:
Neither did I. It just felt like writing the whole object to a stream was defeating the purpose a little, so I did some googling. I have no idea if what I suggested will work (or will be available in whatever version framework we're supporting these days).

Cheesebaron:
Seems like it is a ASP.NET MVC thing only. Source is here though: https://github.com/ASP-NET-MVC/aspnetwebstack/blob/master/src/System.Net.Http.Formatting/PushStreamContent.cs

Cheesebaron:
Could we grab the PushStreamContent and redistribute that ourselves? Or is that too much?

onovotny:
It's open source under an MIT license. I'd say that if it works and is useful, then copy the type to here.

Cheesebaron:
It is Apache 2.0

onovotny:
IANAL, but I think that's still ok, esp if we keep the copyright header intact on that file?

Cheesebaron:
I would think it would be OK, especially if we keep header, yes.

Cheesebaron:
I've added PushStreamContent, seems to work.

PR #331, opened 2017-06-28
Use JsonPropertyAttribute for aliasing form parameter names

### LINKED ISSUE #330: UrlEncoded Body Serialization Aliasing
Based on the source from: https://github.com/paulcbetts/refit/blob/master/Refit/FormValueDictionary.cs it appears that aliasing is supported however many people, including myself, will have annotated our models with JSON.NET. Since you already have JSON.NET as a dependency why not also include JsonProperty reflection to support the aliasing feature on models that are only annotated with JSON.NET attributes.

bennor:
This adds support for Json.NET's `JsonPropertyAttribute` if present on a
property when using it to generate a request as a form post.

If both `AliasAsAttribute` and `JsonPropertyAttribute` are present,
`AliasAsAttribute` takes precedence.

I also added some tests for `FormValueDictionary`.

Fixes #330.

### DIFF
--- a/README.md
+++ b/README.md
@@ -219,6 +219,23 @@ var measurement = new Measurement {
 await api.Collect(measurement);
 ``` 
 
+If you have a type that has `[JsonProperty(PropertyName)]` attributes setting property aliases, Refit will use those too (`[AliasAs]` will take precedence where you have both). 
+This means that the following type will serialize as `one=value1&two=value2`:
+
+```csharp
+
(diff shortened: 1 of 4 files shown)


onovotny:
@bennor LGTM but can you please update the docs to mention this so we don't lose track of it?

bennor:
@onovotny Sorry for the delay. I forgot. 😳

Issue #218, opened 2016-02-08
Getting stream content


paulanthonymorris:
Just want to make sure I am not missing something here:
I have an API that has a few JSON endpoints but also a few endpoints with small binary files for download. If I expose the binary as a stream as in...
`[Get("/v2/file.db.gz")]
Task<Stream> GetFileAsync();`
or as in...
`[Get("/v2/file.db.gz")]
Task<byte[]> GetFileAsync();`
Refit wants to parse the result as JSON (which of course blows up).
I could get the raw response as in...
`[Get("/v2/file.db.gz")]
Task<HttpResponseMessage> GetFileAsync();`
and then grab the stream or the bytes from `response.Content` in the caller.
Is this The Right Way?

bennor:
As far as I'm aware, `Task<HttpResponseMessage>` is the only way at the moment. A PR for `Task<byte[]>` or `Task<Stream>` support would be pretty easy. :wink:  ([start here](https://github.com/paulcbetts/refit/blob/master/Refit/RequestBuilderImplementation.cs#L340))

bennor:
Nevermind. I overlooked the `HttpContent` support. If you set your return type `Task<HttpContent>`, you should be able to access the stream from the result.

paulanthonymorris:
Thank you. That works well. I can get the bytes/stream without the extra overhead of the whole response.

PR #338, opened 2017-08-07
Return stream

MaxDeg:
Enable api methods to return the body stream of the response directly (useful in case of file download). 
related to #218


Refit/RequestBuilderImplementation.cs
```diff
@@ -330,7 +330,11 @@ void AddMultipartItem(MultipartFormDataContent multiPartContent, string fileName
         Func<HttpClient, object[], Task> BuildVoidTaskFuncForMethod(RestMethodInfo restMethod)
         {                      
             return async (client, paramList) => {
+<<<<<<< HEAD
                 var factory = buildRequestFactoryForMethod(restMethod.Name, client.BaseAddress?.AbsolutePath, restMethod.CancellationToken != null);
+=======
```
WilkaH: Looks like there's some unresolved merge conflicts here.

Refit/RequestBuilderImplementation.cs
```diff
@@ -382,6 +382,11 @@ void AddMultipartItem(MultipartFormDataContent multiPartContent, string fileName
                 if (restMethod.SerializedReturnType == typeof(HttpContent)) {
                     return (T)(object)resp.Content;
                 }
+
+                if (restMethod.SerializedReturnType == typeof(Stream))
+                {
```
Cheesebaron: Use same style as if above

MaxDeg: done ;)

onovotny:
Please remove `Upgradelog.htm` and then :shipit:

MaxDeg:
Btw any idea about when a new version of Refit will be available? Is it plan for 4.0.0?
Thanks ;)

Issue #315, opened 2017-05-24
Dynamic querystring!


RaminMT:
Hi,
I've got an endpoint which has about 15 query string parameters for filtering purposes. As they are optional parameters, there will be a lot of parameter combination so overloads are useless and Also I should avoid setting null-value parameters as it does affect on the result.
I searched a little seems implementing `IUrlParameterFormatter` is the answer, but I don't have any idea of how to.
Would somebody guide me on this?!

Thanks
Ramin

Cheesebaron:
Isn't this: https://github.com/paulcbetts/refit#form-posts what you want?

RaminMT:
No, mine is a GET request & the parameters are query string parameters

Cheesebaron:
It shouldn't matter that yours is a GET. If you look at the example it is converted into query string parameters.

Cheesebaron:
So:

```
public interface IDerpApi
{
    [Get("/durr")]
    Task Herp([Body(BodySerializationMethod.UrlEncoded)] Dictionary<string, object> stuff);
}

var stuff = new Dictionary<string, object> {
    {"herpy", 1}, 
    {"derpy", "bar"}, 
    {"doo", new Guid("d1e9ea6b-2e8b-4699-93e0-0bcbd26c206c")}
};
```

This will give:

`/durr?herpy=1&derpy=bar&doo=d1e9ea6b-2e8b-4699-93e0-0bcbd26c206c`

RaminMT:
Thank you, I tried that but no success.

Cheesebaron:
What does "no success" mean?

RaminMT:
No success means I didn't get the desired result!!
I expected that to narrow the results but it didn't go that way. I also sniffed the request using fiddler & saw that It has body just like it's a post request not the way you said according to documentations.

RaminMT:
@bennor Well, thank you.
Extending to work with dictionary as #252? Would you please teach me by code or explain it more?

Thanks

RaminMT:
Friends, No comment on this? I really don't have any idea.
@bennor @Cheesebaron @ravensorb

RaminMT:
Friends, No comment on this? I really don't have any idea.
@bennor @Cheesebaron @ravensorb

bennor:
Sorry. I'll try to have a look this weekend.

RaminMT:
Thanks man that would be appreciated, Hope you get back with good news.

bennor:
[...]

PR #355, opened 2017-09-03
Added feature for 'DynamicQueryParameters'

windischb:
I have the need like in #315 to use an Object or Dictionary as Queryparamters.

With this Pull Request this is now possible

```
public class MySimpleQueryParams
{
    public string FirstName { get; set; }

    [AliasAs("lName")]
    public string LastName { get; set; }
}

public class MyComplexQueryParams
{
    public string FirstName { get; set; }

    public string LastName { get; set; }

    [AliasAs("Addr")]
    public Address Address { get; set; } = new Address();

    public Dictionary<string, object> MetaData { get; set; } = new Dictionary<string, object>();

    public List<object> Other { get; set; } = new List<object>();
}

public class Address
{
    [AliasAs("Zip")]
    public int Postcode { get; set; }
    public string Street { get; set; }
}
```  
--------------------
```
public interface IHttpBinApi<TResponse, in TParam, in THeader>
    where TResponse : class
    where THeader : struct
{
    [Get("")]
    Task<TResponse> Get(TParam param, [Header("X-Refit")] THeader header);

    [Get("/get?hardcoded=true")]
    Task<TResponse> GetQuery([Query("_")]TParam param);

    [Get("")]
    Task<TResponse> GetQueryWithIncludeParameterName([Query(".", "search")]TParam param);
}
```  
--------------------
```
[Fact]
public async Task SimpleDynamicQueryparametersTest()
{

    var myParams = new MySimpleQueryParams();
    myParams.FirstName = "John";
    myParams.LastName = "Rambo";

    var fixture = RestService.For<IHttpBinApi<HttpBinGet, MySimpleQueryParams, int>>("https://httpbin.org/get");

    var resp = await fixture.Get(myParams, 99);

    Assert.Equal("John", resp.Args["FirstName"]);
    Assert.Equal("Rambo", resp.Args["lName"]);
}
```
creates following Url: "https://httpbin.org/get?hardcoded=true&FirstName=John&lName=Rambo"

If have created a new Attribute -> QueryAttribute, with the Properties Delimiter and Prefix.
If you do not specifiy the Attribute on the Parameter, the Default Values will be used.
Delimiter = "."
Prefix = null

This is needed for Nested Objects or Dictionaries.
Nested Objects or Dictionaries are "flatted" ->

```
[Fact]
public async Task ComplexDynamicQueryparametersTestWithIncludeParameterName()
{

    var myParams = new MyComplexQueryParams();
    myParams.FirstName = "John";
    myParams.LastName = "Rambo";
    myParams.Address.Postcode = 9999;
    myParams.Address.Street = "HomeStreet 99";

    var fixture = RestService.For<IHttpBinApi<HttpBinGet, MyComplexQueryParams, int>>("https://httpbin.org/get");

    var resp = await fixture.GetQueryWithIncludeParameterName(myParams);

    Assert.Equal("John", resp.Args["search.FirstName"]);
    Assert.Equal("Rambo", resp.Args["search.LastName"]);
    Assert.Equal("9999", resp.Args["search.Addr.Zip"]);
}
```
In combination with Delimiter and Prefix following Url will be created:
https://httpbin.org/get?search.FirstName=John&search.LastName=Rambo&search.Addr.Zip=9999&search.Addr.Street=HomeStreet%2099 

In the Class you can use the AliasAs Attribute for overriding the "Key" of the Queryparameter.

Thats very usefull when you need QueryParameters like "search.Firstname" "search.Lastname"...
So you can create a Class with all possible search parameters. All Properties wich are not null will then be in the Queryparameters. With or without Prefix and/or Delimiter.

If do not like to create a whole Class you can use a simple Dictionary instead.
But this doesn't give you intellisense or any other Class features like Typesafe or the AliasAs Attribute. 

The Second thing i have changed is the format of Arrays as Queryparameters.
Arrays Parameters are now formated as value=1&value=2&value=3.
As far i have found, this is the most used way to specifiy Arrays as Queryparameters...

### DIFF
--- a/README.md
+++ b/README.md
@@ -79,6 +79,39 @@ GroupList(4, "desc");
 >>> "/group/4/users?sort=desc"
 ```
 
+If you specify an `object` as query parameter, all public properties which are not null, where used as query parameters. 
+Use the `Query` attribute the change the behavior of 'flatten' your query parameter object. If using this Attribute you can specify values for the Delimiter and the Prefix which are used for 'flatten' the object.
+
+```csharp
+public class MyQueryParams
+{
+    [AliasAs("order")]
+    public string SortOrder { get; set; }
+
+    public int Limit { get; set; }
+}
+
+
+[Get("/group/{id}/users")]
+Task<List<User>> GroupList([AliasAs("id")] int groupId, MyQueryParams params);
+
+[Get("/group/{id}/users")]
+Task<List<User>> GroupListWithAttribute([AliasAs("id")] int groupId, [Query(".","search")] MyQueryParams params);
+
+
+params.SortOrder = "desc";
+params.Limit = 10;
+
+GroupList(4, params)
+>>> "/group/4/users?order=desc&Limit=10"
+
+GroupListWithAttribute(4, params)
+>>> "/group/4/users?search.order=desc&search.Limit=10"
+```
+
+A similar behavior exists if using a Dictionary, but without the advantages of the `AliasAs` attributes and of course no intellisense and/or type safety.
+
+
 ### Body content
 
 One of the parameters in your method can be used as the body, by using the
--- a/Refit.Tests/RestService.cs
+++ b/Refit.Tests/RestService.cs
@@ -55,6 +55,12 @@ namespace Refit.Tests
     {
         [Get("")]
         Task<TResponse> Get(TParam param, [Header("X-Refit")] THeader header);
+
+        [Get("/get?hardcoded=true")]
+        Task<TResponse> GetQuery([Query("_")]TParam param);
+
+        [Get("")]
+        Task<TResponse> GetQueryWithIncludeParameterName([Query(".", "search")]TParam param);
     }
 
     public interface IBrokenWebApi
@@ -77,7 +83,7 @@ namespace Refit.Tests
 
     public class HttpBinGet
     {
-        public Dictionary<string, string> Args { get; set; }
+        public Dictionary<string, object> Args { get; set; }
         public Dictionary<string, string> Headers { get; set; }
         public string Origin { get; set; }
         public string Url { get; set; }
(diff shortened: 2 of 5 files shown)


windischb:
I have added an example to the README and have tried to update the formatting...
But i'm not sure if the formatting is now better.
Maybe you can give me some hints what i have to change?

Refit/RequestBuilderImplementation.cs
```diff
@@ -138,7 +138,19 @@ public RequestBuilderImplementation(Type targetInterface, RefitSettings refitSet
                     // for anything that fell through to here, if this is not
                     // a multipart method, add the parameter to the query string
                     if (!restMethod.IsMultipart) {
-                        queryParamsToAdd[restMethod.QueryParameterMap[i]] = settings.UrlParameterFormatter.Format(paramList[i], restMethod.ParameterInfoMap[i]);
+                        var attr = restMethod.ParameterInfoMap[i].GetCustomAttribute<QueryAttribute>() ?? new QueryAttribute();
+                        if (DoNotConvertToQueryMap(paramList[i]))
+                        {
```
bennor: :arrow_up: 

The style the project uses can be a little hard to follow, but the gist is that we don't use access modifiers where they're unnecessary (e.g. `private`) and we use the [K&R indent style](https://en.wikipedia.org/wiki/Indent_style#K.26R) i.e. opening bracket on the same line as the condition.

Refit/RequestBuilderImplementation.cs
```diff
@@ -138,7 +138,19 @@ public RequestBuilderImplementation(Type targetInterface, RefitSettings refitSet
...
+                            queryParamsToAdd.Add(new KeyValuePair<string, string>(restMethod.QueryParameterMap[i], settings.UrlParameterFormatter.Format(paramList[i], restMethod.ParameterInfoMap[i])));
+                        }
```
bennor: These three lines whould be written as `} else {`

Refit/RequestBuilderImplementation.cs
```diff
@@ -425,5 +441,106 @@ void AddMultipartItem(MultipartFormDataContent multiPartContent, string fileName
...
+        {
+
+            if (value == null) return false;
+
+            var type = value.GetType().GetTypeInfo();
+
+            if (type.IsPrimitive) return true;
+
+            switch (value)
+            {
```
bennor: This should also be on the previous line.

bennor:
Fixes #315.

README.md
```diff
@@ -79,6 +79,39 @@ GroupList(4, "desc");
 >>> "/group/4/users?sort=desc"
 ```
 
+If you specify an `object` as query parameter, all public properties which are not null, where used as query parameters.
```
bennor: Might pay to put a subheading in here (e.g. "Dynamic querystring parameters").

bennor:
One more thing that needs to be considered -- I'm not sure we want this to be automatic. I'm about 50/50 on thinking that to enable this behaviour with objects you should need a `[Query]` parameter to make it explicit. Although perhaps this automatic behaviour is exactly how people would expect it to work? 

What do you think @onovotny or @paulcbetts?

windischb:
Thanks for helping and explanation!
I have tried to recognized the formatting from other code pieces in the Project.

There a many places where the K&R Style isn't used and which i have never touched.
For example:
```csharp
                     // if marked as body, add to content
                    if (restMethod.BodyParameterInfo != null && restMethod.BodyParameterInfo.Item2 == i) {
                        var streamParam = paramList[i] as Stream;
                        var stringParam = paramList[i] as string;

                        if (paramList[i] is HttpContent httpContentParam)
                        {
                            ret.Content = httpContentParam;
                        }
                        else if (streamParam != null)
                        {
                            ret.Content = new StreamContent(streamParam);
                        }
                        else if (stringParam != null)
                        {
                            ret.Content = new StringContent(stringParam);
                        }
                        else
                        {
                            switch (restMethod.BodyParameterInfo.Item1)
                            {
                                case BodySerializationMethod.UrlEncoded:
                                    ret.Content = new FormUrlEncodedContent(new FormValueDictionary(paramList[i]));
                                    break;
                                case BodySerializationMethod.Json:
                                    var param = paramList[i];
                                    ret.Content = new PushStreamContent((stream, _, __) => {
                                        using(var writer = new JsonTextWriter(new StreamWriter(stream))) {
                                            serializer.Serialize(writer, param);
                                        }
                                    }, "application/json");
                                    break;
                            }
                        }

                        continue;
                    }
```

That makes it  really hard for others who would like to help, to find out how they should format there pieces of code.
Maybe it's possible to change the .editorconfig for that??

Anyway... i change my code as you suggested!

Many thanks for helping!

bennor:
Yeah, there have been a few PRs that have slipped through without the same style applied. We need to clean that up.

bennor:
This looks better in terms of formatting. Just need to confirm that we're happy with the behaviour being automatic.

SuperJMN:
I'm looking forward to it being merged 😍

chadly:
> Just need to confirm that we're happy with the behaviour being automatic.

FWIW, I feel like this behavior being automatic is less surprising since that is the way Asp.Net MVC Core model binding works (which many using this library might already be familiar with). Especially, since the `[Body]` attribute is already required to send an object in the request body, the thinking from a user perspective is "if I don't put `[Body]` it will show up in the query string."

jackjwilliams:
Any idea where we are at on this? I agree with @chadly about it being automatic. For me, this removes the need for a bunch of parameters everywhere (for gets with a large amount of arguments). I can re-use my SearchDto object and I'm good to go.

Refit.Tests/RestService.cs
```diff
@@ -419,5 +425,92 @@ public void NonRefitInterfacesThrowMeaningfulExceptions()
...
+        public async Task SimpleDynamicQueryparametersTest()
+        {
+
+            var myParams = new MySimpleQueryParams();
+            myParams.FirstName = "John";
+            myParams.LastName = "Rambo";
+
+            var fixture = RestService.For<IHttpBinApi<HttpBinGet, MySimpleQueryParams, int>>("https://httpbin.org/get");
+
+            var resp = await fixture.Get(myParams, 99);
```
onovotny: can we stub this test out? Don't use a real service as it may cause intermittent test failures. There are other examples in the tests of stubbing out a service.

Refit.Tests/RestService.cs
```diff
@@ -419,5 +425,92 @@ public void NonRefitInterfacesThrowMeaningfulExceptions()
...
+            myParams.MetaData.Add("Initials", "JR");
+            myParams.MetaData.Add("Birthday", DateTime.Now.AddYears(-99));
+
+            myParams.Other.Add(12345);
+            myParams.Other.Add(DateTime.Now);
+            myParams.Other.Add(Guid.NewGuid());
+
+
+            var fixture = RestService.For<IHttpBinApi<HttpBinGet, MyComplexQueryParams, int>>("https://httpbin.org");
+
```
onovotny: Same as above, stub the service out

Refit.Tests/RestService.cs
```diff
@@ -419,5 +425,92 @@ public void NonRefitInterfacesThrowMeaningfulExceptions()
...
+            myParams["LastName"] = "Rambo";
+            myParams["Address"] = new
+            {
+                Zip = 9999,
+                Street = "HomeStreet 99"
+            };
+
+
+            var fixture = RestService.For<IHttpBinApi<HttpBinGet, Dictionary<string, object>, int>>("https://httpbin.org");
+
```
onovotny: Same

Refit.Tests/RestService.cs
jackjwilliams: Are you referring to using a MockHttpMessageHandler?

PR #366, opened 2017-09-14
Prevent CS1591 warnings/errors in generated code

martincostello:
Use `<inheritdoc/>` to add perfunctory comments to Refit-generated code so that projects using Refit that generate XML comments don't get CS1591 warnings/errors from the compiler due to missing XML documentation.


martincostello:
It doesn't _really_ work in the sense of the documentation being imported and used etc., but it counts as a valid comment as far as the compiler's concerned, whereas an empty `/// <summary</summary>` might generate complains of _actual_ empty comments. Happy to change it (and the constructor) to have a proper comment though if you prefer.

martincostello:
Just discovered this wasn't a 100% fix as I missed the constructor as well.

PR #368, opened 2017-09-15
Added Windows authenticaion settings

SaarCohen:
We have some clients who use Windows authentication so have added this setting class to help with setting the required credentials. Hope this helps the community.

### DIFF
--- a/Refit/WindowsAuthenticationRefitSettings.cs
+++ b/Refit/WindowsAuthenticationRefitSettings.cs
@@ -1,0 +1,33 @@
+﻿using System.Net;
+using System.Net.Http;
+
+namespace Refit
+{
+    public class WindowsAuthenticationRefitSettings : RefitSettings
+    {
+        public WindowsAuthenticationRefitSettings()
+        {
+            SetCredentials(CredentialCache.DefaultCredentials);
+        }
+
+        public WindowsAuthenticationRefitSettings(string username, string password)
+        {
+            SetCredentials(new NetworkCredential(username, password));
+        }
+
+        public WindowsAuthenticationRefitSettings(string username, string password, string domain)
+        {
+            SetCredentials(new NetworkCredential(username, password, domain));
+        }
+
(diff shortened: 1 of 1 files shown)


onovotny:
I think this would be great for documentation, but I don't think this belongs directly in the main library. Can you create a PR showing this as an example in the documentation?

Cheesebaron:
Also, will this work with NSUrlSessionHandler and AndroidHttpHandler on Xamarin.iOS and Xamarin.Android respectively? You are forcing `HttpClientHandler` here.

SaarCohen:
Will add it to the docs, thanks.
It work with Xamarin for iOS  and Android.

napalm684:
Thank you for this, was cursing the windows authentication gods!

onovotny:
Closing PR. This should be docs.

Styxxy:
@onovotny But it is not added to the docs at the moment.

There are two options to solve this, create a HttpClient yourself, or set the clienthandlerfactory in the RefitSettings;

```csharp
// Option 1 : create HttpClient yourself
var client1 = RestService.For<IMyService>(
    new HttpClient(new HttpClientHandler() { AllowAutoRedirect = true, UseDefaultCredentials = true })
    {
        BaseAddress = new Uri("http://example.com/")
    });

// Option 2 : set the ClientHandler factory
var client2 = RestService.For<IMyService>(
    "http://example.com/",
    new RefitSettings()
    {
        HttpMessageHandlerFactory = () => new HttpClientHandler() { AllowAutoRedirect = true, UseDefaultCredentials = true }
    });
```

onovotny:
@Styxxy Yup, true. Can you please open a separate issue to track that (or a PR adding it to the docs?)

Thanks!

Styxxy:
@onovotny I created an issue. I don't have time at the moment to add this to the documentation. Maybe at a later point I'll do it, but it is up-to-grabs for anyone.

PR #374, opened 2017-09-27
Allow switching to Pull stream method

### LINKED ISSUE #362: Content-Length no longer set when using Refit 4
I am currently using `3.1.0`, the `Content-Length` header is set on requests. After upgrading to `4.0.1`, it is no longer set.

Cheesebaron:
This should fix #362 by allowing to use `[Body(StreamMethod.Pull)]`.

### DIFF
--- a/Refit.Tests/RequestBuilder.cs
+++ b/Refit.Tests/RequestBuilder.cs
@@ -52,7 +52,7 @@ namespace Refit.Tests
         Task<bool> OhYeahValueTypes(int id, [Body] int whatever);
 
         [Post("/foo/{id}")]
-        Task<bool> PullStreamMethod(int id, [Body(StreamMethod.Pull)] Dictionary<int, string> theData);
+        Task<bool> PullStreamMethod(int id, [Body(buffered: true)] Dictionary<int, string> theData);
 
         [Post("/foo/{id}")]
         Task VoidPost(int id);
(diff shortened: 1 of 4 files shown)


Cheesebaron:
@bennor @onovotny not sure what else to add to the test suite.

bennor:
Sorry - meant to comment on this the other day, but forgot to.

I am a tiny bit concerned that the terminology "push vs pull" isn't clear to the user. Wouldn't something like `Buffered`/`Streamed` on an enum or just a boolean `Buffered` property be clearer?

Cheesebaron:
The Push/Pull terminology is similar to what MS has for the PushStreamContent. I have no issues with changing it or making this a bool instead.

So `Buffered` here would mean that `StringContent` is used just like before I added `PushStreamContent`?

bennor:
That's what I was thinking for buffered, but I'm open to other options. I just think push vs pull might not be that clear i.e. what's pulling?

Either way I think we'll need something in the docs.

Cheesebaron:
Pulling is seen from the server side. The Client is asking the server to pull content until there is no more. This is why the `Content-Length` is set to -1 as it is unclear how much data there is, or it is so large that you don't want it loaded into your memory to calculate the size.

Pushing is when `Content-Length` is known, and it is part of the body.

bennor:
That's not really what's happening though, right? We're still pushing the content, but we're doing it all in one hit instead of streaming it.

Cheesebaron:
Can we maybe switch to ValueTuples at some point? I'd love to have named tuple items instead of Item1, Item2 and Item3 etc.

Issue #371, opened 2017-09-26
Question: How to make 404 return null?


nvivo:
I have a rest API that returns 404 if the record is not found. How do I make Refit return null instead of throwing an exception?

Cheesebaron:
You can't. Catch the exception

nvivo:
Is there any way I could implement this as a "filter" plugin, or an attribute on the method? The main reason to use Refit is it's simplicity. Once we're forced to add lots of code to each call to handle common things, it starts to lose it's value.

Cheesebaron:
PR's are welcome. You could probably add a attribute like `[NullOnException]` or whatever.

However, I personally don't see how catching a valid exception is "loss of value".

nvivo:
Hi. It's not that "cathing an exception is a loss of value". Is the fact that I expect this to be the default behavior against any call to the API. Having to add a try/catch to every call ends up being too much boilerplate for this.

I'll see if I can add this as an attribute. 

Thanks.

Cheesebaron:
Why would you need to write that for each of the calls?

You could just wrap your stuff in a simple method:

```csharp
async Task<(bool, T)> TryCallStuffAsync<T>(Func<Task<T>> whatever)
{
	try
	{
		var derp = whatever();
		var herp = await derp.ConfigureAwait(false);
		return (true, herp);
	}
	catch (Exception ex)
	{
		Logger.LogException(ex);
		return (false, default(T));
	}
}
```

or simpler:

```csharp
async Task<T> TryCallStuffAsync<T>(Func<Task<T>> whatever)
{
	try
	{
		var derp = whatever();
		return await derp.ConfigureAwait(false);
	}
	catch (Exception ex)
	{
		Logger.LogException(ex);
		return default(T);
	}
}
```

Then invoke with:

```csharp
TryCallStuffAsync(() => Derp());
```

Returning null instead of throwing the exception seems kind of silly to me as default behavior, because what does null then mean? No internet connection? Error in the call because you provided bogus values? Internal Server Error? Bad authentication? How the hell do you expect someone to figure that out by just returning null?

nvivo:
@Cheesebaron I'm not saying to return null or any case of error. I'm saying that 404 maps to null on some cases and it would be nice to have this directly. This is pretty much how `DbContext.Find()` or `.SingleOrDefault` works.

Maybe we just have different views on this, but I think it's a good idea. The PR is there for analysis.

PR #375, opened 2017-09-27
Adds a generic way to filter unsuccessful responses and a DefaultOn404 attribute

nvivo:
This PR is related to issue #371. It adds an `UnsuccessfulResponseFilterAttribute` that allows the user to handle non 200 responses and return something else.

It also adds a `NullOn404` attribute that will return null when the response is 404. This is useful for APIs that work this way, and return 404 when the resource is not found which maps to null on .NET (pretty much how `DbContext.FindAsync()` or `SingleOrDefault()` works.

I'd make a `DefaultOn404` that returns default(T) but netstandard1.4 doesn't have have a Type.IsValueType to test and I think anyone using this would be returning classes anyway.

In order to use this, just add a `[NullOn404]` attribute to any api method, and it will return null instead of throwing an `ApiException` if 404 is returned from the server. For example:

```
interface MyApi {

     [NullOn404]
     [Get("/resource/{id}"]
     SomeResource GetResource(int id);

}
```

If there is anything to fix or improve, let me know.


Refit/Attributes.cs
```diff
@@ -148,4 +148,19 @@ public AuthorizeAttribute(string scheme = "Bearer")
...
+    {
+        public int Order { get; set; }
+        public abstract void Filter(UnsuccessfulResponseFilterContext ctx);
+    }
+
+    public class NullOn404 : UnsuccessfulResponseFilterAttribute
+    {
+        public override void Filter(UnsuccessfulResponseFilterContext ctx)
+        {
+            if (ctx.HttpResponse.StatusCode == System.Net.HttpStatusCode.NotFound)
```
Cheesebaron: Add `using System.Net` instead of fully qualifying here.

Refit/RequestBuilderImplementation.cs
```diff
@@ -38,7 +38,12 @@ public RequestBuilderImplementation(Type targetInterface, RefitSettings refitSet
                     var hasHttpMethod = attrs.OfType<HttpMethodAttribute>().Any();
                     if (!hasHttpMethod) return Enumerable.Empty<RestMethodInfo>();
 
-                    return EnumerableEx.Return(new RestMethodInfo(targetInterface, x, settings));
+                    var unsuccessfulFilters = attrs.OfType<UnsuccessfulResponseFilterAttribute>();
+
+                    return EnumerableEx.Return(new RestMethodInfo(targetInterface, x, settings)
+                    {
+                        UnsuccessfulResponseFilters = unsuccessfulFilters.Any() ? unsuccessfulFilters.OrderBy(f => f.Order).ToArray() : null
```
Cheesebaron: Multiple enumeration of `IEnumerable`

Refit.Tests/GitHubApi.cs
```diff
@@ -76,5 +76,9 @@ public interface IGitHubApi
 
         [Get("/give-me-some-404-action")]
         Task NothingToSeeHere();
+
+        [NullOn404]
+        [Get("/give-me-some-404-action")]
+        Task<object> ReturnsNullOn404();
```
Cheesebaron: Add test for value types too

Refit/RequestBuilderImplementation.cs
nvivo: Here, `.Any()` checks only for the first item, and is better than allocating memory with `ToList` or `ToArray`. Considering the source is already an array, `.Any()` is amost free, and it only iterates the enumerable if it really needs to.

Refit/Attributes.cs
nvivo: Will do.

Refit.Tests/GitHubApi.cs
nvivo: The null filter can't work with value types. Do you mean adding a test to make sure value types won't go through the filter and will throw the regular exception? Or do you mean changing the implementation to a `DefaultOn404`?

Cheesebaron: It should still work with `Task<int> ReturnsNullOn404()` probably throwing an exception or something like that.

onovotny: I think Default makes more sense than Null as it covers all the bases. That said, I'm not sure if all of these attributes belong in Refit at all. The base implementation, perhaps, but I'm not sure what's special about 404 to warrant it having an implementation here when all of the other responses aren't touched.

nvivo:
@onovotny I'm about to push a new commit that fix and changes it to a DefaultOn404 attribute, that makes things more consistent and solves this issue of value types.

nvivo:
@onovotny, @Cheesebaron, I did some changes to the implementation:

1. Change `NullOn404` to `DefaultOn404`, so it handle reference and value types as expected
2. Make the filter logic faster by allowing the filer to be pre-built instead of having to reanalyze the RestMethodInfo every time.

Still, per @onovotny comment, I feel I should defend why I think this is a good idea. Please bear with me.

When I found refit, the first thing I noticed was how simple it was to express the API in a single interface. There was no need for additional code, all details were encapsulated into attributes. If I wanted to change how the body was encoded, or add an additional header, a simple attribute on the method or the parameter would do it. It's simple and doesn't require any boilerplate when calling the method to make these changes.

This implementation follows the same idea. For REST services, non-200 responses are not exceptions, it's the way the API was built. Most REST APIs will return 404 when the resource is not found, instead of a 200 with a null response body . Forcing exceptions for something that is expected of the api is not the best behavior IMO.

So far, Refit allows us to control only what is sent to the API, but not how to handle the result. I don't believe this is something against the idea of Refit. Yes, I'm committing only the 404 filter because it's the most basic filter and I think it's common enough that it should be included by default, but the implementation has a base attribute that can be used to filter any type of unsuccessful response.

For example, most REST APIs return an empty 200 response on OK, but a JSON object for 400 responses to indicate that something went wrong. The way to map this to methods is that a method returns something on error, or nothing on success. 

So, why have an api that returns a empty Task, and throws an exception with the actual return value, instead of just returning the actual value? Refit should allow the user to do something like:

```cs
public interface SomeApi
{
    [HandleApiError] // just an example
    Task<ApiResponse> CreateAsync(Record record);
}
...
var error = await api.CreateAsync(record)

if (error != null)
   // do something
```

This implementation would let anyone do that by implementing the attribute in their own application.

Maybe there is more that must be done here, maybe there is less. Maybe we could add only the base attribute so at least we can handle these cases in the app with some support from the library. I'm ok with changing or improving the PR, but I don't think Refit should let everything to be done with exception handling in user code.

Saying that anything like this can be done with some lines of code is just like saying that anything Refit does can be done with System.Net.Http and some lines of code. The power of the library is to have these things built-in so we don't have to code them later.

bennor:
I think the idea of returning something on error and nothing otherwise is the wrong approach. It's confusing as it would nmean some API methods would work the opposite way to others.

I would prefer to see that situation handled by something like support for returning a tuple e.g. 

```csharp
[Get("/whatever")]
Task<(HttpStatusCode, SomeResultType)> Whatever();
```

onovotny:
+1 @bennor. I like the idea of enriching the return type better than changing the behavior of particular methods.

Cheesebaron:
This is also somewhat what I expressed in #371. Just annotating to have Refit "handle" the error feels strange. Returning `null` would start to mean so many different things and you won't be able to see how this is handled unless you look at the interface itself and see what it is annotated with.

@bennor in your example, would this mean that no exception is thrown when using this tuple as the status is always returned? Will Refit in this case also try to get the response body on error codes? I.e. could it be a 404 with JSON body indicating what is wrong?

bennor:
What I was thinking was either a tuple containing an `HttpStatusCode` or a special Refit response type like: 

```csharp
public interface IApiResponse<TData, TErrorData> // This could also be `class ApiResponse`
{
    bool IsSuccessStatsusCode {  get; }
    TData Data { get; }
    TErrorData ErrorData { get; }
}
```

If the return type is this, we populate it according to the status (i.e. if it's not an error we populate `Data` and if it's an error we try to populate `Error`), but we never actually throw an exception. (Similar to how returning a raw `HttpResponseMessage` works.)

Both examples would make for pretty straightforward usage for @nvivo's API: 

```csharp
// Tuple
var (status, resource) = await GetResource(id);
if(resource != null) 
{
     // Do something with resource
}

// IApiResponse/ApiResponse
var response = await GetResource(id);
if(response.IsSuccessStatusCode) // or `response.Data != null`
{
    // Do something with response.Data
}
```

nvivo:
Just to be clear, I'm not saying `var error = api.Create(resource)` is a good API. I'm saying it's how existing APIs work and it would be nice if refit supported a simple mapping of that. Most of the APIs I use refit with are created by other companies and I have no control over what they did.

So, I really don't think that Refit should care if *my* application has some methods that work in a different way than *you* expect. In the NullOn404 for example, it's just not an issue at all. It is what already happened before moving to refit, it is what the documentation says it should happen, it's what the official .NET API for the service does for these cases. Having to handle exceptions for every non 200 response that is what is actually confusing for me. And since perception of what is good is not the same for everybody, the library shouldn't focus on saying what is right or wrong, it should simply allow the user to do stuff.

Now, having said that, I like @bennor ideas for both tuples and the interface.
But for the ApiResponse, I'd vote to a return more raw object like this:

```cs
public interface IApiResponse
{
    bool IsSuccessStatsusCode { get; }
    string ResponseBody { get; }
}
```

And then a generic typed class version to get the typed response if needed:

```cs
public class ApiResponse<T> : IApiResponse
{
    private T _data;
    public T Data => _data != default(T) ? _data : (_data = Deserialize<T>(ResponseBody));
}
```

I think this separation is needed because some API calls like POST, PUT and DELETE never have any response on 200, so there would be no value to cast to.

Having the raw interface would allow implementation of specific response classes, and allow the user to build whatever it wants to handle that specific response, and Refit wouldn't have to care about it:
```cs
// in application code
class CreateRecordResponse : ApiResponse
{
    public bool ShouldRetryLater => StatusCode == 429;
    public CreateError Error => Deserialize<CreateError>(ResponseBody);
    public string AlreadyExists => Error.ErrorCode == "DUPLICATE_ID";
    public Dictionary<string,string> ConvertErrorsToAngularAppFormat() { ... };
}
```

Again, just brainstorming here, it's an example. But I think keeping Refit to the minimum and letting the user implement whatever they want in appland would be better than adding something like `TErrorData` to Refit.

nvivo:
So, how will this go? I'm not comfortable changing Refit at this level, and probably you guys will decide better what you want to do or not.

Still, I just need the responses filtered and no exceptions. For now I think I'll just use this fork as it is.

nvivo:
Just to follow up. Has any decision come from this? Any chance I can go back using the official package anytime?

onovotny:
The latest CI code now supports `ApiResponse<T>` as a return type to get both the content and headers without automatically throwing. I think that's the best option for this as you can use whatever logic you need when unrapping the result.

czeslav87:
Hello,
I've encountered same issues while working with Refit. What's current suggested approach to handle such situations?
I like the idea of [NullOn404] or [DefaultOn404] attributes. Other solutions that comu to my mind have drawbacks i don't like:
1. Wrapping Interface implemented by Refit client with another handler. Kinda ruins the idea of using Refit in the first place.
2. Using custom wrapper that handles the 404 Exception (catch ApiException) and returns default/null ("HttpHelpers.DefaultOn404(() => IMyInterface.MyMethodWithReturn()"). 2 issues i see here: DRY violation (i have to wrap it whenever i use 404 exposed method) and the fact that I need to know that the Interface invokation is in fact Refit implementation (due to ApiException catching requirement).

I'd appreciate any other ideas/suggestions.
Best regards.

PR #399, opened 2017-12-20
Added support for nested interface. 

tobiasz256:
I'm wrote a simple feature thats allow to add nested interface in class.


onovotny:
LGTM @bennor ?

Issue #401, opened 2017-12-20
QueryParams are serialized using the current culture


koen-unifly:
We had some strange api calls/urls-requests where doubles are represented differently based on the device culture.
For as far as I know there is no standard on how to _serialize_ query parameters.
However this is counter intuitive since Json has a clear format. From my opinion Refit should enforce the same format (use Culture.Invariant), or make it configurable?

onovotny:
@koen-unifly I have a fix for this in the query string, just wanted to confirm you're seeing the correct behavior for JSON serialization or is there something we need to fix there as well?

koen-unifly:
Thanks. We didn't observer the issue with JSON serialization.

koen-unifly:
I am not sure if there is a slack channel, gitter or something else to ask these kind of questions, but:
Is there a nuget package / release planned in the near future?

PR #402, opened 2017-12-20
Fix for #401

onovotny:


bennor:
Worth noting that this could be customised by creating a custom `IUrlParameterFormatter` (or just overriding that method in the default one).

onovotny:
@bennor yeah, true, but I think the 99% case is that serialization formats should be invariant.

Issue #23, opened 2013-12-30
Support for getting response headers


ray2k:
(This issue is at the request of @paulcbetts over twitter, in which I stated I'm willing to give a pull req a try, but some discussion on scenarios first.)

I would like to see support for getting response headers out of the remote api call while still using a strongly typed response (or none at all). This would be for making calls against APIs (ex the azure management api) that return a unique request-id as a response header, which can be used in support cases when something unexpected/bad happens, or more frequently in the case of asynchronous calls in which a call is made against the api such as swapping an azure deployment, in order to take that returned request-id and query its status in a future api call. 

Given the way refit is exposed to it's user, the most natural way I feel would be to use attributes i.e.

``` C#
// async api that returns no body
[Post("/some/asyncapi/{someId}")]
Task SomeAsyncRemoteApiCall(int someId, [ResponseHeader("x-request-id")] out string requestId);

// async api that returns body and requestid header
[Post("/some/asyncapi/{someId}")]
Task<SomeResponse> SomeAsyncRemoteApiCall(int someId, [ResponseHeader("x-request-id")] out string requestId);
```

I suppose that an alternative to the first one could be

``` C#
[Post("/some/asyncapi/{someId}")]
[return: ResponseHeader("x-request-id")]
Task<string> SomeAsyncRemoteApiCall(int someId);
```

where the returned Task<string> is for the header, not the string content because the request doesn't return a body. But that seems a little wonky IMO.

Thoughts/comments/insults?

fonix232:
Any update on this? I'm in need of retrieving headers for a project, where a CSRF Token is returned via the header in the response.

martingemme:
I would be too interested to have the status code.

onovotny:
The latest code in `master` supports returning an `HttpResponseMessage` object and that would contain the headers.
https://github.com/paulcbetts/refit/blob/master/Refit/RequestBuilderImplementation.cs#L318

bennor:
Fairly sure that's been there since v1.x. Working with `HttpResponseMessage` is a decent workaround at the moment, but I think it would be nice to allow returning something with a bit more of a stripped down interface. I'll look at this when I next have some time.

fonix232:
Good to know. It could be added to the documentation (no matter how small it is right now), so that others who want to work with received headers.

auriou:
Does that has been implemented since?

 I would like get the header responses , how can we do?

ahmedalejo:
[...]

PR #414, opened 2018-01-26
Support for getting response headers

tcfialho:
Implementing support for getting response headers as requested in issue #23  most voted suggestion.

To get the response metadata you should wrap your return type with RefitResponse

Task<RefitResponse> GetUserWithMetadata(string userName);
or
IObservable<RefitResponse> GetUserObservableWithMetadata(string userName);


Refit/IRefitResponse.cs
```diff
@@ -0,0 +1,18 @@
+﻿using System;
+using System.Net;
+using System.Net.Http;
+using System.Net.Http.Headers;
+
+namespace Refit
+{
+    public interface IRefitResponse<T>
```
bennor: I don't think there's a need for this interface. It doesn't seem to serve any real purpose -- you'd never want to mock it or inject another implementation. so it's just adding noise IMO.

Refit/RefitResponse.cs
```diff
@@ -0,0 +1,35 @@
+﻿using System;
+using System.Net;
+using System.Net.Http;
+using System.Net.Http.Headers;
+
+namespace Refit
+{
+    public class RefitResponse<T> : IRefitResponse<T>
```
bennor: Please rename to `ApiResponse` for consistency with `ApiException`.

Refit/RequestBuilderImplementation.cs
```diff
@@ -170,8 +170,16 @@ void AddMultipartItem(MultipartFormDataContent multiPartContent, string fileName
                             return (T)(object)await reader.ReadToEndAsync().ConfigureAwait(false);
                         }
 
-                        using(var jsonReader = new JsonTextReader(reader))
+                        using (var jsonReader = new JsonTextReader(reader))
                         {
+                            if (restMethod.SerializedReturnType.IsConstructedGenericType &&
+                                restMethod.SerializedReturnType.GetGenericTypeDefinition() == typeof(RefitResponse<>).GetGenericTypeDefinition())
```
onovotny: Should be IsAssignableTo `IRefitResponse<T>` or it'll miss other implementation types.

bennor: The `GetGenericTypeDefinition()` call on `typeof(RefitResponse<>).GetGenericTypeDefinition()` is a noop -- `typeof(RefitResponse<>)` is already a generic type definition.

bennor: I'm not sure there's a point in that interface at all -- there's only ever one implementation, right? And it's just a DTO.

Refit/IRefitResponse.cs
onovotny: Yeah, probably no need for that.

Refit/RefitResponse.cs
onovotny: I would suggest this be a `struct` and delegate the internal calls to the `HttpResponseMessage` instance.

onovotny: Finally, have a "ThrowIfNotSuccess" method on it that throws an `ApiException` just like it would have if it was a `T` response.

Refit/RequestBuilderImplementation.cs
tcfialho: @bennor  Removing the interface as suggested.
@onovotny restMethod.SerializedReturnType.IsAssignableFrom(typeof(ApiResponse<>)) returns false due T is ApiResponse<MyClass>.

Refit/IRefitResponse.cs
tcfialho: @bennor  @onovotny Removing the interface as suggested.

Refit/RequestBuilderImplementation.cs
onovotny: It's moot if it's a struct with one impl now.

Refit/RefitResponse.cs
tcfialho: @bennor renammed as suggested.
@onovotny changed to struct. I do not quite understand the other suggestions can give me some examples. I think you are asking for something like ApiResponse.Create(...,...); but T is ApiResponse<T> so pass T means ApiResponse<ApiResponse<T>>.Create(...,...).

Refit/RequestBuilderImplementation.cs
tcfialho: Nice! I made the changes as suggested.

Refit/ApiResponse.cs
```diff
@@ -27,7 +27,7 @@ public class RefitResponse<T> : IRefitResponse<T>
             Content = content;
         }
 
-        internal RefitResponse(HttpResponseMessage response,
+        internal ApiResponse(HttpResponseMessage response,
```
onovotny: Do we need this here? Can we remove this overload and do the cast where needed externally?

tcfialho: I removed the internal costructor and the cast as is not needed.

onovotny:
The location of where the `ApiResponse` checks happen should be higher up:

https://github.com/tcfialho/refit/blob/0639d7fa025b62151d2196c51ee91f8a31aa6c0d/Refit/RequestBuilderImplementation.cs#L176

The issue it that it only applies for the fallback case of json objects. I should be able to have `ApiReponse<Stream>`, `ApiResponse<HttpContent>`, etc.

It also doesn't allow a failure response to be returned since https://github.com/tcfialho/refit/blob/0639d7fa025b62151d2196c51ee91f8a31aa6c0d/Refit/RequestBuilderImplementation.cs#L147 will fire first.

Here's my thoughts:

I'd put the ApiResponse creation above line 147 if that's there requested type. It should work regardless of SuccessStatusCode.

There might need to be some refactoring so logic isn't being duplicated -- perhaps a `Func<T>` passed into the `ApiResponse` constructor to read the value, etc.

You might need a factory to create the ApiResponse given that there's an await that'll need to be called -- `static Task ApiResponse Create(...)` where it reads processes the `Func<Task<T>>`, extracts the value, then creates and returns the response.

onovotny:
Thanks, I rebased on master and will review/merge this in the next couple days.

PR #349, opened 2017-08-25
implement MethodOverload

windischb:
I have changed the code in RequestBuilderImplementation.
RestMethodInfos are collected in a ConcurrentDictionary<string, List<RestMethodInfo>>.
This allows to collect also overloaded Methods.
When executing the RestMethod, the best matching Method will be picked.
Finding the correct Method is based on Reflection, to find the Method with matching parameter types.

PR #416, opened 2018-01-27
MethodOverload

windischb:
Replaces #349 

There were too many gaps between the two Repositories!  
I have removed the old one, and started from scratch with current codebase...
Hopefully it can be implemented now.


Refit/RequestBuilderImplementation.cs
```diff
@@ -32,23 +33,80 @@ public RequestBuilderImplementation(Type targetInterface, RefitSettings refitSet
...
-                                                               .OfType<HttpMethodAttribute>().Any())
-                                                  .Select(x => new RestMethodInfo(targetInterface, x, settings))
-                                                  .ToDictionary(k => k.Name, v => v);
+
+            foreach (var methodInfo in targetInterface.GetMethods()) {
+                var attrs = methodInfo.GetCustomAttributes(true);
+                var hasHttpMethod = attrs.OfType<HttpMethodAttribute>().Any();
+                if (hasHttpMethod) {
+                    var restinfo = new RestMethodInfo(targetInterface, methodInfo, settings);
+                    interfaceHttpMethods.AddOrUpdate(methodInfo.Name, s => new[] { restinfo }.ToList(), (s, list) => {
```
onovotny: I'm not sure what the concurrency guarantees are trying to be here. On one side there's a concurrent dictionary but then inside there's a regular list. If the concern is that the generator method may be called on different threads at the same time, this could lead to duplicate items in the list even if the keys are the same.

Perhaps it would be easier to use a local Dictionary instance inside of the method to populate, then set the class member to the instance after it's populated? Worst case is that the two different dictionaries are generated but they're functionally the same and have the same content. Last one overwrites the other, but the overwrite of the field reference to the dictionary is atomic.

Refit/RequestBuilderImplementation.cs
```diff
@@ -32,23 +33,80 @@ public RequestBuilderImplementation(Type targetInterface, RefitSettings refitSet
...
+            }
+
         }
 
         public IEnumerable<string> InterfaceHttpMethods => interfaceHttpMethods.Keys;
 
-        public Func<HttpClient, object[], object> BuildRestResultFuncForMethod(string methodName)
+        public Func<HttpClient, object[], object> GetHttpMethod(string key, object[] parameters = null)
+        {
+            var parameterTypes = parameters?.Select(p => p?.GetType() ?? typeof(object)).ToArray();
```
onovotny: is the `p?` necessary inside the select? Should it be an error if they put a `null` in the `Type` array?

Refit/RequestBuilderImplementation.cs
```diff
@@ -32,23 +33,80 @@ public RequestBuilderImplementation(Type targetInterface, RefitSettings refitSet
...
+            return BuildRestResultFuncForMethod(key, parameterTypes);
+        }
+
+        RestMethodInfo FindMatchingRestMethodInfo(string key, Type[] parameterTypes)
+        {
+            if (interfaceHttpMethods.TryGetValue(key, out var httpMethods)) {
+                if (parameterTypes == null) {
+                    if (httpMethods.Count > 1) {
+                        throw new ArgumentException("MethodName exists more than once, ParameterTypes mut be defined");
```
onovotny: Please specify `nameof(parameterTypes)` on the `ArgumentException`

Refit/RequestBuilderImplementation.cs
windischb: I dont't think it should be an error.
The parameterTypes variable is generated from "objects" which are set as parameters to the RestService and  i think its totally possible to set a Parameter to null...

onovotny: This can't be based on the runtime types, it'd need to be based on the types of the method parameters as defined?

windischb: Ok, i understand!
Then i think we get get rid of "GetHttpMethod"...
The only thing this method has done was to build the parameterTypes Array.
I'll change it in the code, please check if thats a better approach!

onovotny:
Looks better, but there's one additional thing I think it needs -- it needs a "mangling" of parameter types in the arguments array to a key in the request method dictionary.

My concern is to avoid doing a search for the matching method on each invocation. I'd much prefer that it generate a key name like `Method+Type1_Type2_Type3` where that becomes the key on the ` `RequestBuilderImplementation` and then what's what the implementation uses in the mustache file to avoid doing the `FindMatchingRestMethodInfo` each time.

windischb:
generate a key name like `Method+Type1_Type2_Type3`  is a litte a problem with generics, but i take a look if i found a solution for this...

windischb:
Its not the prettiest solution, but does exactly what it should do...

windischb:
Or maybe we should use something like a cache for the methods?
Don't run the BuildRestResultFuncForMethod for each method in the constructor, instead run it at first need, and store it in the Dictionary for later?

onovotny:
Thanks!

PR #422, opened 2018-01-29
Add support for generic methods based on https://github.com/joelweiss…

onovotny:
Implement support for generic methods on refit interfaces. Fixes https://github.com/paulcbetts/refit/issues/156

Based on the work @joelweiss did in https://github.com/joelweiss/refit/commit/81cdaef82e73a58d20b5a06f7f6b7596b0e85d54


onovotny:
While this compiles, I'm currently hitting reflection issues trying to make it work, at least for return types:

Getting an exception on this line:
https://github.com/paulcbetts/refit/blob/5c18e90f7aa8902b6fa7e87821689163deceee8d/Refit/RequestBuilderImplementation.cs#L127

`System.InvalidOperationException : Late bound operations cannot be performed on types or methods for which ContainsGenericParameters is true`

Putting this aside for now, but maybe someone can pick it up.

onovotny:
@joelweiss would appreciate any help on this if you're still interested in it.

joelweiss:
@onovotny I don't know how (or if I can) commit to this PR.

Check out the updated [branch](https://github.com/joelweiss/refit/tree/generic-methods) (based on you progress) it fixes the failing test.

Should I open a new PR with my changes?

And I don't understand how the tests are set up, can you please add 2 tests (I suspect at least test 2 will fail)

1. a generic method that returns a Task `Task Foo<T>(T value);`
2. a call to the same service method with a different generic argument i.e. call [this](https://github.com/joelweiss/refit/blob/generic-methods/Refit.Tests/RestService.cs#L657) with an int argument.

onovotny:
@joelweiss Thanks for this! I pulled in your commit and added a couple more tests (that fail, as you suspected).

InterfaceStubGenerator.Core/GeneratedInterfaceStubTemplate.mustache
```diff
@@ -63,7 +63,7 @@ namespace {{Namespace}}
         {
 {{#IsRefitMethod}}    
             var arguments = new object[] { {{ArgumentList}} };
-            var func = methodImpls.GetOrAdd("{{Name}}({{ArgumentListWithTypes}})", _ => requestBuilder.BuildRestResultFuncForMethod("{{Name}}", new Type[] { {{ArgumentTypesList}} }{{#MethodTypeParameterList}}, new Type[] { {{.}} }{{/MethodTypeParameterList}}));
+            var func = methodImpls.GetOrAdd("{{Name}}{{#MethodTypeParameters}}<{{.}}>{{/MethodTypeParameters}}({{ArgumentListWithTypes}})", _ => requestBuilder.BuildRestResultFuncForMethod("{{Name}}", new Type[] { {{ArgumentTypesList}} }{{#MethodTypeParameterList}}, new Type[] { {{.}} }{{/MethodTypeParameterList}}));
```
joelweiss: Yeah, this won't work since it will still use the same key for all generic arguments. 😢 

I will take another look soon.

joelweiss:
I added a commit to make argument types as part of the key, but there is still an error `'Response status code does not indicate success: 404 (No matching mock handler).'` I'm not sure weather it's an issue with the test or the code.

onovotny:
@joelweiss I fixed the test. Nice idea to use interpolation and `typeof(T).Name` as the key; do you think it should be `Name`, `FullName` or `AssemblyQualifiedFullName` to avoid potential conflicts?

Lastly, and I realize this may not be possible, but is there any way for the builder to create an open generic func so that we don't need one per combination of types? If so, wonder what the perf tradeoff is vs the current approach.

My feeling in all reality is that there aren't likely to be huge numbers of different combinations used at runtime.

joelweiss:
@onovotny I agree `AssemblyQualifiedFullName` is the way to go.

I will take a look and see whether an open generic func is an option.

joelweiss:
@onovotny I added [commit](https://github.com/joelweiss/refit/commit/916db48a60fd04a88c769227067253d52a1cc1ec) with `AssemblyQualifiedFullName`.

I don't see a way to return an open generic func, and based on this [comment](https://github.com/paulcbetts/refit/blob/45b3ce87d3057796da816072ddfa816719a4615c/Refit/RequestBuilderImplementation.cs#L127) it seems the author though so as well.

onovotny:
Thanks, this is a really great feature to finally support!

joelweiss:
If the legendary @onovotny asks for your help, you just do it!

Thanks for all you OSS work.

PR #424, opened 2018-01-29
Allow adding content headers to empty POST requests

### LINKED ISSUE #112: Can't send a POST request without Body
There's an issue where a lack of post body does not receive the proper content type header. I've set the the header Content-Type: application/x-www-form-urlencoded; charset=UTF-8, but when I send a post request containing only headers the server complains about a bad request content type, expecting application/x-www-form-urlencoded.

I can get around this by including a [Body(BodySerializationMethod.UrlEncoded)] with an empty dictionary, for example. However, I don't think this should be necessary.

bennor:
Fixes #112.

This will also work for other request methods that  `HttpClient` allows a body on.


Refit/RequestBuilderImplementation.cs
```diff
@@ -16,6 +16,11 @@ namespace Refit
 {
     partial class RequestBuilderImplementation : IRequestBuilder
     {
+        static readonly ISet<HttpMethod> bodylessMethods = new HashSet<HttpMethod>
+        {
+            HttpMethod.Get,
+            HttpMethod.Head
```
onovotny: Are there other body-less methods? How about OPTIONS?

It may be better to white-list the ones with a body? PUT/PATCH/POST? Not sure about DELETE.

@darylmiller?

bennor: I dug into the source of `HttpWebRequest`:

```csharp
    internal static KnownHttpVerb Get = new KnownHttpVerb("GET", false, true, false, false);
    internal static KnownHttpVerb Connect = new KnownHttpVerb("CONNECT", false, true, true, false);
    internal static KnownHttpVerb Head = new KnownHttpVerb("HEAD", false, true, false, true);
    internal static KnownHttpVerb Put = new KnownHttpVerb("PUT", true, false, false, false);
    internal static KnownHttpVerb Post = new KnownHttpVerb("POST", true, false, false, false);
    internal static KnownHttpVerb MkCol = new KnownHttpVerb("MKCOL", false, false, false, false);
```
The second boolean parameter is `contentBodyNotAllowed`

onovotny: Good enough for me.

bennor: I also ran the following:

```csharp
await Task.WhenAll(
	typeof(HttpMethod).GetProperties(BindingFlags.Public | BindingFlags.Static)
		.Select(p => (HttpMethod)p.GetValue(null))
		.Select(async method =>
		{
			var request = new HttpRequestMessage
			{
				RequestUri = new Uri("http://httpbin.org/anything"),
				Method = method,
				Content = new StringContent("blah")
			};
			request.Content.Headers.TryAddWithoutValidation("Content-Type", "DERP");
			bool error = false;
			try
			{
				await new HttpClient().SendAsync(request);
			}
			catch (ProtocolViolationException)
			{
				error = true;		
			}
			return new { method.Method, error };
		})
	)
```

Output:

Method | error
-- | --
GET | True
PUT | False
POST | False
DELETE | False
HEAD | True
OPTIONS | False
TRACE | False

bennor: So yeah, I agree that it's weird that those two are the only ones it doesn't like, but that seems to be the way it is.

onovotny: We don't really have OPTIONS either but someone can easily add it. FWIW, may be worthwhile adding the rest of those methods to the Attributes.cs?

bennor: OPTIONS may make sense, but CONNECT is used to set up two way communication (usually when using SSL). It's not something that would typically be used directly for an API call, so I don't think it makes sense to add it.

I could add support for` [Options]`, but is this the right PR for it and are we sure we have a use case?

darrelmiller: If you want the spec's opinion, you can send a body with any request.  However, some methods are allowed to use the body, some are not. HEAD, GET and DELETE being the ones where the body is not allowed to mean anything.

onovotny:
LGTM, merge/squash whenever you're ready.

PR #433, opened 2018-02-02
Query string parameter fixes

### LINKED ISSUE #432: Enumerable serialization in query strings failing
We run into an issue when updating from `4.0.1` to `4.2.0`. We relied on a custom `IUrlParameterFormatter` to serialize `IEnumerable` into comma separated values. After updating to `4.2.0` we started having issues with query string parameters and it looks like `IEnumerable` is being serialized with `BuildQueryMap` probably because of the `DoNotConvertToQueryMap` being incorrect in this scenario.

Here is an example of our interface:

```csharp
private interface ISomeApi
{
    [Get("/companies")]
    Task GetCompanies(IEnumerable<Guid> ids);
}
```
When calling `apiClient.GetCompanies(new[] {Guid.NewGuid(), Guid.NewGuid()})` we get a query string like 

```
?Length=2&LongLength=2&Rank=1&SyncRoot=de802b69-a204-4df4-b5e0-7d658777d056&SyncRoot=547f8167-1e48-4204-8081-c132c13c558a&IsReadOnly=False&IsFixedSize=True&IsSynchronized=False
```

onovotny:
Should fix #432

### DIFF
--- a/Refit.Tests/RequestBuilder.cs
+++ b/Refit.Tests/RequestBuilder.cs
@@ -8,6 +8,7 @@ using System.Reflection;
 using System.Threading.Tasks;
 using System.Threading;
 using Xunit;
+using System.Collections;
 
 namespace Refit.Tests
 {
@@ -425,6 +426,9 @@ namespace Refit.Tests
 
         [Get("/foo/bar/{id}")]
         Task<string> FetchSomeStuffWithQueryFormat([Query(Format= "0.0")] int id);
+
+        [Get("/query")]
+        Task QueryWithEnumerable(IEnumerable<int> numbers);
     }
 
     interface ICancellableMethods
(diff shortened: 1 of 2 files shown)


Refit.Tests/RequestBuilder.cs
```diff
@@ -425,6 +426,9 @@ public interface IDummyHttpApi
 
         [Get("/foo/bar/{id}")]
         Task<string> FetchSomeStuffWithQueryFormat([Query(Format= "0.0")] int id);
+
+        [Get("/query")]
+        Task QueryWithEnumerable(IEnumerable<int> numbers);
```
eduardocampano: What about arrays? The behavior should be the same, right?

onovotny: I can add another specific test, but yes, it should. The changes underneath in there work on `object` as the `value`

onovotny: New test added

Refit/RequestBuilderImplementation.cs
```diff
@@ -653,20 +653,33 @@ static bool DoNotConvertToQueryMap(object value)
             if (value == null)
                 return false;
 
-            var type = value.GetType().GetTypeInfo();
+            var type = value.GetType();
 
-            if (type.IsPrimitive)
+            bool ShouldReturn() => type == typeof(string) ||
```
bennor: Does this need to be a local function? Can't you just set a boolean?

onovotny: We call the same logic twice

bennor: 👀

onovotny: The "should we keep it" logic runs once for the normal input types, then again for the `T` type if it's an enumerable.

bennor: Ah, so a local function doesn't trap the state of `type`, and when we call it later, the value has changed? Makes sense. Ignore me then. 😂

onovotny: Yeah, works like a closure except without needing the allocation of a delegate. The `type` is captured and modified by the time it executes again.

bennor: That's very cool.

PR #437, opened 2018-02-12
Using Action/Func Parameters

windischb:
### Using Action/Func Parameters

It's possible to use simple Action<T> or Func<T> as Parameters. This enables a more functional programming.

 ```csharp
public interface IActionOrFuncParameterTest
{
    [Get("")]
    Task<HttpResponseMessage> Get([Body]ActionOrFuncOptions options);
    [Get("")]
    Task<HttpResponseMessage> Get([Body]Action<ActionOrFuncOptions> options);
    [Get("")]
    Task<HttpResponseMessage> Get([Body]Func<ActionOrFuncOptions> builder);
}

public class ActionOrFuncOptions
{
    public string Text { get; set; }
    public int Number { get; set; }
    public Boolean Enabled { get; set; }
}

public class ActionOrFuncOptionsBuilder
{
    private ActionOrFuncOptions options = new ActionOrFuncOptions();
    
    public ActionOrFuncOptionsBuilder WithText(string text) {
        options.Text = text;
        return this;
    }

    public ActionOrFuncOptionsBuilder WithNumber(int number) {
        options.Number = number;
        return this;
    }

    public ActionOrFuncOptionsBuilder Enabled(bool value) {
        options.Enabled = value;
        return this;
    }

    public static implicit operator ActionOrFuncOptions(ActionOrFuncOptionsBuilder builder) {
        return builder.options;
    }
}


 ```

 Can be used like this:
```csharp

var api = RestService.For<IActionOrFuncParameterTest("http://api.example.com/");

// build optionObject for standard Method Parameter
var optionObject = new ActionOrFuncOptions();
optionObject.Text = "TextValue";
optionObject.Number = 7;
optionObject.Enabled = true;

var objectResponse = await api.Get(optionObject);


// use Action for building the Options
var actionResponse = await api.Get(options => {
                options.Text = "TextValue";
                options.Enabled = true;
                options.Number = 7;
            });


// use Func which returns the Options.
// int this case, the ActionOrFuncOptionsBuilder will be casted to ActionOrFuncOptions
var funcResponse = await api.Get(() => new ActionOrFuncOptionsBuilder()
                                            .WithNumber(7)
                                            .WithText("TextValue")
                                            .Enabled(true));

```


bennor:
Can you give a real world example of where this would be useful? I'm struggling to think of a situation where you couldn't just call your function and pass in the return value.

windischb:
Of course you can!
For me it's more a practical thing...
This is more or less a shortcut for building objects from scratch and pass afterwards.
Many libraries use Fluent Api for instanciate a object.
In the library which i currently create i wan't to use fluent api to create the parameters.
There are many RestServices and RestMethods with many different options.
With that modification its easier and much faster, instead of create each parameter object myself.

But yes, you are right, the result is the same as just call a function und pass the return value...

totht91:
Sorry but your example makes no sense. If you want to use your fancypants object builder fluent API, you will be able to easily use it without refit support. 

In only one case you need refit support: 
```
var actionResponse = await api.Get(options => {
                options.Text = "TextValue";
                options.Enabled = true;
                options.Number = 7;
            });
```
This also makes no sense because people create APIs like this when the function's parameter created outside their scope. I think you are missmatching this with the configuration API's in ASP.NET Core where the options object comes from the framework and you can tweek it with a function. In this case the Get logic instentiate the options object but why do the Get method want to do that? It's your responsible, not the Get merhods.

bennor:
I hate to do this, but I just can't see any real world value it provides, so I'm closing this.

We really appreciate the contribution, but new features all add complexity to the code and are something we'll have to maintain, so they need to provide enough value to justify them.

totht91:
It's an open source best practice that you should raise an issue first then create a pull request if the issue/feature request was confirmed.

bennor:
@totht91 That's true, but not everybody knows that. I wonder if there's something we could do to make it clearer people should chat about any new features before implementing them?

Cheesebaron:
@bennor provide a Pull Request template which mentions exactly this? I.e. reject PR's when there is no linked issue.

PR #445, opened 2018-02-21
Fix bool query params

### LINKED ISSUE #443: Boolean not serializing in Refit 4.3
Hello!
Bool type fields  are not serialized in the query.
This problem appeared in Refit version 4.3. All worked in version 4.2.

_Example_
```
[Get("/myobject")]
Task<List<MyObject>> GetMyObjectsAsync([AliasAs("admin")] bool? admin);
```

bennor:
Fixes #443


onovotny:
@bennor are there any other primitives we need to consider? Should we use `Type.IsPrimitive`?

bennor:
I couldn't see any others that aren't covered by `IFormattable`, but if it's safe to use `Typ.IsPrimative` we could use that instead of explicitly checking.

felipetofoli:
Hi,

First of all, thanks for this fix.
Do you know when this fix will be released?

Regards,

PR #451, opened 2018-03-10
Clarify AliasAs use during Response serialization

### LINKED ISSUE #450: AliasAs does not appear to work on responses
Hi all,

Working on a unit test repro for this but while I familiarize myself I figured I'd drop this here.

I've got an API that looks like this (irrelevant pieces omitted)

```csharp
public interface IMyApi
{
    Get("/Organisations?brief=false&skip={skip}&top={top}")]
    Task<List<Organisation>> GetOrganizations(int skip, int top);
}
```

```csharp
public class Organisation
{
    public string ORGANISATION_NAME { get; set; }
}
```

I can retrieve the information accurately using this syntax, but if I want to alias the name using `AliasAs`, I run into trouble.

This works:

```csharp
public class Organisation
{
    [JsonProperty(PropertyName="ORGANISATION_NAME")]
    public string Name { get; set; }
}
```

But this does not:

```csharp
public class Organisation
{
    [AliasAs("ORGANISATION_NAME")]
    public string Name { get; set; }
}
```

When using `AliasAs`, the value is being serialized as null.

As I mentioned, I'll try to see if I can get this working via a test/pr.

SeanKilleen:
Resolves #450.

## Progress

* [x] Write passing test to verify that `JsonProperty` works
* [x] Write failing test to verify that `AliasAs` does not work
* [x] ~~...fix it I guess?~~ Nope. Expected behavior
* [x] Update tests to clarify expected behavior
* [x] Update `README` to clarify expected behavior


SeanKilleen:
@paulcbetts @onovotny am I correct in my preliminary understanding that any use of `AliasAs` in the response serialization would actually need to be taken into account somewhere in `RequestBuilderImplementation`, such as `BuildRestResultFuncForMethod`? 

I think I'm starting to grok this little by little but still drinking from a fire hose.

SeanKilleen:
After discussion on #451, it's clear to me that `AliasAs` is not intended to be used on the response. So, rather than trying to fix it, I'll: 

* Codify this with a unit test so someone in the future can understand that it's intended and
* Attempt to add clarifying text to the README.

SeanKilleen:
`README` updated & tests clarified. I think this is ready for review. 

@bennor, could I ask for your 👀 on this when you get the chance? Thanks!

bennor:
@SeanKilleen This is a really great PR BTW. ✨

PR #454, opened 2018-03-20
Collection formatting

vitalyal:
PR for issue https://github.com/paulcbetts/refit/issues/93

### DIFF
--- a/README.md
+++ b/README.md
@@ -127,7 +127,7 @@ Search(new [] {10, 20, 30})
 Task Search([Query(CollectionFormat.Csv)]int[] ages);
 
 Search(new [] {10, 20, 30})
->>> "/users/list?ages=10%2C2%2C30"
+>>> "/users/list?ages=10%2C20%2C30"
 ```
 
 ### Body content
--- a/README.md
+++ b/README.md
@@ -112,6 +112,23 @@ GroupListWithAttribute(4, params)
 
 A similar behavior exists if using a Dictionary, but without the advantages of the `AliasAs` attributes and of course no intellisense and/or type safety.
 
+### Collections as Querystring parameters
+
+Use the `Query` attribute to specify format in which collections should be formatted in query string
+
+```csharp
+[Get("/users/list")]
+Task Search([Query(CollectionFormat.Multi)]int[] ages);
+
+Search(new [] {10, 20, 30})
+>>> "/users/list?ages=10&ages=20&ages=30"
+
+[Get("/users/list")]
+Task Search([Query(CollectionFormat.Csv)]int[] ages);
+
+Search(new [] {10, 20, 30})
+>>> "/users/list?ages=10%2C2%2C30"
+```
 
 ### Body content
 
--- a/Refit.Tests/RequestBuilder.cs
+++ b/Refit.Tests/RequestBuilder.cs
@@ -453,6 +453,9 @@ namespace Refit.Tests
 
         [Get("/query")]
         Task QueryWithArray(int[] numbers);
+
+        [Get("/query")]
+        Task QueryWithArrayFormattedAsMulti([Query(CollectionFormat.Multi)]int[] numbers);
     }
 
     interface ICancellableMethods
@@ -1068,6 +1071,19 @@ namespace Refit.Tests
 
             var uri = new Uri(new Uri("http://api"), output.RequestUri);
             Assert.Equal("/query?numbers=1%2C2%2C3", uri.PathAndQuery);
+        }
+
+        [Fact]
+        public void QueryStringWithArrayFormattedAsMulti()
+        {
+            var settings = new RefitSettings { UrlParameterFormatter = new TestEnumerableUrlParameterFormatter() };
+            var fixture = new RequestBuilderImplementation(typeof(IDummyHttpApi), settings);
+
+            var factory = fixture.BuildRequestFactoryForMethod("QueryWithArrayFormattedAsMulti");
+            var output = factory(new object[] { new int[] { 1, 2, 3 } });
+
+            var uri = new Uri(new Uri("http://api"), output.RequestUri);
+            Assert.Equal("/query?numbers=1&numbers=2&numbers=3", uri.PathAndQuery);
         }
 
         [Fact]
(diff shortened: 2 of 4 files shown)


Refit/Attributes.cs
```diff
@@ -192,6 +192,42 @@ public AuthorizeAttribute(string scheme = "Bearer")
...
+
+        /// <summary>
+        /// Multiple parameter instances
+        /// </summary>
+        Multi,
+
+        /// <summary>
+        /// Values formatted wiht RefitSettings.UrlParameterFormatter
+        /// </summary>
+        RefitParameterFormatter
```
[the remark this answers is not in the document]
vitalyal: fixed

Refit/RequestBuilderImplementation.cs
```diff
@@ -491,6 +491,32 @@ void AddMultipartItem(MultipartFormDataContent multiPartContent, string fileName
...
+                                    case CollectionFormat.Ssv:
+                                    case CollectionFormat.Tsv:
+                                    case CollectionFormat.Pipes:
+                                        var delimiter = attr.CollectionFormat == CollectionFormat.Csv ?  ","
+                                            : attr.CollectionFormat == CollectionFormat.Ssv ? " "
+                                            : attr.CollectionFormat == CollectionFormat.Tsv ? "\t" : "|";
+
+                                        queryParamsToAdd.Add(new KeyValuePair<string, string>(
+                                            restMethod.QueryParameterMap[i],
+                                            string.Join(delimiter, paramValues.Cast<object>())));
```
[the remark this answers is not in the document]
vitalyal: fixed

Refit.Tests/RequestBuilder.cs
```diff
@@ -1070,6 +1079,55 @@ public void QueryStringWithArrayCanBeFormatted()
             Assert.Equal("/query?numbers=1%2C2%2C3", uri.PathAndQuery);
         }
 
+        [Fact]
+        public void QueryStringWithArrayFormattedAsMulti()
```
bennor: This and the following two tests could be rewritten as a `[Theory]`, which would make it simpler for you to test the remaining option. 

We're still missing tests for pipe and tab separated values.

vitalyal: fixed

Refit.Tests/RequestBuilder.cs
```diff
@@ -1079,41 +1085,23 @@ public void QueryStringWithArrayCanBeFormatted()
...
         }
 
-        [Fact]
-        public void QueryStringWithArrayFormattedAsMulti()
+        [Theory]
+        [InlineData("QueryWithArrayFormattedAsMulti", "/query?numbers=1&numbers=2&numbers=3")]
+        [InlineData("QueryWithArrayFormattedAsCsv", "/query?numbers=1%2C2%2C3")]
+        [InlineData("QueryWithArrayFormattedAsSsv", "/query?numbers=1%202%203")]
+        [InlineData("QueryWithArrayFormattedAsTsv", "/query?numbers=1%092%093")]
+        [InlineData("QueryWithArrayFormattedAsPipes", "/query?numbers=1%7C2%7C3")]
```
onovotny: Can you please add one more that covers the default case calling `QueryWithArray`? This can help prevent a regression in the future.

vitalyal: test case would look like this 
[InlineData("QueryWithArray", "/query?numbers=System.Int32%5B%5D")]
i'm not sure if this behaviour is used by anybody so i have doubts if it should be added. So please confirm that this is what you think makes sense

README.md
```diff
@@ -112,6 +112,23 @@ GroupListWithAttribute(4, params)
...
+Task Search([Query(CollectionFormat.Multi)]int[] ages);
+
+Search(new [] {10, 20, 30})
+>>> "/users/list?ages=10&ages=20&ages=30"
+
+[Get("/users/list")]
+Task Search([Query(CollectionFormat.Csv)]int[] ages);
+
+Search(new [] {10, 20, 30})
+>>> "/users/list?ages=10%2C2%2C30"
```
joelweiss: Missing a zero, No?
"/users/list?ages=10%2C2**0**%2C30"

vitalyal: thanks, fixed

Refit.Tests/RequestBuilder.cs
vitalyal: @onovotny just in case you missed previous comment. Please confirm that adding the following test case makes sense 
[InlineData("QueryWithArray", "/query?numbers=System.Int32%5B%5D")]

onovotny: @vitalyal that test does seem strange. Instead, how about one that ensures that the url parameter formatter is called with the expected value in that scenario? Basically a test with two values -- nothing, default, and `RefitParameterFormatter` that show the execution calls an `IUrlParmateterFormatter`. You can use a mock one for verifying the flow.

vitalyal: @onovotny i'm not sure i understand what you suggest, but QueryStringWithArrayCanBeFormatted test may be covering that already. That is the test that was added in scope of other feature and does some regression for approach with RefitParameterFormatter (which is default value for Query attribute)

vitalyal: @onovotny in case you missed previous comment, could you please describe in more details what kind of test do you expect

onovotny: Sorry, just been swamped this week.

onovotny:
Thanks!

vitalyal:
Thanks! when could we expect these changes published in Nuget? we'll need to plan removing workaround in our application accordingly.

kaspars-ozols:
@paulcbetts is there any chance that this PR could be published to Nuget any time soon?

andrewgunn:
Any news on updating NuGet? FYI [AppVeyor](https://appveyor.com) can do this automagically for you.

vitalyal:
@onovotny @bennor seems like quite a few people are looking forward on getting this available in NuGet. Are there any expectations when it could happen?

Thanks in advance!

Refit.Tests/RequestBuilder.cs
ahmedalejo: > @onovotny just in case you missed previous comment. Please confirm that adding the following test case makes sense
> [InlineData("QueryWithArray", "/query?numbers=System.Int32%5B%5D")]

How would i achieve the below which weabapi complient(default FromUri): 
```csharp
[InlineData("QueryWithArray", "/query?search=MyTags&tags=1&tags=2&tags=3")]
//with
class SearchQuery
{
    public string Search {get;set;}
    public int[] Tags {get;set;}
}
```

PR #467, opened 2018-04-23
Move PushStreamContent Namespace

khalidabuhakmeh:
Previous behavior caused issues of mbiguous reference around
PushStreamContent when both Refit and System.Net.Http.Formatting are
included in a project. This is essentially a show stopper to upgrading
to the latest Refit.

The solution is to change the internal implementation of PushStreamContent
namespace from System.Net.Http -> Refit.Net.Http and to update
the usages internally. Note this may be a breaking change to those who
have taken a dependency on PushStreamContent in their own projects from
the Refit assemblies.

### DIFF
--- a/Refit/PushStreamContent.cs
+++ b/Refit/PushStreamContent.cs
@@ -23,11 +23,13 @@ using System;
 using System.Diagnostics.CodeAnalysis;
 using System.Diagnostics.Contracts;
 using System.IO;
+using System.Net;
 using System.Net.Http.Headers;
 using System.Threading;
 using System.Threading.Tasks;
+using System.Net.Http;
 
-namespace System.Net.Http
+namespace Refit.Net.Http
 {
     /// <summary>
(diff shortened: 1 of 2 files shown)


Cheesebaron:
@bennor System.Net.Formatting is part of ASP.NET and does not seem to be .NET Standard: https://github.com/aspnet/AspNetWebStack/blob/master/src/System.Net.Http.Formatting/System.Net.Http.Formatting.csproj

It is not even released as a NuGet.

bennor:
In that case we don't have a lot of choice, do we. 😂

Refit/PushStreamContent.cs
```diff
@@ -23,11 +23,13 @@
...
 using System.Diagnostics.Contracts;
 using System.IO;
+using System.Net;
 using System.Net.Http.Headers;
 using System.Threading;
 using System.Threading.Tasks;
+using System.Net.Http;
 
-namespace System.Net.Http
+namespace Refit.Net.Http
```
onovotny: Why change the namespace? Why not just make the types internal?

PR #469, opened 2018-04-26
Exclude private getters from querystring & form posts

### LINKED ISSUE #462: Public property with private getter is present in URL
Hello everybody,

I don't know if this is an issue or a misunderstanding of Refit but here it is : 
For some calls to my web service I have an object to be serialized as parameters in the URL. The object has some public properties with public getters and some public properties with private getters. The problem is, when sending a request, the public properties, with private getters are sent in the URL.

Here is the object:
``` C#
public class Page
{
    public int Number { get; set; }
    public int Size { get; set; }
    public string Sort
    {
        get
        {
            return SortMember + "," + (SortDirection == SortDirection.ASC ? "asc" : "desc");
        }
    }

    public string SortMember { private get; set; }
    public SortDirection SortDirection { private get; set; }
} 
```

Here is the call in the Refit interface:
``` C#
[Get("")]
Task<Response<Company>> FindAll(string socialReason, Page page);
```

You can see in this snippet from Fiddler the properties in the URL : 
capture

bennor:
The way we were reflecting over objects for inclusion in querystrings and form posts was including public properties with inaccessible getters.

Now we make sure that the getter of the property is public.

Fixes #462

### DIFF
--- a/Refit.Tests/FormValueDictionaryTests.cs
+++ b/Refit.Tests/FormValueDictionaryTests.cs
@@ -56,6 +56,31 @@ namespace Refit.Tests
         }
 
         [Fact]
+        public void ExcludesPropertiesWithInaccessibleGetters()
+        {
+            var source = new ClassWithInaccessibleGetters
+            {
+                A = "Foo",
+                B = "Bar"
+            };
+            var expected = new Dictionary<string, string>
+            {
+                { "C", "FooBar" }
+            };
+
+            var actual = new FormValueDictionary(source, settings);
+
+            Assert.Equal(expected, actual);
+        }
+
+        public class ClassWithInaccessibleGetters
+        {
+            public string A { internal get; set; }
+            public string B { private get; set; }
+            public string C => A + B;
+        }
+
+        [Fact]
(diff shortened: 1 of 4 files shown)


bennor:
@onovotny So something that might need a bit of thought -- I removed the explicit binding flags in one of the `GetProperties()` calls, because I mistakenly thought the default was `BindingFlags.Public | BindingFlags.Instance`. It's not. It's just `BindingFlags.Public`, so you get static properties back.

Now I've realised that, my first thought was to change them both to _public instance_, but maybe it's desirable to allow static? A potential example of a property you might want to make static and set once is an API key or something. On the other hand, there are potentially [better ways to handle that use case](https://github.com/paulcbetts/refit/issues/438#issuecomment-372556847)?

onovotny:
I'm not crazy about including statics, but changing it at this point would be a breaking change.

bennor:
It would only be a breaking change for one of the places (form posts). The query string already had the correct binding flags applied. 

I think it's probably okay. The fact that statics worked for form posts was undocumented and not really that logical either. I'd be surprised if anyone was intentionally using them.

PR #488, opened 2018-05-25
Add .Net Framework 4.0.3 as a supported target

thecontrarycat:
With the Multi-targeting capability of the new VS2017 project tooling, it is much easier to add backwards compatibility to projects.  This opens refit up to an even wider array of consumers (we need it for calling APIs from WPF kiosk applications running on Windows XP Embedded (yes, I know!) )

This PR adds support to refit for .Net Framework 4.0.3 by adding conditional compilation shims where appropriate and adding .Net 4.0.3 specific package references for async and HttpClient.  

Note that any project using refit will still need to be built on a machine with net462 so that the build tasks are correctly called.

### DIFF
--- a/Refit/FormValueDictionary.cs
+++ b/Refit/FormValueDictionary.cs
@@ -74,7 +74,7 @@ namespace Refit
         PropertyInfo[] GetProperties(Type type)
         {
             return type.GetProperties(BindingFlags.Instance | BindingFlags.Public)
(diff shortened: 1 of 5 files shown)


thecontrarycat:
Looks like your build agent doesn't have the Net403 Targeting Pack installed on it :(

onovotny:
We will not be supporting .NET below 4.5.2 since those are all out of support by Microsoft and are likely vulnerable to exploits. 

Sorry.

Issue #489, opened 2018-05-28
Add support for HttpClient MultipartItem


mjalil:
Hi Refit members,
I need to be able add `HttpContent` MultipartItem as used in [this library](https://github.com/jlucansky/Camunda.Api.Client/blob/master/Camunda.Api.Client/UserTask/IUserTaskRestService.cs#L114) :

```csharp
        [Post("/task/{id}/localVariables/{varName}/data"), Multipart]
        Task SetBinaryLocalVariable(string id, string varName, BinaryDataContent data, ValueTypeContent valueType);
```
In the example above `data` and `valueType` are inherited instances of `HttpContent`
[that library](https://github.com/jlucansky/Camunda.Api.Client) without support for `HttpContent` as MultipartItem falls back to Json serialization and causes exception. see issue [Camunda.Api.Client/#10](https://github.com/jlucansky/Camunda.Api.Client/issues/10)

onovotny:
Merged in, thanks.

PR #490, opened 2018-05-28
add support for HttpContent MultipartItem (issue #489)

mjalil:


mjalil:
I think I should write a unit test to prove that it works

mjalil:
I think I should write a unit test to prove that it works

mjalil:
Oh ! My tests pass in my PC! whats up with VSTS?

Issue #491, opened 2018-05-30
Using Refit with ASP.Net Core 2.1's HttpClientFactory


thecontrarycat:
The ASP.Net Core team have created a new way of working with HttpClient: the HttpClientFactory https://github.com/aspnet/HttpClientFactory

The idea behind `HttpClientFactory` is that you no longer need to worry about keeping singleton `HttpClient` instances around.  Instead, you can get as many instances of `HttpClient` as you like from the factory, and the underlying `HttpClientHandler` will have their lifecycles managed internally.  This ensures that the issues with `HttpClient` that added the singleton requirement such as socket exhaustion are solved, while the issues with using a singleton (such as failing to respond to DNS changes due to long-lived connections) are also handled.

Currently, to use `HttpClientFactory` with Refit you can simply new up an instance of your service with `RestService.For<T>(HttpClient)`.  However, this can lead to a performance reduction due to the way methods are cached (in a dictionary on the generated service class).

To make this work more smoothly, I propose moving the method caching out of the generated service class and adding it as a decorator around the `RequestBuilderImplementation`.  This then makes the `IRequestBuilder` separate from the generated class, enabling us to provide a new overload to `RestService.For<T>(HttpClient, IRequestBuilder)`, where the `IRequestBuilder` can be created by the caller (using `RequestBuilder.For<T>()` and kept around as a singleton.

I have the PR ready to go if you would like this feature.

thecontrarycat:
Working commit: https://github.com/thecontrarycat/refit/commit/bf8b0f97c7fb9b9bfd6b2a118283cc8ff5432c7d

bennor:
I think this is a good idea in theory, but do you have any measurements for the performance improvement?

thecontrarycat:
All I've done is moved the existing caching to a different place so that people who aren't using a singleton HttpClient can benefit from it in the same way that those with a singleton HttpClient currently do.

I presume the existing caching was put there for a reason :)

bennor:
I'm not sure when the caching was added or why. My only real concern is that we have an outstanding bug relating to concurrency (#449), and I'm not sure if that has anything to do with the existing caching. (I haven't had time to look at it yet - it may not even exist in the latest version.)

bennor:
Other than a few minor nitpicks, the working commit looks good. I'd be happy to take the PR. Thoughts, @onovotny?

onovotny:
Certainly happy to take a PR to utilize `IHttpClientFactory`

[...]

PR #494, opened 2018-06-02
Modify interface to support HttpClientFactory users

thecontrarycat:
PR for #491


InterfaceStubGenerator.Core/GeneratedInterfaceStubTemplate.mustache
```diff
@@ -63,7 +62,7 @@ namespace {{Namespace}}
         {
 {{#IsRefitMethod}}    
             var arguments = new object[] { {{ArgumentList}} };
-            var func = methodImpls.GetOrAdd({{#MethodTypeParameterNames}}${{/MethodTypeParameterNames}}"{{Name}}{{#MethodTypeParameterNames}}<{{.}}>{{/MethodTypeParameterNames}}({{ArgumentListWithTypes}})", _ => requestBuilder.BuildRestResultFuncForMethod("{{Name}}", new Type[] { {{ArgumentTypesList}} }{{#MethodTypeParameterList}}, new Type[] { {{.}} }{{/MethodTypeParameterList}}));
+            var func = requestBuilder.BuildRestResultFuncForMethod("{{Name}}", new Type[] { {{ArgumentTypesList}} }{{#MethodTypeParameterList}}, new Type[] { {{.}} }{{/MethodTypeParameterList}});
```
onovotny: We can't do this....this means we're running potentially expensive reflection code on every method invocation and not caching the result.

I'd need to see a benchmark comparing the two before removing this cache.

onovotny:
I may have spoken a little too soon in my code review comment as I was going top-down and didn't see the new cache.

Given the structural changes, can you please describe what you changed and how it's working & caching things?

Refit/CachedRequestBuilderImplementation.cs
```diff
@@ -0,0 +1,58 @@
...
+        }
+
+        string GetArgumentString(Type[] parameterTypes)
+        {
+            if (parameterTypes == null || parameterTypes.Length == 0)
+            {
+                return "";
+            }
+
+            return parameterTypes.Select(t => t.Name).Aggregate((s1, s2) => s1 + ", " + s2);
```
bennor: `string.Join(", ", parameterTypes.Select(t => t.Name)` would be a little clearer here. There's another one like this in the next method too.

Refit/RequestBuilderImplementation.cs
```diff
@@ -27,7 +27,7 @@ partial class RequestBuilderImplementation : IRequestBuilder
         readonly ConcurrentDictionary<CloseGenericMethodKey, RestMethodInfo> interfaceGenericHttpMethods;
         readonly JsonSerializer serializer;
         readonly RefitSettings settings;
-        readonly Type targetType;
+        public Type targetType { get; }
```
onovotny: Since this is a property, change to PascalCase

thecontrarycat: Good spot.  Fixing now

onovotny:
Haven't forgotten about this. Will try to get this reviewed and merged in the next few days.

onovotny:
@thecontrarycat This looks good, thanks! One thing that would be helpful though is a doc update that shows how to use this with `IHttpClientFactory`. Can you please add that and I'll complete the merge?

thecontrarycat:
@onovotny I've added a note to the documentation showing how you can use it with HttpClientFactory

onovotny:
@thecontrarycat Thanks, that makes sense.

In light of the pattern in the docs, what would we need to expose a single extension method (or two with overloads) to add Refit? Something like `.AddRefit<T>(...)` with the settings, or whatever?

thecontrarycat:
I think it would be complicated to do a single `IServiceCollection.AddRefit<T>()` method because we'd need to implement the whole `HttpClient` configuration part (there are additional lines you can put between the `AddHttpClient` and `AddTypedClient` methods to customise the handlers used by the client).

I did start thinking that we could make `IRequestHandler` become `IRequestHandler<T>` and we could then replace the `AddTypedClient` call with an `AddRefitClient<T>` call where we could singleton register the appropriate `IRequestHandler<T>` in the same call.  

Refactoring the IRequestHandler was a larger job than I wanted to do unless you think it makes sense.

thecontrarycat:
I did that refactor.  `IRequestBuilder` and `IRequestBuilderInternal` have become `IRequestBuilder<T>` and `IRequestBuilder`, with the generic version being used during creation.  This allows us to have a single extension method to configure Refit in the `HttpClientFactory` world (see updated docs).

Refit/HttpClientFactoryExtensions.cs
```diff
@@ -0,0 +1,17 @@
+﻿#if NETSTANDARD2_0
+using Microsoft.Extensions.DependencyInjection;
+
+namespace Refit
+{
+    public static class HttpClientFactoryExtensions
+    {
+        public static IHttpClientBuilder AddRefitClient<T>(this IServiceCollection services) where T : class
```
onovotny: I think we need an overload that can specify the refit settings. Can either be as another parameter or as an `Action<RefitSettings>` where the method news it up and passes it to the user for setting?

onovotny:
This looks really great, I love having an easy one-liner to add in the startup that "does the right thing"

thecontrarycat:
Added the overload for `RefitSettings`.  It's not an `Action<RefitSettings>` because the `IRequestBuilder<T>` will be a singleton and therefore there's no ability to change the value at runtime.

onovotny:
Thank for this!

onovotny:
@thecontrarycat I know I already merged it, but any chance you could add a test or two that covers what the extension method is supposed to do and any tests that cover the new behavior?

PR #497, opened 2018-06-03
Don't specify modifer on auto generated class.

### LINKED ISSUE #480: Add support to force autogenerated classes to be internal
We started to use Refit, which is really nice, however, it also adds auto generated clients in the NuGet package. Is there a way to exclude it from the assembly, and thus the NuGet package? Since it clobbers the assembly's namespace.

We're using Refit in both a webservice client package, which we distribute as a NuGet package internally. The target framework is the full .NET Framework. And so are the consumer projects that have the package installed, they also have Refit installed, obviously. In the client, I instantiate the clients using `RestService.For<T>(hostUrl)`, so I would expect that the generated clients would only be needed in the consumer projects, and not in the NuGet package, or does Refit not work like that?

onovotny:
Fixes #480 by removing the access modifier on the generated classes. Given that the generated classes are partial, a public modifier could be added if desired by adding a partial class.

This is technically a breaking change, but no one should be directly accessing the auto generated types anyway.


tiesmaster:
Thanks @onovotny !

Issue #459, opened 2018-03-29
EnumMember value not used when serializing


dnasir:
The library works great when I'm retrieving data from my REST API, and so far it manages to correctly map all of my enums. However, when I need to use the same enums in my requests as querystring parameters, Refit doesn't seem to use the `EnumMember` value set in my data contracts.

Example:
```csharp
[DataContract]
[JsonConverter(typeof(StringEnumConverter))]
public enum ServiceType
{
    [EnumMember(Value = "ondemand")]
    Ondemand = 1,

    [EnumMember(Value = "flex")]
    Flex = 2,

    [EnumMember(Value = "carpool")]
    Carpool = 3
}
```

When I debug this, the request sent out would use the enum property names rather than the `EnumMember` values. i.e. `https://api-endpoint/services?type=Flex` instead of `https://api-endpoint/services?type=flex`. This causes the endpoint to return an exception.

I managed to solve this by implementing my own `IUrlParameterFormatter`, like so:

```csharp
public class EnumAsEnumMemberValueParameterFormatter : IUrlParameterFormatter
{
    public string Format(object parameterValue, ParameterInfo parameterInfo)
    {
        if(parameterValue == null)
            return null;

        var type = parameterValue.GetType();
        if(!type.IsEnum)
            return parameterValue.ToString();

        var enumValue = type.GetMember(parameterValue.ToString());
        if(enumValue.Length == 0)
            return null;

        var attr = enumValue[0].GetCustomAttribute<EnumMemberAttribute>();
        return attr != null ? attr.Value : enumValue.ToString();
    }
}
```

But I feel like this is something that should be taken care of by either Refit or the Json.NET library. Maybe I missed something.

dnasir:
I found that the following code works just as well:

```csharp
public class JsonConvertParameterFormatter : IUrlParameterFormatter
{
    public string Format(object parameterValue, ParameterInfo parameterInfo)
    {
        return parameterValue == null ? null : JsonConvert.SerializeObject(parameterValue).Replace("\"", "");
    }
}
```

I'm not sure why `JsonConvert.SerializeObject()` would add double quotes around the output, and I'm not that well-versed in Json.NET to figure it out. Maybe someone here can give me some pointers regarding this.

bennor:
@dnasir We don't have support for the `EnumMemberAttribute` yet. 

[...]

PR #502, opened 2018-06-13
Use EnumMember attribute for enum values (if defined)

Styxxy:
Fix for #459


Styxxy:
@bennor @onovotny 
The PR is now ready for review.

Refit/RefitSettings.cs
```diff
@@ -38,32 +40,55 @@ public interface IFormUrlEncodedParameterFormatter
...
                                            ? "{0}"
                                            : $"{{0:{formatString}}}",
-                                       parameterValue);
+                                       enummember?.Value ?? parameterValue);
         }
     }
 
     public class DefaultFormUrlEncodedParameterFormatter : IFormUrlEncodedParameterFormatter
     {
+        static readonly ConcurrentDictionary<Type, ConcurrentDictionary<string, EnumMemberAttribute>> enumMemberCache
```
onovotny: What is the benefit of having this static? The instance is tied to the setting object, which is tied to the interface lifetime.

Styxxy: static isn't bound to any instance. The static member is thus for the lifetime of the AppDomain. (https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/keywords/static)

PR #513, opened 2018-06-26
Add inheritdoc to generated constructor to prevent xml build warnings

njannink:
When building a project with Refit api clients with xml documentation generation on you get build warnings for missing xml comments on the generated constructor. Instead of turning build warnings off which might lead to undocumented client code fix the warning by adding <inheritdoc/> to the generated constructor aswell.


onovotny:
Thanks!

Issue #176, opened 2015-08-24
Allow: Aggregated interfaces


ahmedalejo:
Currently, if an API interface **A**(HasRefitHttpMethodAttribute) is an aggregation of other interfaces(HasRefitHttpMethodAttribute) then the generated stub for **A** doesn´t implement the aggregated interfaces;

``` csharp
using Refit
...
interface InterfaceA : InterfaceB
{
    [Get("/ping")]
    Task Ping(); 
}

interface InterfaceB
{
    [Get("/ping")]
    Task Pong(); 
}
```

The generated stub look like this, and doesn´t obviously compile

``` csharp
using Refit
...
class AutoGeneratedA : InterfaceA
{
    ...
    public Task Ping() { ... }
}

class AutoGeneratedB : InterfaceB
{
    ...
    public Task Pong() { ... }
}
```

This should/could have been the following

``` csharp
using Refit
...
class AutoGeneratedA  : AutoGeneratedB, InterfaceA
{
    ...
    public Task Ping() { ... }
}

class AutoGeneratedB : InterfaceB
{
    ...
    public Task Pong() { ... }
}
```

I´m trying to seggregate the interface to better know what client a particular ViewModel(of my MVVM) uses or is supposed to use, i use a IoC container to inject specific implementations.

The registration uses reflection:

``` csharp
using System;
using System.Linq;
using System.Net.Http;
using System.Net.Http.Headers;
using System.Reflection;
using Refit;

namespace Service.Clients
{
    using Core.Common;
    using IoC;

    public class RestServicesWrapper
    {
        static RestServicesWrapper()
        {
            RestServiceType = typeof(RestService);

            var parameterTypes = new Type[] { typeof(HttpClient) };
            RestServiceFactory = RestServiceType.GetRuntimeMethod("For", parameterTypes);
            WebApiClientsAssembly = typeof(RestServicesWrapper).GetTypeInfo().Assembly; ;
        }

        private static MethodInfo RestServiceFactory { get; set; }
        private static Type RestServiceType { get; set; }
        private static Assembly WebApiClientsAssembly { get; set; }

        public static object For(Type serviceContractType, HttpClient httpClient)
        {
            return RestServiceFactory
                .MakeGenericMethod(serviceContractType)
                .Invoke(RestServiceType, new object[] { httpClient });
        }

        public static void RegisterServices()
        {
            var httpClient = new HttpClient();

            var services =
                from service in WebApiClientsAssembly.ExportedTypes
[...]

PR #523, opened 2018-07-13
#176: Adding support for aggregated/interhited interfaces

iss0:
Like explained in issue #176 it was not possible to do something like the following

```c#
public interface IAmInterfaceA
{
    [Get("/ping")]
    Task<string> Ping();
}

public interface IAmInterfaceB
{
   [Get("/pong")]
   Task<string> Pong();
}

public interface IAmInterfaceC : IAmInterfaceB, IAmInterfaceA
{
   [Get("/pang")]
   Task<string> Pang();
}
```

I created a simple solution which makes it possible to have inheritance like in the example.


onovotny:
Please add some integration tests with the http mocks as well.

iss0:
I've added an integration test and found an issue which I've fixed already. Sorry for not coming up with this in the first place. Is there anything else which is needed from your perspective?

InterfaceStubGenerator.Core/InterfaceStubGenerator.cs
```diff
@@ -169,9 +169,20 @@ public TemplateInformation GenerateTemplateInfoForInterfaceList(List<InterfaceDe
...
             };
 
+            AddInheritedMethods(ret);
+
             return ret;
         }
 
+        private static void AddInheritedMethods(TemplateInformation ret)
+        {
+            foreach (var c in ret.ClassList.Where(c => c.BaseClassNames != null && c.BaseClassNames.Any()))
```
onovotny: Does this logic handle two interfaces with the same name but in different namespaces?

onovotny:
Also, what if the interfaces have the same method name but different return types? 

I think that the code gen should be updated to do explicit interface implementations as that's probably the only way to ensure correctness?

onovotny:
Not yet because the questions I have had not been answered yet. It's not clear yet they won't cause issues.

wocar:
thank you for your fast response, how can i build it to test it, i have a scenario where i need this feature

onovotny:
It should be directly buildable in Visual Studio 2017.

wocar:
couldn't manage to do it, removed the inheritance from my code. hopefully this scenario is supported soon.

Switched from restease to refit because dynamic code generation isn't supported on iOS.

Thank you for your support.

Sergio0694:
This looks like a very useful feature, is there an ETA on when this will be merged, or is there a preview package that includes this code that we can try out? 😄

onovotny:
@Sergio0694 this PR needs additional validation tests and potential fixes as I describe here: https://github.com/reactiveui/refit/pull/523#issuecomment-406367710

Sergio0694:
Hey @iss0 - are you still working on this? This pull request looks great so far!

@onovotny I have a question, why isn't it okay if the build just fails in case someone tries to inherit from two interfaces that have two methods with the same name and different signature?
I mean, I don't see why someone would want to do that in the first place, and then, wouldn't a build error still be fine at that point? I mean, the lib right now already fails to build whenever you try to inherit from a single interface, so this would still be miles better already.
This is an honest question, I don't see the point in blocking a merge request that works and in adding even more complexity just to handle a rare edge case that devs should just avoid to begin 🤔

Thanks!

iss0:
Hi @Sergio0694! Actually I totally forgot about this. I can see @onovotny points and I'm planning to work on it soon. Hope to find the time in next week.

Sergio0694:
Hi @iss0 - thank you so much for your quick reply and for your help!
Glad to hear you plan to implement what was left here, I can't wait for this to be integrated into Refit 😄

Another great upside to having inherited interfaces is that it'll be possible to write extension methods for the base interfaces, to be able to perform some custom post processing of the data returned from the APIs in use. Right now you either have to setup a custom JSON converter every time (for each interface) or to put the additional code somewhere else. Instead with this it could all be hidden in an extension method that worked on all the derived interfaces.

I guess this could be improved even more once C# 8.0 comes out, by just integrating the additional code directly in the base interfaces using base interface implementations. This would make this new Refit feature even more powerful by then 🎉

Sergio0694:
Hey @iss0 - sorry to bother, any updates on this?
I'd be willing to help if you need, if you give me some guidance on what it is exactly that's left to add to this PR. Thanks again! 😊

PR #524, opened 2018-07-13
fix #512 refit interface outside namespace

### LINKED ISSUE #512: GenerateStubsTask throws NullReferenceException if service interface is outside namespace
I accidentally created a RestService interface outside of a namespace and it results in the following message ```.nuget\packages\refit\4.6.16\build\netstandard1.4\refit.targets(29,5): error : Object reference not set to an instance of an object. [C:\temp\refitbug\refitbug.csproj]```

Here is a gist with a .NET Core 2.1 console app: https://gist.github.com/Rassi/39e1e0e50e291ceae8742d15a3b025a1

Styxxy:
fixes #512 : build error when (refit) interface is outside any namespace

If no namespace is defined, the auto generated class will be placed in a namespace "AutoGenerated"+interfacename.


onovotny:
Please add an integration test using the http mocks. There's quite a bit that happens via reflection at runtime that we need to ensure work.

Styxxy:
@onovotny Good catch. Added it some integration tests (and fixed a bug I found because of that).

Issue #537, opened 2018-08-01
ApiResponse<T> tries to deserealize response even for non-success codes


Veikedo:
I don't know if it is by design, but it makes impossible to deserialize error in response - because the RequestBuilderImplementation,BuildCancellableTaskFuncForMethod already closed the response stream.

With such a behavior this issue #531 can't be done using ApiResponse<T> as suggested [here](https://github.com/reactiveui/refit/issues/531#issuecomment-408284094)

btw it's not so cool to get the error using try-catch (see the code)

    public static async Task<TError> GetErrorAsync<TResponse, TError>(this ApiResponse<TResponse> apiResponse)
    {
      if (!apiResponse.IsSuccessStatusCode)
      {
        try
        {
          await apiResponse.EnsureSuccessStatusCodeAsync().ConfigureAwait(false);
        }
        catch (ApiException e)
        {
          var errorAsync = e.GetContentAs<TError>();
          return errorAsync;
        }
      }

      return default(TError);
    }

Veikedo:
create PR #538

onovotny:
I've fixed this, but in a slightly different way. You can now get the `ApiException` directly as the `Error` property on the `ApiResponse`.

PR #538, opened 2018-08-01
fix early response stream closing.

Veikedo:
Resolves this https://github.com/reactiveui/refit/issues/537
Not quite sure about this since it leaves stream open.


Veikedo:
@onovotny sorry, it blocks us a bit, any thoughts on this?

SergeyKharchenko:
@onovotny +1

PR #546, opened 2018-08-15
Format URL query argument name

quexy:
Allow user-defined formatting of query arguments.

Edit: I know it's not argument (=value) but property (=name), but that is already used for value formatting. It should be the other way around, alas renaming them would be a breaking change.


quexy:
So. The names really bugged me, so I renamed `UrlParameterFormatter` to `UrlArgumentValueFormatter`, and `UrlArgumentFormatter` to `UrlParameterNameFormatter`. For backwards compatibility I added the previous `UrlParameterFormatter`, a descendant of `UrlArgumentValueFormatter`, as `Obsolete` back to the pool.

quexy:
It was the missing generated files.

quexy:
If something is wrong with this PR, please let me know, and I'd be happy to fix it.

onovotny:
My concern here is that formatting parameter names doesn't feel like something that should be a global setting. It seems like that should be an attribute on the parameter. 

The reason the values are global is that it seems more likely you'd want to do something for "all doubles".

@bennor thoughts?

quexy:
Well, you could say that swapping values and names, and tha argument would still be valid. If, for example, I want to format all arrays as `name[]=value1&name[]=value2`, then this method would be the most simple, less intrusive way to do that.

quexy:
Since no action was taken, there is no discussion about why it should or should not be merged, and the owners could not decide if they want anything with these changes or not, I close this PR.

PR #549, opened 2018-08-20
fix: prevent addrefitclient to add same client names

trenoncourt:
addrefitclient was using addhttpclient with only name of the type used. name is not unique for generic types, so i've added a type extensions to get a unique friendly name
to reproduce: 
```c#
services
.AddRefitClient<ISharePointList<Domain1>>().ConfigureHttpClient(...).AddHttpMessageHandler<Handler1>().Services
.AddRefitClient<ISharePointList<Domain2>>().ConfigureHttpClient(...).Services
```
before the fix, `Handler1` is intercepting these two clients. after the fix, `Handler1` is intercepting only the first client.


onovotny:
Thank you. Can you please add some tests around this?

onovotny:
Fixed by #561

Issue #526, opened 2018-07-19
HttpClientFactoryExtensions.AddRefitClient<T> bug for generic interfaces


iss0:
I think I've found a bug when using the extension method Refit.HttpClientFactoryExtensions.AddRefitClient<T> to initialize a generic interface.

Given the following interface:
```cs
public interface ICrudApi<T>
{
    [Post("")]
    Task<T> Create([Body]T payload);

    [Get("/{id}")]
    Task<T> ReadOne(Guid id);

    (....)
}
```
calling the extension method like:
```cs
 services
  .AddRefitClient<ICrudApi<User>>()
  .ConfigureHttpClient(c => c.BaseAddress = "https://localhost:5000/v1/user");
 services
  .AddRefitClient<ICrudApi<Role>>()
  .ConfigureHttpClient(c => c.BaseAddress = "https://localhost:5000/v1/role");
```
results in having every ICrudApi<T> using the same base address which is wrong for all but the last one added to the service collection.

This is because within the extension method the AddHttpClient(string name) method is called with the parameter typeof(T).Name. In my example typeof(T).Name equals to "ICrudApi`1" which is equal for all calls to AddRefitClient with a ICrudApi<T> so all of them using a single HttpClient with the same base address.

Correct me if I'm using the whole thing wrong or what, but from my point of view the name used for the HttpClient should take generic interfaces into account.

onovotny:
Sounds like a bug.

The name logic should probably be the same as https://github.com/reactiveui/refit/blob/master/Refit/RestService.cs#L60-L76 to ensure there's a 1:1 match.

digitalfuller:
I'll create a PR for this.

PR #561, opened 2018-10-05
Ensure Refit clients can be registered in the DI container for HttpClientFactory.

digitalfuller:
Address #526 

I extracted the logic for generating the unique type name into an internal class and shared it between the RestService and HttpClientFactoryExtensions.
I wasn't sure if you would like that class to be exposed public or not, so I made it internal and shared using InternalsVisibleTo.

I attempted to test the HttpClientFactoryExtensions by resolving the services, but there is no obvious way to examine the HttpClient properties for the service, so I kept it simple and just tested that the names were different.


onovotny:
Thanks!

PR #564, opened 2018-10-08
add nuget & myget badges

### LINKED ISSUE #563: Nuget & MyGet badges
Many projects have badges with nuget version (and myget if applicable).
Maybe interesting to add here as well?

eg with https://shields.io/

Styxxy:
fixes #563


onovotny:
thanks!

Issue #566, opened 2018-10-11
Add support for custom serializers


stevewgh:
I have a requirement to integrate with an API which uses XML instead of JSON.

I noticed that a similar issue was raised (#83) a few years back but it was closed and the branch was deleted before the feature was added.

I'd like to add support for custom serializers as I would find it useful for my particular use-case but before I start looking at adding the support, is this something that the project maintainers would like adding?

xljiulang:
[WebApiClient](https://github.com/dotnetcore/WebApiClient) support commit with xml content:
```c#
// PUT webapi/user  
// Body user的xml文本
[HttpPut("/webapi/user")]
ITask<UserInfo> UpdateUserWithXmlAsync([XmlContent] UserInfo user);
```

benjaminhowarth1:
I like this. +1 from me

stevewgh:
@benjaminhowarth1 well that makes at least two of us so I'll start taking a look to see what would be involved.

Cheesebaron:
@benjaminhowarth1 @stevewgh please, use the reaction to +1. Adding a comment about +1 doesn't contribute anything to the discussion.

Currently the `BodySerializationMethod` isn't super extensible by the user. I guess we would need to find a better way to provide your own.

So maybe providing a serialization factory instead of having it be an enum would be an option.

I guess this factory should provide:
- Content-Type
- Serialized content as `HttpContent`

Similarly this should be done for deserilization where we should add some way to extend which deserializer should be used. This currently seems very hardly tied to this line of code: https://github.com/reactiveui/refit/blob/master/Refit/RequestBuilderImplementation.cs#L293

stevewgh:
@benjaminhowarth1  @Cheesebaron pull request #571 covers the isolation of serialization and provides XML and JSON support. 

There are questions on how the ```BodySerializationMethod``` enum and other things mentioned in the PR would still work given the approach I've taken - so any feedback is welcome.

onovotny:
The PR has been merged. Will bake on the MyGet feed for a bit and then be published to nuget. Thanks!

stevewgh:
Hi @onovotny I've noticed an issue with the version published to MyGet which is related to the changes I've made. I think I've isolated the problem but I want to be able to recreate the issue in a test before submitting a fix. I'd recommend not publishing to nuget just yet.

stevewgh:
Fixed with #590

richbryant:
[...]

PR #571, opened 2018-10-18
Isolated Content Serialization

stevewgh:
My first effort at adding support for custom serializers #566 . I'm interested in hearing what you think, getting feedback etc. Thanks.

### Overview of Changes 
- Isolated all content serialization to the IContentSerializer interface
- Added JsonContentSerializer and XmlContentSerializer implmentations
- RefitSettings allows the IContentSerializer to be specified but still defaults to JsonContentSerializer for back-compat reasons
- .NET Standard 1.4 has a new package dependency on System.Xml.XmlSerializer

### Tests
- Tests modified to exercise Json and Xml serialization
- XmlContentSerializer unit test added
- Some existing tests have been modified to use both JsonContentSerializer and XmlContentSerializer using the an xunit theory

### How to use
```c#
// 1. uses Json serialization - no change
var service = RestService.For<IService>("https://localhost");

// 2. uses Json serialization and passes the JsonSerializerSettings to the JsonContentSerializer
var service = RestService.For<IService>("https://localhost",
    new RefitSettings
    {
        JsonSerializerSettings = new JsonSerializerSettings {Formatting = Formatting.Indented}
    });

// 3. equivalent of 2. 
var service = RestService.For<IService>("https://localhost",
    new RefitSettings
    {
        ContentSerializer = new JsonContentSerializer(new JsonSerializerSettings {Formatting = Formatting.Indented})
    });

// 4. uses Xml serialization
var service = RestService.For<IService>("https://localhost",
    new RefitSettings
    {
        ContentSerializer = new XmlContentSerializer()
    });

// 5. uses Xml serialization with additional configuration
var service = RestService.For<IService>("https://localhost",
    new RefitSettings
    {
        ContentSerializer = new XmlContentSerializer(
            new XmlContentSerializerSettings
            {
                XmlReaderWriterSettings = new XmlReaderWriterSettings(
                    new XmlReaderSettings
                    {
                        ConformanceLevel = ConformanceLevel.Document
                    })
            })
    });
```

### Possible further changes
- Should the ```RefitSettings.JsonSerializerSettings``` property as marked as ```[Obsolete]```? It should only be used with the JsonContentSerializer.
- Simplify how the ContentSerializer property is set. One idea is to add a ```RefitSettings``` extension class - something like this:

```c#
public static class RefitSettingsJsonContentExtensions
{
    public static RefitSettings UseJsonSerialization(this RefitSettings settings, JsonSerializerSettings jsonSerializerSettings)
    {
        settings.ContentSerializer = new JsonContentSerializer(jsonSerializerSettings);
        return settings;
    }
}
```
The extension class would be used like this:
```c#
var service = RestService.For<IService>("https://localhost", new RefitSettings().UseJsonSerialization(new JsonSerializerSettings {Formatting = Formatting.Indented}));
```


onovotny:
Thanks for this!

> Should the RefitSettings.JsonSerializerSettings property as marked as [Obsolete]? It should only be used with the JsonContentSerializer.

Yes, I would say so, but with a warning, not an error, and the message should point to the new type and docs.

I'm not keen on using an extension method to configure a new instance of the settings object -- @paulcbetts any thoughts?

stevewgh:
Glad to help! 

I've rebased my branch on master and made some further changes. 

- Marked RefitSettings.JsonSerializerSettings as obsolete
- Marked the BodySerializationMethod.Json enum as obsolete, suggest using the newly added BodySerializationMethod.Serialized instead
- Updated the README document with examples of how to use Refit with XML

I hope this is OK? 

Thanks.

Refit/RefitSettings.cs
```diff
@@ -13,21 +12,44 @@ namespace Refit
...
         public IUrlParameterFormatter UrlParameterFormatter { get; set; }
         public IFormUrlEncodedParameterFormatter FormUrlEncodedParameterFormatter { get; set; }
         public bool Buffered { get; set; } = true;
     }
 
+    public interface IContentSerializer
+    {
+        HttpContent Serialize(object item);
+
+        Task<object> DeserializeAsync(HttpContent content, Type objectType);
```
onovotny: Can this be `Task<T> DeserializeAsync<T>(HttpContent content)` to improve type safety?

onovotny: Also, why do we need both a sync & async deserialize? if we need it, do we also need an async serialize?

Refit.Tests/MultipartTests.cs
```diff
@@ -429,7 +429,7 @@ public async Task MultipartUploadShouldWorkWithAnObject(Type contentSerializerTy
                     Assert.Equal("theObject", parts[0].Headers.ContentDisposition.Name);
                     Assert.Null(parts[0].Headers.ContentDisposition.FileName);
                     Assert.Equal(mediaType, parts[0].Headers.ContentType.MediaType);
-                    var result0 = serializer.Deserialize<ModelObject>(await parts[0].ReadAsStringAsync());
+                    var result0 = await serializer.DeserializeAsync<ModelObject>(parts[0]);
```
onovotny: needs `ConfigureAwait(false)`

Refit/ValidationApiException.cs
```diff
@@ -27,7 +27,6 @@ public static ValidationApiException Create(ApiException exception)
         /// <summary>
         /// The problem details of the RFC 7807 validation exception.
         /// </summary>
-        public new ProblemDetails Content => GetContentAs<ProblemDetails>();
-
+        public new ProblemDetails Content => GetContentAsAsync<ProblemDetails>().ConfigureAwait(false).GetAwaiter().GetResult();
```
onovotny: This can deadlock. Async calls need to be awaited, thus cannot really be a property.

Refit/RequestBuilderImplementation.cs
```diff
@@ -574,12 +575,12 @@ void AddMultipartItem(MultipartFormDataContent multiPartContent, string fileName
                     {
                         foreach (var item in enumerable)
                         {
-                            AddMultipartItem(multiPartContent, itemName, parameterName, item);
+                            await AddMultipartItemAsync(multiPartContent, itemName, parameterName, item);
                         }
```
onovotny: needs `ConfigureAwait(false)`

Refit/RequestBuilderImplementation.cs
```diff
@@ -574,12 +575,12 @@ void AddMultipartItem(MultipartFormDataContent multiPartContent, string fileName
...
                     }
                     else
                     {
-                        AddMultipartItem(multiPartContent, itemName, parameterName, itemValue);
+                        await AddMultipartItemAsync(multiPartContent, itemName, parameterName, itemValue);
```
onovotny: needs `ConfigureAwait(false)`

Refit/RequestBuilderImplementation.cs
```diff
@@ -574,12 +575,12 @@ void AddMultipartItem(MultipartFormDataContent multiPartContent, string fileName
                     {
                         foreach (var item in enumerable)
                         {
-                            AddMultipartItem(multiPartContent, itemName, parameterName, item);
+                            await AddMultipartItemAsync(multiPartContent, itemName, parameterName, item);
```
onovotny: needs `ConfigureAwait(false)`

Refit/RequestBuilderImplementation.cs
```diff
@@ -670,7 +671,7 @@ void AddMultipartItem(MultipartFormDataContent multiPartContent, string fileName
                     throw new InvalidOperationException("BaseAddress must be set on the HttpClient instance");
 
                 var factory = BuildRequestFactoryForMethod(restMethod, client.BaseAddress.AbsolutePath, restMethod.CancellationToken != null);
-                var rq = factory(paramList);
+                var rq = await factory(paramList);
```
onovotny: needs `ConfigureAwait(false)`

Refit/RequestBuilderImplementation.cs
```diff
@@ -212,23 +214,23 @@ void AddMultipartItem(MultipartFormDataContent multiPartContent, string fileName
...
+        Func<HttpClient, CancellationToken, object[], Task<T>> BuildCancellableTaskFuncForMethod<T, TBody>(RestMethodInfo restMethod)
         {
             return async (client, ct, paramList) =>
             {
                 if (client.BaseAddress == null)
                     throw new InvalidOperationException("BaseAddress must be set on the HttpClient instance");
 
                 var factory = BuildRequestFactoryForMethod(restMethod, client.BaseAddress.AbsolutePath, restMethod.CancellationToken != null);
-                var rq = factory(paramList);
+                var rq = await factory(paramList);
```
onovotny: needs `ConfigureAwait(false)`

stevewgh:
Good spot with the the missing ConfigureAwait's, I'll add them in.

The ValidationApiException.Content property is a bit trickier. If I change this factory method to `async Task<ValidationApiException>`
https://github.com/reactiveui/refit/blob/c14b2c265e3ebb1f72915ec44e8ea73ef136fed4/Refit/ValidationApiException.cs#L22-L25

I could then set the Content property from within the `Create` method and the property would avoid deadlocks.

We already have an async context where the `ValidationApiException.Create` is called, so I think this could work.

Refit/ValidationApiException.cs
```diff
@@ -14,19 +15,24 @@ public class ValidationApiException : ApiException
...
         /// <param name="exception">An instance of an ApiException to use to build a ValidationException.</param>
         /// <returns>ValidationApiException</returns>
-        public static ValidationApiException Create(ApiException exception)
+        public static async Task<ValidationApiException> Create(ApiException exception)
+#pragma warning restore VSTHRD200
         {
-            return new ValidationApiException(exception);
+            return new ValidationApiException(exception)
+            {
+                Content = await exception.GetContentAsAsync<ProblemDetails>()
```
onovotny: Needs `ConfigureAwait(false)`

Refit/RefitSettings.cs
```diff
@@ -13,21 +12,42 @@ namespace Refit
...
+
+        public IContentSerializer ContentSerializer { get; set; }
         public IUrlParameterFormatter UrlParameterFormatter { get; set; }
         public IFormUrlEncodedParameterFormatter FormUrlEncodedParameterFormatter { get; set; }
         public bool Buffered { get; set; } = true;
     }
 
+    public interface IContentSerializer
+    {
+        Task<HttpContent> SerializeAsync(object item);
```
onovotny: Can we make this generic, so it's `SerializeAsync<T>(T item)` that could prevent a box and allows for a custom serializer that's optimized for a particular type.

onovotny:
Looks good! One last thing and then I think it can be merged.

Refit/RefitSettings.cs
stevewgh: That was an easier change than I was expecting! The method swapped over to a generic with no issues.

stevewgh:
Hi @onovotny are the changes that I made in https://github.com/reactiveui/refit/pull/571/commits/cc05c3a9feea80200309f40beed9cbe4b304883f  ok?

richbryant:
This still doesn't make it easy to use one's own ContentSerializer, since the IContentSerializer interface is buried in RefitSettings.  I'm using Blazor, Json.NET is not supported under WASM, I can easily do my own deserialization based on returning HttpResponseMessage but posting is proving more than a little evil.

If I could simply throw in something that implements IContentSerializer, it'd be easy.  But I can't.  Why not?

stevewgh:
Hi @richbryant if you want to implement your own IContentSerializer then you can. Here's one which sends everything as Base64 strings and uses a non-existent (I think!) media type of text/base64

```cs
public class Base64ContentSerializer : IContentSerializer
{
    BinaryFormatter formatter = new BinaryFormatter();

    public async Task<T> DeserializeAsync<T>(HttpContent content)
    {
        var bytes = Convert.FromBase64String(await content.ReadAsStringAsync());

        using (var ms = new MemoryStream(bytes))
        {
            return (T)formatter.Deserialize(ms);
        }
    }

    public Task<HttpContent> SerializeAsync<T>(T item)
    {
        using (var stream = new MemoryStream())
        {
            formatter.Serialize(stream, item);
            var content = (HttpContent)new StringContent(Convert.ToBase64String(stream.ToArray()), System.Text.Encoding.UTF8, "text/base64");
            return Task.FromResult(content);
        }
    }
}

// use it like this

var client = RestService.For<IApi>("http://localhost:5000",
    new RefitSettings
    {
        ContentSerializer = new Base64ContentSerializer()
    });
```

The `IContentSerializer` interface is available in versions 4.6.68 onwards which is only available on the 'MyGet' feed for now. 

@onovotny do you have an ETA for when 4.6.68 be published to NuGet?

PR #578, opened 2018-11-09
Implement RFC 7807 to support Problem Details in bad validation requests

### LINKED ISSUE #470: Support for RFC 7807 - Problem Details for HTTP APIs
ASP.NET Core 2.1 will be supporting [RFC 7807 - Problem Details for HTTP APIs](https://tools.ietf.org/html/rfc7807) (see [here](https://blogs.msdn.microsoft.com/webdev/2018/02/27/asp-net-core-2-1-web-apis/) for more discussion by the ASP.NET Core team). This defines a data structure (the RFC provides for encoding in either JSON or XML) for communicating more information about the cause of an error.

It would be good if Refit supported parsing this in its `ApiException`? It would need to identify if the response to any non-success HTTP code was of type `application/problem+json` and deserialize appropriately.

faniereynders:
Added support for handling validation-based exceptions according to the [RFC 7807](https://tools.ietf.org/html/rfc7807) specification.

A new exception `ValidationApiException` derived from `ApiException` is introduced and allows one to catch specific validation exceptions in a declarative way:

```csharp
//...
try
{
   var result = await awesomeApi.GetFooAsync("bar");
}
catch (ValidationApiException ex)
{
   //handle validation here by using ex.Content, which is type of ProblemDetails
}
catch (ApiException ex)
{
   //other exception handling
}
//...
```
Resolved #470


faniereynders:
.

paulcbetts:
Seems reasonable. @onovotny?

Refit/ProblemDetails.cs
```diff
@@ -0,0 +1,15 @@
+﻿using System.Collections.Generic;
+
+namespace Refit
+{
+    public class ProblemDetails
+    {
+        public Dictionary<string, string[]> Errors { get; set; } = new Dictionary<string, string[]>();
```
onovotny: @paulcbetts any thoughts on using the concrete type here vs an `IDictionary<string,string>` or `IReadOnlyDictionary<string, string>`?

onovotny:
@paulcbetts  LGTM with the one style question design question you.

onovotny:
@faniereynders Would you please add/update relevant documentation to cover this scenario?

faniereynders:
@onovotny certainly, are you referring to the documentation in the code itself? Please point me to the specific place, I'm new here :)

onovotny:
Hi @faniereynders there are two things:

1. Any XML doc comments for public members of types you added
2. Usage here: https://github.com/reactiveui/refit/blob/master/README.md

Please add them to this PR and we can merge them together.

Refit/ProblemDetails.cs
```diff
@@ -4,11 +4,34 @@ namespace Refit
 {
     public class ProblemDetails
```
onovotny: Please add a doc comment to the type itself

onovotny:
@faniereynders Looks like one missing doc comment and then we can merge

PR #581, opened 2018-11-11
Conflicting PreserveAttribute when using InternalsVisibleTo

### LINKED ISSUE #565: Conflicting PreserveAttribute in RefitStubs.g.cs when using InternalsVisibleTo attribute
I have two projects in a solution, my main assembly named Foo and a unit test assembly named Foo.Tests.
Both assemblies are using Refit for my particular testing needs. This was working perfect until I needed to add unit tests that needed access to an internal class in Foo. I added the InternalsVisibleTo attribute to the AssemblyInfo.cs in Foo so Foo.Tests had access.  Doing this causes this error:

`obj/Release/netcoreapp2.0/RefitStubs.g.cs(43,6): error CS0436: The type 'PreserveAttribute' in '/home/build/Tests/Foo.Tests/obj/Release/netcoreapp2.0/RefitStubs.g.cs' conflicts with the imported type 'PreserveAttribute' in 'Foo, Version=0.0.0.999, Culture=neutral, PublicKeyToken=null'. Using the type defined in '/home/build/Tests/Foo.Tests/obj/Release/netcoreapp2.0/RefitStubs.g.cs'. [/home/build/Tests/Foo.Tests/Foo.Tests.csproj]`

The problem is that when refit generates it stubs, it uses the same namespace and/or class names in both projects and then the compiler throws a warning because it doesn't know which one to use. Warnings at my company are treated as failures so I needed to fix this issue.

My current work around was to remove Refit from Foo.Tests and add all my test classes directly in Foo. Since Refit was not added to Foo.Tests, I don't have the problem.  Obviously, this is not ideal. 

One proposed solution is the ability to change the namespace of the stubs so there is not a conflict in this specific use case.

Styxxy:
fixes #565 

Using the project `$(RootNamespace)`, this will be set as a separate MSBuild property (RefitInternalNamespace), so the user can (if he wants) overrule the namespace.
Eg:

**refit.targets**
```xml
<Project>
  <!-- ... -->
  <PropertyGroup>
    <RefitInternalNamespace Condition=" '$(RefitInternalNamespace)' == '' ">$(RootNamespace)</RefitInternalNamespace>
  </PropertyGroup>

  <!-- ... -->

  <Target Name="GenerateRefitStubs" BeforeTargets="CoreCompile">
    <!-- ... -->
    
    <GenerateStubsTask
        SourceFiles="@(Compile)"
        BaseDirectory="$(MSBuildProjectDirectory)" 
        OutputFile="$(IntermediateOutputPath)\RefitStubs.g.cs"
        RefitInternalNamespace="$(RefitInternalNamespace)"
        />

    <!-- ... -->    
  </Target>
</Project>
```

We pass this _RefitInternalNamespace_ along to the `InterfaceStubGenerator` class (InterfaceStubGenerator.Core), which it will include when generating the mustache template.

I updated the App and the UnitTest project to provide the RootNamespace as well, so the feature is immediately being tested as well.


onovotny:
The `RootNamespace` is more likely to conflict again if it's repeated. What if we just use `AssemblyName` since you can't have two different assemblies with the same name in the output directory?

Styxxy:
@onovotny In projects AssemblyName and RootNamespace are by default the same. Besides, assembly name can be in a wrong format, include invalid characters etc. RootNamespace can indeed be the same for projects, for that I added an MSBuild property so the usercan override if necessary.

onovotny:
A property for overriding is good, but I think namespace is more likely to be the same than assemblyname. For assembly name, we can use `_` for any invalid characters?

Styxxy:
> 
> 
> A property for overriding is good, but I think namespace is more likely to be the same than assemblyname. For assembly name, we can use `_` for any invalid characters?

Converting an assembly name to a valid namespace is more complex than I initially thought. See also the docs ([Identifier names](https://docs.microsoft.com/en-us/dotnet/csharp/programming-guide/inside-a-program/identifier-names)). I think it is make the MSBuild target a lot more complex, and not anymore readable. By default the RootNamespace of a project is already a valid namespace (and equals (in a valid way) to the assembly name). Only when overriden, this might give a problem (in combination with internals visible & both assemblies generating the same Refit clients).

Do you still want me to try to write a "convert" _method_ within the MSBuild targets file to try to cover all cases? Especially if you think about reserved words, white space, invalid namespace characters, ... It will get really complex to write this in MSBuild targets itself.

PR #584, opened 2018-11-27
 Add support for url form encoding collections

MariusVolkhart:
When form encoding objects that have a property that is an IEnumerable type, use the QueryAttribute and it's CollectionFormat to determine how to encode the object.

 In general, this feature could be fleshed out much more deeply. Future work could include respecting custom formats and delimiters specified on the QueryAttribute, as well as support for collections stored in dictionaries that are form encoded.

### DIFF
--- a/Refit/FormValueDictionary.cs
+++ b/Refit/FormValueDictionary.cs
@@ -7,6 +7,12 @@ using Newtonsoft.Json;
 
 namespace Refit
 {
+    /// <summary>
+    /// Transforms a form source from a .NET representation to the appropriate HTTP form encoded representation.
+    /// </summary>
+    /// <remarks>Performs field renaming and value formatting as specified in <see cref="QueryAttribute"/>s and
+    /// <see cref="RefitSettings.FormUrlEncodedParameterFormatter"/>. Note that this is not a true dictionary, rather, a
+    /// form of MultiMap. A given key may appear multiple times with the same or different values.</remarks>
     class FormValueDictionary : IEnumerable<KeyValuePair<string, string>>
     {
         static readonly Dictionary<Type, PropertyInfo[]> propertyCache
@@ -54,7 +60,7 @@ namespace Refit
                             switch (attrib?.CollectionFormat) {
                                 case CollectionFormat.Multi:
                                     foreach (var item in enumerable) {
-                                        Add(fieldName, settings.FormUrlEncodedParameterFormatter.Format(item, null));
+                                        Add(fieldName, settings.FormUrlEncodedParameterFormatter.Format(item, attrib.Format));
                                     }
                                     break;
                                 case CollectionFormat.Csv:
@@ -67,7 +73,7 @@ namespace Refit
 
                                     var formattedValues = enumerable
                                         .Cast<object>()
-                                        .Select(v => settings.FormUrlEncodedParameterFormatter.Format(v, null));
+                                        .Select(v => settings.FormUrlEncodedParameterFormatter.Format(v, attrib.Format));
                                     Add(fieldName, string.Join(delimiter, formattedValues));
                                     break;
                                 default:
--- a/Refit/FormValueDictionary.cs
+++ b/Refit/FormValueDictionary.cs
@@ -3,15 +3,16 @@ using System.Collections;
 using System.Reflection;
 using System.Collections.Generic;
 using System.Linq;
-using System.Threading.Tasks;
 using Newtonsoft.Json;
 
 namespace Refit
 {
-    class FormValueDictionary : Dictionary<string, string>
+    class FormValueDictionary : IEnumerable<KeyValuePair<string, string>>
     {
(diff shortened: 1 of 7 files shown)


Refit/FormValueDictionary.cs
```diff
@@ -45,15 +45,74 @@ public FormValueDictionary(object source, RefitSettings settings)
...
+                                case CollectionFormat.Ssv:
+                                case CollectionFormat.Tsv:
+                                case CollectionFormat.Pipes:
+                                    var delimiter = attrib.CollectionFormat == CollectionFormat.Csv ?  ","
+                                        : attrib.CollectionFormat == CollectionFormat.Ssv ? " "
+                                        : attrib.CollectionFormat == CollectionFormat.Tsv ? "\t" : "|";
+
+                                    var formattedValues = enumerable
+                                        .Cast<object>()
+                                        .Select(v => settings.FormUrlEncodedParameterFormatter.Format(v, null));
```
onovotny: Is there still value providing the attribute format string? Could these be combined? For example, for a specific double formatting in addition to the collection format?

MariusVolkhart: Wasn't sure before how the format strings fit into the picture. The attribute format string should _definitely_ be respected

Refit/FormValueDictionary.cs
```diff
@@ -3,16 +3,23 @@
...
 namespace Refit
 {
-    class FormValueDictionary : Dictionary<string, string>
+    /// <summary>
+    /// Transforms a form source from a .NET representation to the appropriate HTTP form encoded representation.
+    /// </summary>
+    /// <remarks>Performs field renaming and value formatting as specified in <see cref="QueryAttribute"/>s and
+    /// <see cref="RefitSettings.FormUrlEncodedParameterFormatter"/>. Note that this is not a true dictionary, rather, a
+    /// form of MultiMap. A given key may appear multiple times with the same or different values.</remarks>
+    class FormValueDictionary : IEnumerable<KeyValuePair<string, string>>
```
MariusVolkhart: Is having this be called a `Dictionary` confusing? Should I rename it? Outside of `FormValueDictionaryTests` there is only 1 line that would be affected by a rename.

onovotny: It can be renamed as it's an internal class

Refit/FormValueDictionary.cs
```diff
@@ -45,15 +51,74 @@ public FormValueDictionary(object source, RefitSettings settings)
                     var value = property.GetValue(source, null);
                     if (value != null)
                     {
+                        var fieldName = GetFieldNameForProperty(property);
+
                         // see if there's a query attribute
                         var attrib = property.GetCustomAttribute<QueryAttribute>(true);
 
-                        Add(GetFieldNameForProperty(property), settings.FormUrlEncodedParameterFormatter.Format(value, attrib?.Format));
+                        if (value is IEnumerable enumerable) {
```
onovotny: Nit: please put braces on the next line

Refit/FormValueDictionary.cs
```diff
@@ -45,15 +51,74 @@ public FormValueDictionary(object source, RefitSettings settings)
...
+        /// Returns a key for each entry. If multiple entries share the same key, the key is returned multiple times.
+        /// </summary>
+        public IEnumerable<string> Keys => this.Select(it => it.Key);
+
+        /// <summary>
+        /// Returns the value of the first entry found with the matching key. Multiple additional entries may use the
+        /// same key, but will not be considered. Returns <c>null</c> if no matching entry is found.
+        /// </summary>
+        public string this[string key]
+        {
```
onovotny: Where is this currently used? Is there a danger in losing in this path that needs to be addressed? In other words, should this indexer be removed and coded around a different way?

MariusVolkhart: Existing tests use it. I think there is a larger than necessary risk of future misuse, and the tests can be satisfied comfortably with a LINQ expression. Fixed in c63b62c7164a70a32fa74c7dcfbd50fd4b23b534

MariusVolkhart:
Thanks for this library and for taking the time to review the CL!

ahmedalejo:
Hi @MariusVolkhart first of all thanks for the PR.

I was wandering if the PR fortunately addresses this issue as well

[QueryParameters] Are broken for classes with IEnumerable properties #587

i took a quick look into the code and i couldn´t assert that your PR would work for non-decorated query parameters. inormally use refit attributes on the Refit interface, and not on my POCOs

ahmedalejo:
An extract of the class QueryParameter class is:

```csharp
public class PostQueryOptions
{
     public PostQueryOptions()
     {
          KeywordIds = new int[0];
          MaxResults = 100;
     }

     [DataMember("tags")]
     public int[] KeywordIds { get; set; }

     [DataMember("limit")]
     public int MaxResults { get; set; }

     [DataMember("phrase")]
     public string SearchPhrase { get; set; }
}
```

Issue #610, opened 2019-02-16
Error when building URL if hostUrl has trailing /


tstivers1990:
Reproduce:

Create a new `RestService` and pass a `hostUrl` that ends with `/`. For example:

`RestService.For<IMyRestApi>("http://localhost/api/");`

Upon making a request, such as `GetObjects`, the url used for the request will look as follows:

`http://localhost/api//objects`

Proposed solution:
The solution is fairly simple. The constructor could simply check if the `hostUrl` has a trailing `/` and remove it if so.

tstivers1990:
Feel free to assign this to me. I'm currently working on it.

PR #613, opened 2019-02-16
Trim trailing / from hostUrl #610

tstivers1990:
Solves #610 

I attempted to add a unit test for this, but I was unable to find a good way to go about testing it. There doesn't seem to be a way to inject an `HttpClient` while also passing the `hostUrl` and, naturally, there is no way to access the `HttpClient` that is created from the result. At least not that I'm aware of.

If you know how to go about unit testing this change, I'd be glad to implement the unit test as well.


onovotny:
You can add a property called Client of type HttpClient on an interface and it’ll return the concrete instance. 

I’d also suggest adding a NullOrEmpty check against the url parameter.

tstivers1990:
> You can add a property called Client of type HttpClient on an interface and it’ll return the concrete instance.

Ah. I'll add the unit test shortly, then.

> I’d also suggest adding a NullOrEmpty check against the url parameter.

Yes I definitely should have used the null coalescing operator before `.TrimEnd()`. Good catch. As far as checking IsNullOrEmpty, I'm willing to do that but I think it would be more appropriate to put that in it's own issue/pull request as it's a bug in it's own right. What is the expected behavior if "baseUrl" is null or empty? Throw `InvalidArgumentException`? If this is the case, there wouldn't be a need for the null coalescing operator.

onovotny:
Yeah, I think invalid argument exception would work.

tstivers1990:
@onovotny I'm not sure why that test failed after I added the unit test. Are you? Doesn't appear I changed anything that would have caused it to fail. Can you also confirm whether or not my last commit should have been committed, or if those files should be left out of git.

PR #616, opened 2019-02-17
614 should throw argumentexception if hosturl isnullorwhitespace

### LINKED ISSUE #614: RestService.For<T>(string hostUrl) should throw InvalidArgumentException if hostUrl IsNullOrWhiteSpace

tstivers1990:
**This pull request must be merged after #615 due to changes to `Refit.Tests.InterfaceStubGenerator.GenerateTemplateInfoForInterfaceListSmokeTest`**

Closes #614


tstivers1990:
It seems `System.Uri` already throws exceptions when creating the `Uri` an invalid `string` is passed. Maybe it would be better to simply use the null coalescing operator for the `.TrimEnd()` added by #615 and let `System.Uri` handle validating the string? It simplifies the code and provides a more descriptive exception to the end user in my opinion.

tstivers1990:
Also worth noting is that if this PR is merged, it may be a breaking change. Users may have depended on the exceptions thrown by `System.Uri` and this change will break those users' code.

tstivers1990:
I've updated #615 to use the null-conditional operator so that this pull request can be made optional. Merging #615 without merging this pull request should be safe.

PR #633, opened 2019-03-07
Explicit interface inheritance

Sergio0694:
This continues the work from @iss0 in #523 on the interface inheritance for Refit services.
In particular, here I've updated the stub generator as @onovotny asked, so that Refit now does explicit interface inheritance in order to properly handle cases where multiple interfaces declare methods with the same name but different return values.


onovotny:
Looks good - one question came up as I was looking at the code: what happens if there are different, possibly conflicting attributes on the interface -- for Header, Auth, or any of the others. Which one wins and is it clear?

Sergio0694:
@onovotny I'm not sure how refit currently handles these cases, but I'd say the result here depends on the current behavior of the library, that is:
• If it throws when more than a single attribute of that kind is present, it'll do the same here
• If it just picks the first one if finds, it'll use the first one depending on how runtime sorts the attributes across the inherited interfaces (I'd assume that depends on the actual order you use when declaring them?)

If the first case is true then it should be fine, otherwise if you'd like refit to throw when more than one attribute of that kind is present, I guess that's something that isn't strickly related to this PR in particular, and should be fixed for the library as a whole (maybe with a dedicated PR?).

Let me know what you think and I'll be happy to make any changes in here as well if needed 😊

Sergio0694:
Hey @onovotny - not sure if you saw my updated reply (I've edited my comment with more info after I first replied to you), the attributes handling seems to be working fine and as expected (see my code snippet above).

Please let me know if there's anything else you need or if this PR is good to go for you, I'm looking forward for finally being able to use this in my project 😄
Thanks again for your time!

onovotny:
As this is new behavior (Refit didn't have any inheritance before), that seems okay. Please update the main docs (the readme) showing this new behavior and explaining what happens if there's multiple conflicting attributes and then this can be merged.

README.md
```diff
@@ -633,6 +633,64 @@ Which can be used like this:
 // than one type (unless you have a different domain for each type)
 var api = RestService.For<IReallyExcitingCrudApi<User, string>>("http://api.example.com/users"); 
 ``` 
+### Interface inheritance
```
onovotny: I think explaining the need in terms of different authentication needs may be confusing. Can you please use a different example?

Sergio0694: Sure thing, I've replaced that part with a more general description in https://github.com/reactiveui/refit/pull/633/commits/199dedf10c802a8c610d06917625a5ad8853987f.

PR #634, opened 2019-03-09
Fix #625: xml output encoding don't match writer setting

### LINKED ISSUE #625: xml output encoding dose not follow  XmlWriterSettings encoding  setting
`XmlWriterSettings'` encoding  setting in `XmlContentSerializer`/`XmlReaderWriterSettings`  don't take effect, serialized xml `encoding` attribute always `utf-16`
```
<?xml version="1.0" encoding="utf-16"?>
<root>
</root>
```

lhyqy5:
fix #625


Refit/XmlContentSerializer.cs
```diff
@@ -26,12 +26,14 @@ public Task<HttpContent> SerializeAsync<T>(T item)
         {
             var xmlSerializer = new XmlSerializer(item.GetType(), settings.XmlAttributeOverrides);
 
-            using (var output = new StringWriter())
+            using (var stream = new MemoryStream())
             {
-                using (var writer = XmlWriter.Create(output, settings.XmlReaderWriterSettings.WriterSettings))
+                using (var writer = XmlWriter.Create(stream, settings.XmlReaderWriterSettings.WriterSettings))
                 {
+                    var encoding = settings.XmlReaderWriterSettings.WriterSettings?.Encoding ?? Encoding.Unicode;
```
onovotny: What should the default be? Unicode or UTF8? Any reason for Unicode over UTF8 as the default?

lhyqy5: StringWriter default encoding is Unicode, so I keep it

PR #647, opened 2019-04-07
fixes #644 add cancellable content serializer

### LINKED ISSUE #644: Pass CancellationToken to IContentSerializer
I'm implementing IContentSerializer interface and in my implementation, I've no access to cancellation token. Could you please provide it too? Thanks in advance.

Styxxy:
fixes #644 add cancellable content serializer

I didn't add unit tests for now, as I don't know yet exactly how to test it. Any pointers are welcome  :).

I'll provide a separate PR for docs update (if applicable), so it can be merged after new version is being published.

### DIFF
--- a/Refit/RequestBuilderImplementation.cs
+++ b/Refit/RequestBuilderImplementation.cs
@@ -215,7 +215,17 @@ namespace Refit
             Exception e = null;
             try
             {
-                multiPartContent.Add(await settings.ContentSerializer.SerializeAsync(itemValue).ConfigureAwait(false), parameterName);
+                HttpContent cont;
+                if (settings.ContentSerializer is IContentSerializerWithCancellation serializerWithCancellation)
+                {
+                    cont = await serializerWithCancellation.SerializeAsync(itemValue).ConfigureAwait(false);
+                }
+                else
+                {
+                    cont = await settings.ContentSerializer.SerializeAsync(itemValue).ConfigureAwait(false);
+                }
+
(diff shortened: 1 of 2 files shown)


Refit/RequestBuilderImplementation.cs
```diff
@@ -215,7 +215,17 @@ async Task AddMultipartItemAsync(MultipartFormDataContent multiPartContent, stri
             Exception e = null;
             try
             {
-                multiPartContent.Add(await settings.ContentSerializer.SerializeAsync(itemValue).ConfigureAwait(false), parameterName);
+                HttpContent cont;
+                if (settings.ContentSerializer is IContentSerializerWithCancellation serializerWithCancellation)
+                {
+                    cont = await serializerWithCancellation.SerializeAsync(itemValue).ConfigureAwait(false);
```
onovotny: This call is missing a cancellation token parameter

onovotny:
For testing it, I'd suggest creating a mock content type that uses some kind of reset event (like `SepaphoreSlim(1,1)` in the serialize/deserialize and then use the passed-in cancellation token in a WaitOne call.

Refit/RequestBuilderImplementation.cs
Styxxy: Good catch, fixed it :)

clairernovotny:
closing due to age. cancellation is passed into the deserialize method and so far none of the methods to create an HttpContent have required cancellation.

PR #658, opened 2019-05-02
Support round-tripping route parameter syntax

iron9light:
Feature

**What is the new behavior?**
Round-tripping route parameter syntax is a new feature from [ASP.NET Core](https://docs.microsoft.com/en-us/aspnet/core/fundamentals/routing?view=aspnetcore-2.2#differences-from-earlier-versions-of-routing):
Forward slashes aren't encoded when using a double-asterisk (\*\*) catch-all parameter syntax.
During link generation, the routing system encodes the value captured in a double-asterisk (\*\*) catch-all parameter (for example, {**myparametername}) except the forward slashes.

Route | Param | Link
----- | ----- | ----
/search/{**page} | page = "admin/products" | /search/admin/products

**What might this PR break?**

**Other information**:

### DIFF
--- a/README.md
+++ b/README.md
@@ -87,6 +87,20 @@ Task<List<User>> GroupList([AliasAs("id")] int groupId, [AliasAs("sort")] string
 GroupList(4, "desc");
 >>> "/group/4/users?sort=desc"
 ```
+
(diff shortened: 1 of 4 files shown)


iron9light:
@anaisbetts @onovotny Could you please help to review this PR? I think it's a missing part for a REST API client library. And we do need this feature.

onovotny:
LGTM, @anaisbetts?

Issue #662, opened 2019-05-10
[BUG] Performance issue in For<T>(client, builder) and For(refitInterfaceType, client, builder) of Refit.RestService


pit-hub:
**Describe the bug**
RestService uses ConcurrentDictionary to cache pairs of <Refit REST interface, Refit Generated Client Proxy>.

ConcurrentDictionary.GetOrAdd provides the following overloads

* GetOrAdd(TKey, Func<TKey,TValue>) - Adds a key/value pair to the ConcurrentDictionary<TKey,TValue> by using the specified function if the key does not already exist. Returns the new value, or the existing value if the key exists.

* GetOrAdd(TKey, TValue) - Adds a key/value pair to the ConcurrentDictionary<TKey,TValue> if the key does not already exist. Returns the new value, or the existing value if the key exists.

The current code of **_For\<T\>(client, builder)_** and **_For(refitInterfaceType, client, builder)_** use the GetOrAdd(TKey, TValue) overload. That completely removes the usefulness of the cache.

**Steps To Reproduce**
n/a

**Expected behavior**
Regardless on number of times the **_For\<T\>(client, builder)_** or **_For(refitInterfaceType, client, builder)_** are called for a given type (T), GetGeneratedType should be called only once per type.

**Screenshots**
Current code (calculates a value on every call):
```` csharp
    public static T For<T>(HttpClient client, IRequestBuilder<T> builder)
    {
        var generatedType = TypeMapping.GetOrAdd(typeof(T), GetGeneratedType<T>());

        return (T)Activator.CreateInstance(generatedType, client, builder);
    }

    public static object For(Type refitInterfaceType, HttpClient client, IRequestBuilder builder)
    {
        var generatedType = TypeMapping.GetOrAdd(refitInterfaceType, GetGeneratedType(refitInterfaceType));

        return Activator.CreateInstance(generatedType, client, builder);
    }

````

**Environment**
n/a

**Additional context**
Here is a fix (using a delegate to calculate a value when a new entry should be added):
```` csharp
    public static T For<T>(HttpClient client, IRequestBuilder<T> builder)
    {
        var generatedType = TypeMapping.GetOrAdd(typeof(T), (type) => GetGeneratedType(type));

        return (T)Activator.CreateInstance(generatedType, client, builder);
    }

    public static object For(Type refitInterfaceType, HttpClient client, IRequestBuilder builder)
    {
[...]

PR #663, opened 2019-05-15
added RefitRestServiceAttribute for HttpClient relative path

yuri-voloshyn:
added RefitRestServiceAttribute for HttpClient relative path
fixed issue #662


Refit.Tests/RestService.cs
```diff
@@ -139,6 +139,24 @@ public interface ITrimTrailingForwardSlashApi
         Task Get();
     }
 
+    [RefitRestService("/api/Test")]
```
onovotny: Seems like this should be called `BaseAddressAttribute`?

yuri-voloshyn: I renamed attribute

pit-hub:
Hi! I've made some improvements to the work of @yuri-voloshyn.
The idea is not to modify the BaseAddress of the HttpClient. Modifying the BaseAddress may leads to unexpected behaviour by developers and also prevent reusing a HttpClient for multiple interfaces, as each instance of HttpClient uses its own pool of sockets.
Can push to the @yuri-voloshyn fork to update the PR.
The code is at: https://github.com/pit-hub/refit/tree/feature/refit-rest-service-attribute

onovotny:
Sorry, merged this by mistake. @pit-hub can you please open a PR with the combined changes?

PR #671, opened 2019-06-06
Relax Json.NET version requirements

### LINKED ISSUE #670: Be less aggressive with dependency updates in core library?
Refit currently depends on Newtonsoft.Json 12.0.1, but this is problematic in some usage scenarios where the overall dependency graph of an application may have constraints on the version of Newtonsoft.Json that can be used.

Here's a concrete example:

I have an Azure function which consumes Refit, and part of the Azure Functions SDK has a requirement to use Newtonsoft.Json 11.0.2.

This currently makes it impossible to update Refit and pick up any bug fixes or improvements in an Azure Functions app because of the dependency defined in the Azure Functions libraries.

Here's a dependabot PR that highlights this incompatibility: https://github.com/martincostello/azure-functions/pull/34

As users are able to upgrade Newtonsoft.Json themselves in their top-level application to the latest and greatest version (or not) based on their overall application requirements, it would be better if Refit's core library used the lowest-common-denominator version of Newtonsoft.Json (and any other dependencies) if no specific features in the public API surface of these dependencies are used in Refit's own code.

For example the lowest-common-denominator version of Newtonsoft.Json for .NET Standard support is 9.0.1.

In many cases applications using Refit would be ASP.NET Core apps, which would raise the Newtonsoft.Json version to 11.0.2 (see [Microsoft.AspNetCore.JsonPatch](https://www.nuget.org/packages/Microsoft.AspNetCore.JsonPatch/) NuGet package).

bennor:
This change downgrades the Newtonsoft.Json package reference (i.e. Json.NET) to a minimum of 9.01. We're not using any of the newer APIs, and all our tests pass, so I think this is okay. :rocket:

It required adding an explicit dependency on `System.Runtime.Serialization.Formatters` to the netstandard 1.4 version, and in the process of adding that I cleaned up the 1.4 specific package references in the project a tiny bit.

My only concern here is the [fairly large number of fixes in Json.NET between 9.0.1 and 12.0.2](https://github.com/JamesNK/Newtonsoft.Json/releases). This change won't stop stop anyone from referencing a later version though -- explicitly or otherwise. It just might open us up to some annoying bug reports. (As stated in the referenced issue, most people will be using a later version than 9.0.1 anyway.)

So this fixes #670, I think :joy:

-  Downgrades the `Newtonsoft.Json` package reference to a minimum of 9.01

- A dependency on `Newtonsoft.Json` 12.0.1 or higher (see #670.)

**What is the new behavior?**

- People using Azure Functions (or any other library with a strict requirement on a lower version than 12.0.1 but >= 9.0.1 will be able to use the latest version of Refit. 

**What might this PR break?**

It could expose us to bug reports for any of [the issues fixed in Json.NET since 9.0.1](https://github.com/JamesNK/Newtonsoft.Json/releases).

- [x] Tests for the changes have been added (for bug fixes / features) _Sort of. The old ones all run._
- [x] Docs have been added / updated (for bug fixes / features) _Not really applicable._

### DIFF
--- a/Refit/Refit.csproj
+++ b/Refit/Refit.csproj
@@ -8,15 +8,16 @@
   </PropertyGroup>
 
   <ItemGroup>
-    <PackageReference Include="Newtonsoft.Json" Version="12.0.1" />
-    <PackageReference Condition="'$(TargetFramework)' == 'netstandard1.4' " Include="System.Net.Http" Version="4.3.4" />
+    <PackageReference Include="Newtonsoft.Json" Version="9.0.1" />
     <Compile Remove="Features\**\*.*" />
     <None Include="Features\**\*.*" />
   </ItemGroup>
 
   <ItemGroup Condition="'$(TargetFramework)' == 'netstandard1.4' ">
     <PackageReference Include="Microsoft.AspNetCore.WebUtilities" Version="1.1.2" />
+    <PackageReference Include="System.Net.Http" Version="4.3.4" />
     <PackageReference Include="System.Reflection.TypeExtensions" Version="4.5.0" />
+    <PackageReference Include="System.Runtime.Serialization.Formatters" Version="4.3.0" />
     <PackageReference Include="System.Xml.XmlSerializer" Version="4.3.0 " />
   </ItemGroup>
   


onovotny:
I have to say, I'm not keen on using minimum versions as I tend to think that lets bugs fester. I'd venture that a significant number of people only use the transitive dependencies and don't directly reference newer versions.

Is there a specific issue with the Azure Functions runtime tracking this? I thought they had been working on something so their runtime was isolated from user code.

bennor:
Looks like there's a [functions runtime issue for it](https://github.com/Azure/azure-functions-host/issues/4049). Can't tell for certain if it's being fixed.

There are about a hundred other ways AF does my head in, so in the grand scheme of things this is pretty minor. 🤣

bennor:
@onovotny Is there any way to set up the package references so that people will be encouraged to use the latest version but still allow a previous version? Or is that just the default behaviour anyway?

If not I think we can abandon this PR and close the related issue.

onovotny:
@bennor not that I know of

Issue #197, opened 2015-11-02
Support dynamic multiple headers (vs single header) using single parameter


Korayem:
We have more than 5 headers that we set with each call and it's cumbersome to do it with current implementation because it means adding 5 arguments to EACH API call
### What I suggest is the following:

```
[Get("/users/{user}")]
Task<User> GetUser(string user, [Headers] string[] headers);

// Will add list of headers to the request:
var user = await GetUser("octocat", new []{"header1: value1","header2: value2"}); 
```

bennor:
This could (should?) also be achieved by allowing the headers value to be a dictionary.

CreepyGnome:
I know this is old, and would like to add two cents to this. I have a set of headers that have to be provided to all methods as params, and I was thining it would be really nice if I had a POCO that I can use that used the existing HeaderAttribute on properties of this POCO and you just pass the POCO as a single param. Then the Headers would be pulled from this just like as if they where parameters with the HeaderAttribute.

Also it could be useful if I could pass a single POCO that had multiple properties that could be AliasAsAttribute or HeaderAttribute. This would allow for a single "request" POCO that can be passed in and it can set headers, path replacement values, and query string values.

Just a thought, as this came up today and would have made the method definition a lot cleaner if I had a single POCO or even two, one for headers and one for everything else.

Massimo37:
I'm running into the same problem, had to add all the dynamic headers to EACH call, but would be nice if this would be achieved through a collection or through a model (POCO type).

PR #710, opened 2019-07-12
Add support for HeaderCollection

yhvicey:

Add support for specifying multiple headers dynamically in parameter. (#197)

There is no current approach for specifying multiple headers dynamically.

**What is the new behavior?**

Users can specify multiple headers dynamically by using newly added attribute `HeaderCollection`:

```cs
public interface IService
{
    [Get("/api/values")]
    Task<IEnumerable<string>> GetValuesAsync(
        [HeaderCollection]
        IDictionary<string, string> headers);
}
```

And all key-value pairs in dictionary passed in will be added to request's header.

**What might this PR break?**

To minimize the change, I used a specific string value as a placeholder: `Refit:HeaderCollection` (see implementation for more details). So if users use this specific string value as their header, it will be ignored. Besides, there's no breaking change as there's no change to existing classes.

**Other information**:

To make code more maintainable and understandable, I added a new internal static class `Constants` to hold the placeholder I mentioned above. I'm new to this project and don't know if there's anything exists for the same usage. If so, please kindly let me know and I'll update the PR.


yhvicey:
Tests and docs will be added later, and it'd be good if anyone can make sure my change won't violate any conventions or rules.

onovotny:
Overall, the general direction is good. Please ensure that there are tests where the header collection is at the beginning and in the middle of other parameters.

Also, can you find a way to do it without using a special constant?

yhvicey:
This is the easiest & minimal-modification way I can think of for now. Current process is:

1. Build a Position (int) -> Header Name (string) map from parameter list. All parameter with `Header` attribute will be picked up and added to the map.
1. Use this map to see if a parameter is in the map. If so, use the header name in the map, then use the parameter as the value.

The question here is, there's no 'Name' for a header collection, and those headers share a same header position. One possible solution is, we use a more complex header map (e.g. `IDictionary<int, HeaderInfo>`, then get headers from it.

I'll update other test cases later.

yhvicey:
I guess position test can be finished with `IRestMethodInfoTests` - and i'm glad to add it to `IDummyHttpApi` if you think it's required!

yhvicey:
Ping @onovotny - Any feedback? We are actively developing and want this feature be shipped so we won't get blocked

onovotny:
I would like to see this without a magic constant first. Thanks

yhvicey:
Done! Removed magic string and use a `HeaderInfo` instead

Refit/RequestBuilderImplementation.cs
```diff
@@ -543,7 +543,21 @@ async Task AddMultipartItemAsync(MultipartFormDataContent multiPartContent, stri
                     // if header, add to request headers
                     if (restMethod.HeaderParameterMap.ContainsKey(i))
                     {
-                        headersToAdd[restMethod.HeaderParameterMap[i]] = paramList[i]?.ToString();
+                        var headerInfo = restMethod.HeaderParameterMap[i];
+                        if (headerInfo.IsHeaderCollection)
```
onovotny: Why is this check required? Why can't it just check for the `HeaderCollection` attribute on the parameter info?

Refit/RequestBuilderImplementation.cs
```diff
@@ -543,7 +543,21 @@ async Task AddMultipartItemAsync(MultipartFormDataContent multiPartContent, stri
...
+                        {
+                            if (paramList[i] is IDictionary<string, string> headerCollection)
```
onovotny: Seems like this could check for `IEnumerable<string, string>` instead of `IDictionary<string, string>`.

Refit/RestMethodInfo.cs
```diff
@@ -11,6 +11,19 @@
 
 namespace Refit
 {
+    public struct HeaderInfo
+    {
+        public bool IsHeaderCollection { get; }
```
onovotny: As per above, I'm not sure why this is required. There's already an attribute on the parameter, that should be enough?

Refit/RequestBuilderImplementation.cs
yhvicey: Please see the explain above:

1. Build a Position (int) -> Header Name (string) map from parameter list. All parameter with Header attribute will be picked up and added to the map.
1. Use this map to see if a parameter is in the map. If so, use the header name in the map, then use the parameter as the value.

Here in the step 2, the `HeaderParameterMap` is just a `IDictionary<string, string>` and it doesnot contains any attribute info - it's been built in step 1, if we get attribute info here then it's meaning less to build such a map in step 1.

Since I'm not familiar with this project, so I'll try to minimize the change. If you think remove the first step and get attribute info in step 2 in acceptable, I can do that as well.

Refit/RequestBuilderImplementation.cs
yhvicey: There's no `IEnumerable<string, string>` but only `IEnumerable<T>`

Refit/RestMethodInfo.cs
yhvicey: Ditto

Refit/RequestBuilderImplementation.cs
onovotny: Sorry, I meant `IEnumerable<KeyValuePair<string, string>>`, which is what `IDictionary` implements.

Refit/RequestBuilderImplementation.cs
onovotny: I'd prefer to do the "correct" change rather than a workaround. If the code that builds the map needs to change to do something different, then I think that's the better approach.

yhvicey: Yes, that's what I actually did - I changed the map from `IDictionary<string, string>` to `IDictionary<string, HeaderInfo>` to handle more complex cases like `HeaderCollection`. In fact, I reviewed the code and found it's hard to get attribute info in step 2: this map is contained in a `RestMethodInfo` class which is built from C#'s native `MethodInfo`, and there's no attribute info in this `RestMethodInfo` class, so we cannot get attribute info in step 2 actually.

Refit/RequestBuilderImplementation.cs
yhvicey: Fixed

Refit/RequestBuilderImplementation.cs
yhvicey: Ping - @onovotny Any feedback? BTW, this two-steps implementation is not implemented by me, but exist here before I made the change - I'm just improving it, so it's not a **workaround**.

pandarrr:
Is this a dead PR? I need this functionality also and would be willing to finish the work here if necessary.

yhvicey:
> Is this a dead PR? I need this functionality also and would be willing to finish the work here if necessary.

I'm waiting for author's feedback but there's no response since my last commit yet. I think you can checkout my branch and pull master, then build the binary to see if it works for you, or use customized `HttpClient` to add multiple headers dynamically.

Massimo37:
@yhvicey  @pandarrr  glad to see movement on this, I've run into a blocker on my project and this would definitely be a great addition to Refit. Have the conflicts been worked on yet?

yhvicey:
> @yhvicey @pandarrr glad to see movement on this, I've run into a blocker on my project and this would definitely be a great addition to Refit. Have the conflicts been worked on yet?

Hi, this PR is staled and I haven't heard further response from repo owner yet. I guess you can workaround by using `HttpClientHandler` and pass it to Refit client.

yhvicey:
Updated my branch with all new master commits, haven't tested yet.

Massimo37:
@yhvicey Thanks for the work on this, very appreciated that you're adding support for this 👍

danleydmello:
Any update on this? we need to pass a collection of Headers dynamically. Thanks

clairernovotny:
Closing as this was implemented in #1012

PR #757, opened 2019-10-02
Objects Bound to URL Path

nhart12:

Feature
I want to get eyes on this to determine if there is any interest in getting this cleaned up and into the main branch

Currenty there is no support for binding custom objects to the url path. For the below route we need to setup the interface like so
[Post("/foos/{someProperty}/bar/{someProperty2}")]
Task PostFooBar(int someProperty, string someProperty2, [Body]object someObject);

**What is the new behavior?**

Dotnet core Web Api's support having objects properties that bind to the Route/path. This adds the ability to do the same in refit. Where someProperty and someProperty2 are properties on the PathBoundObject class.
[Post("/foos/{someProperty}/bar/{someProperty2}")]
Task PostFooBar(PathBoundObject request, [Body]object someObject);

**What might this PR break?**
I'm worried about a lot. I don't like how I had to change the IUrlParameterFormatter interface especially for dotnet standard 1.4 since there is no ICustomAttributeProvider until netstandard1.5
I'm using these code changes in my own projects but not sure about the wider audience. One thing that got a little weird is deciding which properties on the POCO bind to the path vs query string, I have some tests surrounding this. 

- [ x] Tests for the changes have been added (for bug fixes / features)

**Other information**:


onovotny:
I'm okay with dropping .NET Standard 1.4 at this point and leaving it at 2.0. My main question is syntax, around conflicts between property names and other parameters with the same name.

Maybe the syntax should be different?

Also, what about derived objects. Is it looking at properties just on that one type, or will it look at derived types (which could lead to ambiguity).

nhart12:
@onovotny I'm open to suggestions as to how to handle property names vs another parameter (currently the parameter takes precedence and the property is appended to the query string). I agree this can be confusing. Should we maybe have some attributes to aid in this? Attributes may also help making derived types less ambiguous as yes currently it will look at properties on derived objects

onovotny:
If we fully qualify the parameter names, it could solve this:

Such as `/path/{someObject.SomeProperty}/{anotherObject.SomeProperty}`

nhart12:
Ooo I do like that

nhart12:
@onovotny I am now fully qualifying the parameter names to contain the object name. I also dropped netstandard1.4

onovotny:
How's this coming along? Happy to get this in once it's ready.

nhart12:
@onovotny I think this is ready, Although it looks like I now need to circle back and resolve some conflicts with the main branch. I should have time to do this in the next few days

nhart12:
@onovotny finally had some time to update this branch and resolve conflicts if you have time to look it over

onovotny:
Given some of the earlier bugs we saw with complex objects, can you please add a few other tests and mix an object like you havze, with additional rich objects that are `[Query]` and `[Multipart]` and both.

I'd like to ensure we cover all the various combinations where complex objects may be treated differently.

nhart12:
@onovotny  Just added a few more test cases, and a fix for an issue binding query strings when multipart. Is this what you were looking for?

Refit.Tests/RestService.cs
```diff
@@ -73,15 +73,27 @@ public interface IApiBindPathToObject
 
         [Post("/foos/{request.someProperty}/bar/{request.someProperty2}")]
         Task PostFooBar(PathBoundObject request, [Body]object someObject);
+
+        [Get("/foos/{request.someProperty}/bar/{request.someProperty2}")]
+        Task GetFooBars(PathBoundObjectWithQuery request);
+
+        [Multipart]
+        [Post("/foos/{request.someProperty}/bar/{request.someProperty2}")]
+        Task<HttpResponseMessage> PostFooBarStreamPart(PathBoundObject request, StreamPart stream);
```
onovotny: Can you please add another variant of this where it has a signature like:

```
[Multipart]
        [Post("/foos/{request.someProperty}/bar/{request.someProperty2}")]
        Task<HttpResponseMessage> PostFooBarStreamPart(PathBoundObject request, [Query] ModelObject someQueryParams, StreamPart stream);
```

onovotny: That "should" put the `ModelObject` properties on the query string and the `StreamPart` should be the multipart content. If that's not happening, that'd be a bug.

nhart12: ok yeah it appears to be a bug in the existing code, should I try to fix in this PR?

onovotny: Please!

nhart12: Ok my latest commit has the test cases and the fix for this

onovotny:
thanks!

onovotny:
I just merged this, but we still do need some docs. Can you please create a new PR with an update to the README showing the new functionality?

Issue #769, opened 2019-10-22
[BUG] Refit uses wrong method/url for service if parameter types have same name but diff namespaces


xumix:
**Describe the bug**
Refit uses wrong method/url for service if parameter types have same name but diff namespaces
For example:
```
public interface IIntegrationClient
    {
        [Post("/api/externalCommands/SendContractActions")]
        Task PostSendContractActionsJob(SendContractActionsCommand command);

        [Post("/api/v2/externalCommands/SendContractActions")]
        [Headers("Authorization: Bearer")]
        Task PostSendContractActionsJob(Pp3.Contracts.Commands.V2.Contract.ExternalIntegration.SendContractActionsCommand command);
}
```

I suppose the problem is in `CachedRequestBuilderImplementation` method `GetArgumentString`. It uses type name instead of `FullName`. Same problem is also valid for `GetGenericString` method.

**Steps To Reproduce**

1. Create an interface with polymorphic methods like above
2. Use in runtime
3. Have same urls called

**Expected behavior**
Different urls are requested

PR #770, opened 2019-10-22
Fix cache key construct for CachedRequestBuilderImplementation

Bassardes:
Fix cache key generation in CachedRequestBuilderImplementation
https://github.com/reactiveui/refit/issues/769

Generic and method arguments in cache key use type name without namespace

**What is the new behavior?**
Generic and method arguments in cache key use full type name


onovotny:
Can you please add a unit test or two that shows the broken behavior and can validate this fix? It will also help prevent regressions.

PR #785, opened 2019-11-15
Simpleusing local api

NakWarsi:
this is just a sample showing how we can make our application cleaner using Refit Library,
this shows how we can share the contract(refit Interface and SDK) where a user only needs to call the interface methods.


benjaminhowarth1:
Hi @NakWarsi, I love the sample, thankyou for this work, I'll tidy up some of the variables & labels, but otherwise I'm happy to include this.

NakWarsi:
very happy to hear that,
Thank You

PR #788, opened 2019-11-18
feat: Add support for ProblemDetails extensions

Syrus666:
Updates the ProblemDetails to support extensions to the spec. This behaves the same as the MS implementation of ProblemDetails extensions: https://github.com/aspnet/Mvc/blob/master/src/Microsoft.AspNetCore.Mvc.Core/ProblemDetails.cs#L62

This means additional properties on the response will be populated in the new Extensions property.
I have chosen to not change the behaviour of the Errors property in case it's a breaking change for other users.

Extension problem details extension properties are ignored

**What is the new behavior?**
Extension properties on the problem details response are populated in to new property

**What might this PR break?**
It shouldn't cause any regressions with this being additive.

**Other information**:
If you'd like to discuss further I'm "syrus" on reactivex.slack.com

### DIFF
--- a/README.md
+++ b/README.md
@@ -843,8 +843,11 @@ try
 }
 catch (ValidationApiException validationException)
 {
-   // handle validation here by using validationException.Content, 
+   // handle validation here by using validationException.Content,
    // which is type of ProblemDetails according to RFC 7807
+
+   // If the response contains additional properties on the problem details,
+   // they will be added to the validationException.Content.Extensions collection.
 }
 catch (ApiException exception)
 {
--- a/Refit.Tests/ResponseTests.cs
+++ b/Refit.Tests/ResponseTests.cs
@@ -104,6 +104,42 @@ namespace Refit.Tests
         }
 
         [Fact]
+        public async Task WhenProblemDetailsResponseContainsExtensions_ShouldHydrateExtensions()
+        {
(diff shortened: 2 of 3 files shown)


onovotny:
This looks good. What change were you considering for `Errors`?

Syrus666:
What I'd considered as an alternative was a breaking change of changing the Errors property to be `[JsonExtensionData] IDictionary<string, object>` rather than `Dictionary<string, string[]>` and using that as **the** property to deserialise other properties in same namespace rather than introducing a new one.

Outside of that being a naughty breaking change you could then have scenarios where the problem details responses they were currently getting were like

```json
{
    "type": "type",
    "title": "title",
    "detail": "detail",
    "status": 500,
    "errors": { 
        "foo":[ "bar" ] 
    },
    "baz": 123
}
```

Which would mean even forcibly casting the values of the Errors dictionary to an array to make their code compile again would cause problems for them as post this PR Errors would've also now contained baz.

So I thought it safest to stick with the new property.

Or I could be over thinking it and no-one uses the Errors property for anything :D

onovotny:
Thanks for the explanation. I've merged this, but it would still be helpful for any doc updates that describe how the ProblemDetails works. Would be great if you could add another small PR for that?

Issue #738, opened 2019-09-01
[BUG] Using nullable reference types produces invalid C# (CS8632, CS8639)


martincostello:
**Describe the bug**

If the new C# 8.0 nullable feature is enabled and used for any parameters on a Refit interface, invalid C# is generated which does not compile.

**Steps To Reproduce**

  * Open a project containing Refit where the project targets at least C# 8.0.
  * Enable nullable reference types (e.g. setting `<Nullable>enable</Nullable>` in the `.csproj` file).
  * Add a nullable reference type annotation to one or more parameters on a Refit interface method.
  * Compile the application.

For example, the following interface:

```csharp
public interface ITflClient
{
    [Get("/Line/Mode/{supportedModes}")]
    Task<ApiResponse<ICollection<LineInfo>>> GetLinesAsync(
        string supportedModes,
        [AliasAs("app_id")] string? applicationId,
        [AliasAs("app_key")] string? applicationKey,
        CancellationToken cancellationToken);

    [Get("/Line/{lineId}/StopPoints")]
    Task<ApiResponse<ICollection<StopPoint>>> GetStopPointsAsync(
        string lineId,
        [AliasAs("app_id")] string? applicationId,
        [AliasAs("app_key")] string? applicationKey,
        CancellationToken cancellationToken);
}
```

Generates the following 12 errors:

```
Severity	Code	Description	Project	File	Line	Suppression State
Error	CS8632	The annotation for nullable reference types should only be used in code within a '#nullable' annotations context.	LondonTravel.Site	C:\Coding\alexa-london-travel-site\src\LondonTravel.Site\obj\Debug\netcoreapp3.0\RefitStubs.g.cs	56	Active
Error	CS8632	The annotation for nullable reference types should only be used in code within a '#nullable' annotations context.	LondonTravel.Site	C:\Coding\alexa-london-travel-site\src\LondonTravel.Site\obj\Debug\netcoreapp3.0\RefitStubs.g.cs	56	Active
Error	CS8632	The annotation for nullable reference types should only be used in code within a '#nullable' annotations context.	LondonTravel.Site	C:\Coding\alexa-london-travel-site\src\LondonTravel.Site\obj\Debug\netcoreapp3.0\RefitStubs.g.cs	64	Active
[...]

PR #800, opened 2019-11-29
Added: Support for Nullable References

joelweiss:
Support for Nullable References

#738 

**What is the new behavior?**
Figure out if the nullable is a reference type, if yes 
- Don't include the `?` in the `typeof(string?)`
- Emit `#nullable enable annotations` and `#nullable disable annotations` around the generated stubs where Nullable References are used. 

**What might this PR break?**
Nothing that I can think of.

Before it didn't compile, now it does. Do we need any other tests?

**Other information**:

There are still 2 issues with this PR, I just wanted to start a discussion to see if this is the direction you want to go
1. It doesn't detect if the return type is a Generic Nullable Reference and doesn't emit the `#nullable enable annotations`, I asked on [StackOverflow](https://stackoverflow.com/questions/59097649/how-do-i-check-with-roslyn-if-a-type-is-a-nullable-reference) why this is
2. It currently only supports  Nullable References where the Type is declared in mscorlib, if we want to support it, we will have to figure out all the reference of the dll containing the Refit interfaces.

Not supported:
```csharp
interface ICustomNullableReferenceResultAndParameterService
{
    [Get("/")]
    Task<CustomReference> Get(CustomReference? reference);
}

internal class CustomReference
{
    public int MyProperty { get; set; }
}
```


ilya-spv:
Not sure if you got that covered, but a little note: it can be an indirect reference type, e.g.:

`IDictionary<string, object?>`

This one does not break compilation, but generates a warning, so some annotations might be required I suppose.

I am also wondering if we can just put `#nullable enable annotations` on the whole generated code or it would break too much?

Dreamescaper:
What do you think about creating some helper method (e.g. `ToNullable`) for Type, and use it to handle nullability in runtime (do nothing for reference types, create nullable type for value types)?
That would allow to handle types outside of mscorlib.

joelweiss:
Rebased, and no longer emitting the `#nullable enable` (we can not always due to older compilers and detecting if nullable reference types were used didn't pan out)
I went with @iron9light's [recommendation](https://github.com/reactiveui/refit/issues/738#issuecomment-602167576) and just disabled the nullable warnings.

This PR still only recognizes a _nullable reference type_ from mscorelib, @Dreamescaper I don't understand what you are suggesting, can you please elaborate?

Dreamescaper:
@joelweiss 
So every time `typeof` for nullable type is used (both value type or reference type), I suggest using something like that:
```
		ToNullable(typeof(MyType))

		...

        private static Type ToNullable(Type type)
        {
            if (!type.IsValueType)
                return type;

            return typeof(Nullable<>).MakeGenericType(@this);
        }
```

This way you don't need any semantic info in compile time.

joelweiss:
@Dreamescaper 
Oh nice, I will try it out.

clairernovotny:
How about only adding that method if it's needed?

joelweiss:
@clairernovotny 
Ok I will do that

joelweiss:
@Dreamescaper your suggestion worked very well.
@clairernovotny I changed to emit the `ToNullable()` only when necessary.

Anything else? otherwise this PR is ready for review.

joelweiss:
Thanks @iron9light and @Dreamescaper, your suggestion were incorporated in this PR.

I think we can close Issue #738

Issue #784, opened 2019-11-14
[BUG] Multiple Input Files (May) Lead To Ambiguous Reference Errors


KyleGalvin:
**Describe the bug**
Merging 'using' statements from multiple input files into RefitStubs.g.cs can cause unexpected ambiguity. 

**Steps To Reproduce**
Ambiguous.cs:
```
namespace NS1
{
    public class Ambiguous
    {
    }
}

namespace NS2
{
    public class Ambiguous
    {
    }
}

```
RefitClientA.cs:
```
using NS2;
using Refit;

namespace RefitBugCase
{
    public interface RefitClientA
    {
        [Get("/")]
        Ambiguous Get();
    }
}
```

RefitClientB.cs:
```
using NS1;
using Refit;

namespace RefitBugCase
{
    public interface RefitClientB
    {
        [Get("/")]
        Ambiguous Get();
    }
}

```

**Expected behavior**
both refit clients compile

**Screenshots**
N/A

**Environment**
Refit <=5.0.15

**Additional context**
Two suggested proposals for a fix:
1) Breaking down RefitStubs.g.cs into a single file per input file.
_or_
2) replacing all types with fully qualified namespaces instead of 'using' headers on code generation.

KyleGalvin:
Looks like this is the underlying issue in both
https://github.com/reactiveui/refit/issues/769
and 
https://github.com/reactiveui/refit/issues/768

danlyons-softek:
> Two suggested proposals for a fix:
>
>    1. Breaking down RefitStubs.g.cs into a single file per input file.
>    or
>    1. replacing all types with fully qualified namespaces instead of 'using' headers on code generation.

I'll add another option:
3. generate `using` headers for each stub inside their respective namespaces rather than generating them all at the top

Were I writing the code myself, I'd prefer separate files per stub, but I'm more ¯\\\_(ツ)\_/¯ about generated code.

onovotny:
I'm happy to take a PR that fixes this. The main issue is that we've had is that Refit does not do semantic analysis on the phone. It just does syntax analysis to figure things out.

If there's a way to improve the template/generated code based on this, then I'm all for it. If someone wants to implement Semantic Analysis, then I'm good with that too, but that's a much bigger job (and we'd need to see how that affects build time as it's slower).

PR #803, opened 2019-12-02
Generate using statements inside namespaces

danlyons-softek:
This changes the stub generation's placement of using statements from the top of the file to inside the namespaces of each stub.  It is a quick-and-dirty means to address name collisions seen in Issue #784 


Refit.Tests/InterfaceStubGenerator.cs
```diff
@@ -1,18 +1,16 @@
...
-using System.Text;
 using System.Threading.Tasks;
+
 using Microsoft.CodeAnalysis;
 using Microsoft.CodeAnalysis.CSharp;
 using Microsoft.CodeAnalysis.CSharp.Syntax;
-using Xunit;
-using Nustache;
-using Nustache.Core;
-using Refit; // InterfaceStubGenerator looks for this
```
onovotny: Are the items in this file going to be skipped without this line? The generator looks for the `Using Refit` statement.

danlyons-softek: Mea culpa - I have my editor set up to remove/sort using statements on save and I didn't notice the comment on that line.  I'll re-add it.

onovotny:
Looks like an easy fix that'll capture many of the namespace clashes without requiring use of semantic analysis, thanks!

Please add a few tests that show this working (and to prevent regressions) and I can merge it.

danlyons-softek:
I added a simple test for the `ClassTemplateInfo` objects generated for a couple interefaces referencing `SomeType` from the `CollisionA` and `CollisionB` namespaces.  Does that seem sufficient?

I could possibly add a test added to ensure the generated C# file compiles, but I'm not super familiar with Roslyn yet, and that may be too heavy-handed.

onovotny:
That test is a part of it. I think a test that shows it working is useful too....so force the generated code to compensate for having two different `SomeType` instances in different namespaces, so we can prove the generated code works as intended.

I don't think you need to actually compile it as compiling the unit test project should trigger that code path.

onovotny:
Thanks!

Issue #823, opened 2020-01-10
Refit does not respect the BaseAddress of a provided HttpClient when creating rest services


trinitrotoluene:
**Describe the bug**
Locations provided to endpoints on a refit interface should seamlessly be appended to the HttpClient provided as long as the result is a valid URL when combined with that client's BaseAddress.

Instead refit throws an ArgumentException when calling RestService.For<T>(httpClient) if the endpoint is not prefixed with "/"

This is a problem because if your remote API has a base URL that includes a path, for example http://example.org/api/, HttpClient strips the "/api" location from its BaseAddress if it does not include a trailing "/". This is in conformance with the URI standard and is therefore rather an issue on refit's end.

**Steps To Reproduce**
- Create a refit endpoint
```
[Get("foo/bar")]
async Task FooAsync();
```
- Configure an HttpClient's BaseAddress to be `http://example.org/` (in the URI standard a trailing / is the correct way to indicate directories, HttpClient will strip out any trailing location for example).
- Attempt to call RestService.For() on your interface
- ArgumentException is thrown

**Expected behavior**
Request should be addressed to https://example.org/foo/bar

WesleyJesus99:
Hello, create a pull request to solve this problem.
could you analyze if you agree?

Any doubts, suggestions or constructive criticism will be welcome.

Thanks.

https://github.com/reactiveui/refit/pull/831

@benjaminhowarth1

clairernovotny:
Trying to isolate the issue but all paths in Refit attributes must be absolute as they'll be added to the base url. `[Get("foo/bar")]` is invalid, it should be `[Get("/foo/bar")]`

PR #831, opened 2020-01-31
add fix issue #823

WesleyJesus99:

Resolution for bug 823.
allows to create relative path without starting bars
example:
[Get ("foo")]

if you do not insert a slash at the beginning of the path relative to the application it will return an ArgumentException

**What is the new behavior?**
Allows the creation of an path without the initial bars

**What might this PR break?**
The default behavior.

**Other information**:


clairernovotny:
If we're going to do this, there needs to be tests to ensure that `var expectedBaseAddress = "http://example.com/api";` still works with people who've specified paths that start with a `/` and also tests that show what happens if you use `http://example.com/api` with paths that don't start with a `/`

clairernovotny:
I'm going to decline this change as I'm not sure what the full scenario is here. Currently Refit requires a starting slash for its paths, but that slash is relative to the base address. If something's not working, that's a bug.

Issue #832, opened 2020-02-01
feature: Enable choice between Newtonsoft.Json and System.Text.Json


Sergio0694:
**Is your feature request related to a problem? Please describe.**

Now that the new [`System.Text.Json`](https://docs.microsoft.com/en-us/dotnet/api/system.text.json?view=netcore-3.1) APIs have been around for a while, and they're also available [as a NuGet package](https://www.nuget.org/packages/System.Text.Json), I was thinking it'd be nice to start looking into a way to possibly transition away from `Newtonsoft.Json` in `refit`, remove that dependency entirely and switch to the new `System.Text.Json` namespace. Given that `refit` currently targets .NET Standard 1.4, this would be a breaking change, since the new APIs require a minimum target of .NET Standard 2.0 instead.

This would make that a candidate for a new major version of `refit`, where breaking changes would be allowed. Removing the `Newtonsoft.Json` dependency and switching to `System.Text.Json` would make `refit` much more speed and memory efficient. It would also be an occasion for simplifying the project structure, by dropping some targets that are no longer compatible with .NET Standard 2.0. Users on older platforms could still keep using the previous major version of `refit` anyway.

**Describe suggestions on how to achieve the feature**

-   Drop `Newtonsoft.Json` and switch to `System.Text.Json`
-  Target .NET Standard 2.0 as minimum framework, drop older frameworks

**Additional context**

Also, regarding the library as a whole, the codebase is generally very unorganized and with a serious lack of documentation (compared to other .NET Foundation projects like [`ImageSharp`](https://github.com/SixLabors/ImageSharp)). I'm not criticizing anyone here and I'm sure there were plenty of different reasons why the codebase ended up in this state. I'm thinking that this would be a good occasion to also do a major refactoring of the project, organizing classes with a proper hierarchy, adding missing XML docs, etc.

I'd be happy to contribute of course, but I figured it'd be a good idea to first open this issue to do some preliminary discussion on this feature as a whole, to see if it's actually feasible, what constraints there are, whether it'd be ok to drop all those older frameworks with a new major version, etc.

Paging @onovotny since I believe he's the main maintainer of the repo at the moment.

> **NOTE:** the current `refit` packages **wouldn't go anywhere** and **they would still work**. Older packages targeting older frameworks could also still keep receiving automatic updates when a dependency is updated, if that was really necessary. This is why we have semantic versioning in the first place, and one of the reasons why it exists at all.

onovotny:
There are two things here:

[...]

PR #836, opened 2020-02-08
System.Text.Json-powered IContentSerializer

Sergio0694:

This PR introduces a few changes discussed in #832.
- Remove the `net461` target
- Add a `System.Text.Json`-powered `IContentSerializer` type

`refit` currently only includes a single `IContentSerializer` type, which is powered by `Newtonsoft.Json`.

**What is the new behavior?**

This PR introduces a new performance-oriented serializer, using the `System.Text.Json` APIs.

**What might this PR break?**
The new `IContentSerializer` is not 100% at feature parity with the one using `Newtonsoft.Json` ([see here for the comparison](https://docs.microsoft.com/en-us/dotnet/standard/serialization/system-text-json-migrate-from-newtonsoft-how-to)), but since it's completely optional and up to individual users to use it, there are no breaking changes caused by that.

The other main changes is the removed `net461` target.

**Other information**:

This is the current tests situation after trying to simply drop-in replace the `Newtonsoft.Json`-powered `IContentSerializer` with the new one, without doing any other changes:

image

As mentioned before, some differences are to be expected (eg. `System.Text.Json` doesn't support JSON properties/values using the single quote `'` as a delimiter). Still though, just 16 failing tests out of 240 with a first attempt look promising 🚀

cc. @onovotny - let me know how you think we should proceed from here, whether we want to add new tests for the new serializer, and if so how to properly structure them, etc. 😊


onovotny:
Please don't remove the `net461` target. I don't think there's any reason to and that will let .NET Framework users keep using the library.

`System.Text.Json` supports `net461` as well.

onovotny:
We should have tests for the new `IContentSerializer`, but I think they should be more around ensuring that the correct implementation is used where expected, not on proving serialization/deserialization itself works.

Sergio0694:
@onovotny Ah, my bad, when you said you'd have no problem bumping the library to .NET Standard 2.0 I thought you also meant to drop `net461` by that 😅
I'll revert that commit - if `System.Text.Json` supports both there's no problem anyway!

As for tests, what do you mean by that exactly? Also, we plan to leave `Newtonsoft.Json` as default, right?
I'm currently doing some benchmarks to see if I can come up with some more performance improvements for the new serializer, then I'll be able to help with the new tests if you give me some guidance on how you'd like to structure them.

Thanks again for your time! 😊

onovotny:
I mean we should have tests that ensure that if the new `IContentSerializer` is specified in the config, that it's used later on. This might partially mean relying on subtle behavioral differences in the expected JSON so that we can run the test twice. Once with the defaults to get the result we expect with `Newtonsoft.Json` and again with the new `IContentSeralizer` to get the result we expect from `System.Text.Json`.

Refit/IO/PooledMemoryStream.cs
```diff
@@ -0,0 +1,169 @@
+﻿using System;
+using System.Buffers;
+using System.IO;
+
+namespace Refit.IO
+{
+    /// <summary>
+    /// An in-memory <see cref="Stream"/> that uses memory buffers rented from a shared pool
+    /// </summary>
+    public sealed class PooledMemoryStream : Stream
```
onovotny: Looks like `PooledMemoryStream` should probably be `internal`

onovotny:
Are there other areas in the code base that can benefit from using the `PooledMemoryStream` as well?

Refit/IO/PooledMemoryStream.cs
Sergio0694: Yup, absolutely.
I made it `public` while just writing down a draft and forgot to change that 👍

Sergio0694:
Not sure, will have to take a look. Also note that one of the reasons why the PR is still draft is that I'm not yet fully committed to that `PooledMemoryStream` type, I want to run some more benchmarks with a memory profiler to see if there are some changes I can make there. Looking at the `System.Text.Json` sources in CoreCLR it looks like the serialization overload working on a `Stream` allocates some non-pooled arrays internall, and if possible I'd like to avoid that. I'm going to write another type implementing `IMemoryBuffer<T>` instead and compare that against the current implementation.

Once I have the final set of APIs laid down and test we can figure out if there are some other places in the lib where we could reuse those, absolutely 😊

Sergio0694:
@onovotny Alright, I've finished running some benchmarks and came up with a more optimized version of the `System.Text.Json`-based `IContentSerializer`. Here's a run with a sample JSON response:

```
|          Method |     Mean |     Error |    StdDev | Ratio |   Gen 0 |   Gen 1 |   Gen 2 | Allocated |
|---------------- |---------:|----------:|----------:|------:|--------:|--------:|--------:|----------:|
|  NewtonsoftJson | 1.568 ms | 0.0010 ms | 0.0009 ms |  1.00 | 91.7969 | 29.2969 | 29.2969 | 403.22 KB |
|  SystemTextJson | 1.073 ms | 0.0017 ms | 0.0015 ms |  0.68 | 13.6719 |  1.9531 |       - |  62.09 KB |
```

We have a **~30%** speedup, and a **6.5x** memory usage reduction 🚀

Sergio0694:
Oof, I just realized I had tagged the wrong person!
I meant to tag @clairernovotny in my previous comment, my bad 😅

I just have that remaining open question from that comment, then this PR is ready for review 🚀

clairernovotny:
I'd prefer we don't remove it yet as that'd necessitate a major version bump. We can do it for next time. Please bump the minor version up to 5.1 in the version.json.

clairernovotny:
Can you please add tests for the new serializer and the integration points to ensure that the correct serliazer is used?

richbryant:
many of us working the various forms of mono-wasm implementation have discovered that System.Text.Json relies quite heavily on `Emit` functionality where Json.NET does not, meaning System.Text.Json either throws odd errors or does not function at all under mono-wasm.

This is not to say "don't do this".  Just "be aware that...".

Sergio0694:
@clairernovotny Sure thing, will do! Thanks for the feedbacks 😊

@richbryant Are you sure about that? I'm not seeing any reference to the `*.Emit` namespace in the source code ([see here](https://source.dot.net/#q=system.text.json)), plus I know that the new `System.Text.Json` APIs also work on UWP, which in Release mode uses .NET Native which is 100% AOT, and doesn't support any dynamic code generation feature as far as I know. Maybe there's something else going on there, or it's using some other execution mode that relies on code generation? 🤔

richbryant:
@Sergio0694 that could be it.  But something's up with it.  I don't think Refit is huge in the Webassembly world right now, largely because of some godawful changes the Blazor team pushed with 3.1 that make the `httpclient` settings essentially hardcoded into the framework - you can get around this but you have to know what you're doing and jump through a lot of hoops - but as a C# based system which relies on REST API calls, I expect it will definitely become so once this is fixed.

Sergio0694:
@clairernovotny Alright, I've bumped the version to `5.1` as you mentioned and added some tests. In particular, I've added the new `SystemTextJsonContentSerializer` as a type parameter in the existing serialization tests (https://github.com/reactiveui/refit/pull/836/commits/8c7252ec199c71fd9ea918042a446b3fb23b9c28) and added another test method to ensure the default `IContentSerializer` types created from the `RefitSettings` class (https://github.com/reactiveui/refit/pull/836/commits/9fb3487e15043c9c7130246efefa188fa6b2de17).

Please feel free to let me know if there are any other changes you'd like me to do, or if there is anything else I've missed. Thanks again for your time! 😊

Dreamescaper:
Please replace other mentions of `JsonContentSerializer`, e.g.:
```
        [Obsolete("Set RefitSettings.ContentSerializer = new JsonContentSerializer(JsonSerializerSettings) instead.", false)]
        public JsonSerializerSettings JsonSerializerSettings
```

Refit/Buffers/PooledBufferWriter.Stream.cs
```diff
@@ -0,0 +1,132 @@
...
+using System.IO;
+
+namespace Refit.Buffers
+{
+    internal sealed partial class PooledBufferWriter
+    {
+        /// <summary>
+        /// An in-memory <see cref="Stream"/> that uses memory buffers rented from a shared pool
+        /// </summary>
+        private sealed class PooledMemoryStream : Stream
```
benaadams: Are these `Stream`s exposed to user? If so they should override the `{Read|Write}Async` as the ones provided by the base `Stream` class are massively inefficient.

Sergio0694: Hey Ben, thank you for chiming in!
The `Stream` in question is not exposed to users, it's just used internally to create a `StreamContent` instance during the serialization pass, here:

https://github.com/Sergio0694/refit/blob/db28099069707fe7a7136b67c2eadb6a53e7d43a/Refit/SystemTextJsonContentSerializer.cs#L47

That said, you make a good point about the base `WriteAsync` and `ReadAsync` methods being inefficient, and I'm not perfectly sure those wouldn't be used by the consumer of that `StreamContent` instance. Would it make more sense to inherit from `MemoryStream` instead, to leverage the overrides of those methods from that type?

Something like this, with just the added logic to return the buffer:

```csharp
/// <summary>
/// An in-memory <see cref="Stream"/> that uses memory buffers rented from a shared pool
/// </summary>
private sealed class PooledMemoryStream : MemoryStream
{
    private bool isDisposed;

    /// <summary>
    /// Creates a new <see cref="PooledMemoryStream"/> instance.
    /// </summary>
    /// <param name="writer">The input <see cref="PooledBufferWriter"/> instance to read from.</param>
    public PooledMemoryStream(PooledBufferWriter writer)
        : base(writer.buffer, 0, writer.position, false, true)
    {
    }

    /// <inheritdoc/>
    protected override void Dispose(bool disposing)
    {
        base.Dispose(disposing);

        if (isDisposed) return;

        isDisposed = true;

        byte[] buffer = GetBuffer();

        ArrayPool<byte>.Shared.Return(buffer);
    }
}
```

What do you think?

Sergio0694:
@Dreamescaper That was my original intention as well, but I was told not to do that as that'd require a major version bump. Leaving that there allows us to just bump the version to `5.1` for now. The plan is still to eventually remove those references as well, but we'll do so in the next major update. 👍

Dreamescaper:
I don't mean to remove classes themselves.
Take a look at Obsolete message that I shared above - it suggest using JsonContentSerializer, which becomes obsolete itself.

Sergio0694:
Oh, no, I wasn't talking about removing the classes either, just that property there.
I see what you were saying now, and yeah it makes sense 👍

Done in https://github.com/reactiveui/refit/pull/836/commits/8cd298401ffeb52eca5530d791f224ca8d60d97d.

Refit/Buffers/PooledBufferWriter.Stream.cs
benaadams: No; you'd need to overload them yourself. `MemoryStream` drops back to the inefficient base methods if derived

image

Sergio0694: Oh, yeah I've seen those checks, I was just not looking at them in this context since `refit` is limited to .NET Standard 2.0, so I wouldn't be able to override those anyway 😟

I've implemented `ReadAsync` and `ReadByte` as you suggested though, in https://github.com/reactiveui/refit/pull/836/commits/1ae2b486d282f3c6cccefe86b77bd6b8afbc825a.

If you'd like me to also implement the new `Read(Span<byte>)` and `ReadAsync(Memory<byte>, CancellationToken)` overloads, I guess I could do that if we also added a `netstandard2.1` target to the library, though I suppose that'd require a major version bump as well? 🤔

clairernovotny:
Once we're good on this PR, I can get 5.1 pushed out.

Sergio0694:
@clairernovotny Awesome! 😄
I'd say it looks good on my end, I just have that open question about the overrides for the `Stream` methods working on `Span<byte>` etc., but as I mentioned I think we'd need to multitarget to `.NET Standard 2.1` in order to do that, and I'm not sure that's allowed for a minor update 🤔

If that's out of the picture for this PR, unless there are some other feedbacks or suggested changes (which would be more than welcome), the current code should be alright 👍

clairernovotny:
We can add a netstandard2.1 target, just not remove.

Sergio0694:
I've added the `.NET Standard 2.1` target and the overrides that @benaadams mentioned in https://github.com/reactiveui/refit/pull/836/commits/ab24dff7577b5f5b7c6f306e54c4cd60e7c5a8ec. I think we should now be covered for whatever method will the `System.Text.Json` APIs call on this `Stream` behind the scenes. If I missed anything, please let me know! 😊

Refit/Buffers/PooledBufferWriter.Stream.NETStandard21.cs
```diff
@@ -0,0 +1,73 @@
+﻿#if NETSTANDARD2_1
+
+using System;
+using System.IO;
+using System.Threading;
+using System.Threading.Tasks;
+
+#nullable enable
```
clairernovotny: Nit: We don't need this as the whole codebase has nullable enabled.

Sergio0694: That's weird, I'm not seeing the `<Nullable>enable</Nullable>` tag in the `.csproj` file, and in fact Visual Studio is showing warnings over the fields marked as nullable if I remove that directive 🤔

I might very well be missing something obvious here. Should I ignore the warnings and remove that directive anyway?

clairernovotny: Hmm, you're right. I thought it was already there. Can you please add `<Nullable>enable</Nullable>` to https://github.com/reactiveui/refit/blob/master/Directory.build.props instead?

Also, there's a merge conflict that needs to be resolved as well.

Dreamescaper: @clairernovotny 
Are you sure of that? Currently it leads to tons of `Dereference of a possibly null reference` across all solution.
If we think that nullability should be enabled, I'd suggest to do that in separate PR(s), will all warnings resolved.

Sergio0694: I saw this last comment only after pushing those changes in  https://github.com/reactiveui/refit/pull/836/commits/72f1d4419623ec3b9c870a514a98f565eb00e92e 😅

Of course though, if you change your mind I can revert it.

clairernovotny: You're right. Best to do that as a separate PR.

Sergio0694: Alright, reverted those changes in https://github.com/reactiveui/refit/pull/836/commits/c04b0f27d97fcc5516187d2f187ae66c20151659 👍

clairernovotny:
Okay, final thing -- can you please document this in the readme.md so people will know that it exists and how to choose which JSON serializer to use?

Sergio0694:
Done in https://github.com/reactiveui/refit/pull/836/commits/6f69cb8afc320c1b8cf73ea75e7b966d24779410, let me know if that's fine or if you'd like me to make any other changes 😊

PR #837, opened 2020-02-09
Downgrade version of Microsoft.Extensions.Http to version 2.1.1

Ninja2:

Downgrade version of package Microsoft.Extensions.Http from version 3.1.1 to version 2.1.1 giving control of referencing a greater version to the user instead of forcing an upgrade that may introduce breaking changes.

Forces version 3.1.1 of Microsoft.Extensions.Http package

**What is the new behavior?**

Allows to use versions of package Microsoft.Extensions.Http from version 2.1.1 upguard

### DIFF
--- a/Refit.HttpClientFactory/Refit.HttpClientFactory.csproj
+++ b/Refit.HttpClientFactory/Refit.HttpClientFactory.csproj
@@ -9,7 +9,7 @@
 
   <ItemGroup>
     <ProjectReference Include="..\Refit\Refit.csproj" PrivateAssets="Analyzers" />
-    <PackageReference Include="Microsoft.Extensions.Http" Version="3.1.1" />
+    <PackageReference Include="Microsoft.Extensions.Http" Version="2.1.1" />
   </ItemGroup>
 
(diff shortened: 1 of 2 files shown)


onovotny:
What is the goal of this change?

Ninja2:
The goal is to use a lower version of Microsoft.Extensions.Http to avoid forcing a posible upgrade of all the codebase when referencing the new version of Refit.HttpClientFactory.

Changing to a lower version ensures backwards compatibility with users of the package that are still using APIs for netcoreapp2 and are still in progress of being upgraded to support newer verions of the framework.

I thought I might suggest the change as it looks a simple dependecy in the project that can be downgraded with no further work required.

Chris-ZA:
Just ran into the exact problem this PR is trying to solve today.  I am stuck on older version of Refit (4.7.51) to get it to work with a .Net Core 2.2 project itself which cannot be upgraded due to other incompatible dependences.  

Any chance this can get approved so I can upgrade to a new Refit version in the future?

Thanks!

Issue #587, opened 2018-12-04
[QueryParameters] Are broken for classes with IEnumerable properties


ahmedalejo:
### THE FAILING SPECIFICATION
>### _Given_ `the following class definition`
>>```csharp
>>public class PostQueryOptions
>>{
>>      public PostQueryOptions()
>>      {
>>           KeywordIds = new int[0];
>>           MaxResults = 100;
>>      }
>>
>>      [DataMember("tags")]
>>      public int[] KeywordIds { get; set; }
>>
>>      [DataMember("limit")]
>>      public int MaxResults { get; set; }
>>
>>      public string SearchPhrase { get; set; }
>>}
>>```
>### _And_ `the following Refit interface definition`
>>```csharp
>>public interface INewsFeedClient
>>{
>>      [Get("api/posts")]
>>      Task<ObservableCollection<PostDTO>> GetNotesAsync(
>>             PostQueryOptions options, 
>>             CancellationToken token = default);
>>}
>>```
>### _When_ `the following request is made `
>>```csharp
>>GetNotesAsync(new PostQueryOptions 
>>{ 
>>     KeywordIds  = new [] { 1, 2 },
>>     MaxResults = 10,
>>     SearchPhrase = "play station"
>>})
>>```
>### Then `the following url should be generated: `
>> `GET` [api/posts?tags=1&tags=2&limit=10&SearchPrase=play+station]()
>
>### _But_ instead generates the following:
>>  `GET` [api/posts?KeywordIds=System.Int32%5B%5D&MaxResults=10&SearchPrase=play+station']() 

### Proposal
Add a method overload to  `IUrlParameterFormatter` 
That allows passing the propertyName
```csharp
public interface IUrlParameterFormatter
{
    ...
    string Format(object value, string propertyName, ParameterInfo parameterInfo);
//or
    string Format(object value, ParameterInfo parameterInfo, string propertyName = null);
}
```
Allowing the following customization:
```csharp
    public class CustomUrlParameterFormatter : DefaultUrlParameterFormatter
    {
        public override string Format(object parameterValue, string propertyName, ParameterInfo parameterInfo)
        {
            if (parameterValue is IEnumerable enumerable)
            {
                var result = ToQueryString(enumerable, parameterName: propertyName);
                return result;
            }

            return UrlHelpers.ToString(parameterValue);
        }

        public static string ToQueryString(IEnumerable query, string parameterName)
        {
            var values = query.Cast<object>().Select(ToString).ToArray();
[...]

Issue #522, opened 2018-07-10
Request object collection not serialized correctly


wozzo:
I have a search request class which has a collection which needs to be serialized as a comma delimited list, but it is currently being sent as System.String%5B%5D
```
public class Request {
    [Query(CollectionFormat.Csv)]
    public string[] Wibble { get; set; } = new string[] { "a", "b" }
}
```

the client is
```
    public interface ISearchApiClient
    {
        [Get("search")]
        Task<Response> Search(Request request);
    }
```

But the uri it ends up as is` /search?Wibble=System.String%5B%5D`. Any ideas what I'm doing wrong?

bennor:
I'm not sure the `Query` attribute is designed to work inside request objects. It definitely works for a parameter.

wozzo:
I still have the same issue without the `Query` attribute

bennor:
I think you misunderstood. What I meant was that AFAIK the query attribute only works for method parameters -- not on properties of a request class.

bennor:
If you look at `RequestBuilderImplementation.cs`, it converts request objects to querystring parameters in the `BuildQueryMap` method.

However, the only reference to `QueryAttribute` is here -- before the call to `BuildQueryMap`: https://github.com/reactiveui/refit/blob/master/Refit/RequestBuilderImplementation.cs#L504

wozzo:
The QueryAttribute is a bit of a red herring. If you have a collection on a request object it is serialized using the .ToString() implementation instead of being put onto the query string using either the CSV or Multi type options. I can reproduce with a unit test, but have yet to figure out what needs to change to fix it.

The unit test is 

```
        [Fact]
        public async Task B()
        {
            var mockHttp = new MockHttpMessageHandler();
            var settings = new RefitSettings
            {
                HttpMessageHandlerFactory = () => mockHttp
            };
            HttpRequestMessage request = null;
            mockHttp.Expect(HttpMethod.Get, "http://wibble.com/get").Respond(x =>
            {
                request = x;
                return new HttpResponseMessage(HttpStatusCode.Accepted);
            });

            var client = RestService.For<TestApi>("http://wibble.com", settings);

            var testRequest = new TestRequest
            {
                TestCollection = new List<string>()
                {
                    "A",
                    "B"
                }
            };

            // When
            await client.Get(testRequest).ConfigureAwait(false);

            // Then
            var query = request.RequestUri.Query;
            Assert.Equal("TestCollection=A,B", query);
        }
```

[...]

Issue #847, opened 2020-03-05
[BUG] QueryAttribute is not taken into account with custom objects and query parameters


ljani:
**Describe the bug**
`[QueryAttribute(Format="<format>")]` is not taken into account with passing custom objects as query parameters.

**Steps To Reproduce**
```
[Get("/samples")]
Task<List<Response>> GetSamples(Request request);

class Request {
    [Query(Format = "yyyy'-'MM'-'dd'T'HH':'mm':'ss'Z'")]
    DateTime Start { get;set; }
}
```

**Expected behavior**
`/samples?Start=2020-03-05T13:55:00Z`

**Actual behavior**
`/samples?Start=03/05/2020 13:55:00`

**Environment**
- OS: Windows 10 1909
- Device: PC
- Version: 5.0.23
- Working Version: ?

**Additional information**
Placing a breakpoint in `DefaultUrlParameterFormatter` I can see that `QueryAttribute` will be read as if it was a parameter attribute: `GetSamples([Query(Format="<format">)] Request request)`

ljani:
Thanks!

PR #848, opened 2020-03-06
handle serialization for query objects with IEnumerable properties

FolkCoder:
This PR modifies the Refit request builder to handle serialization of query objects that contain `IEnumerable` properties. This PR addresses the following issues:

https://github.com/reactiveui/refit/issues/587
https://github.com/reactiveui/refit/issues/522

Currently, an `IEnumerable` property on a query object will translate to the query string as the type name. For example, the following query object
```
class QueryObject
{
    IEnumerable<int> TestIntCollection { get; set; }
}
```
will translate as `?testIntCollection=System.Int32%5B%5D`.

**What is the new behavior?**
`IEnumerable` properties on query objects will be serialized in the same manner as `IEnumerable` parameters passed directly.

**What might this PR break?**
No foreseen impacts to existing uses.

### DIFF
--- a/Refit.Tests/RequestBuilder.cs
+++ b/Refit.Tests/RequestBuilder.cs
@@ -129,6 +129,8 @@ namespace Refit.Tests
         public string TestAlias1 {get; set;}
 
         public string TestAlias2 {get; set;}
+
+        public IEnumerable<int> TestCollection { get; set; }
     }
 
 
@@ -219,6 +221,19 @@ namespace Refit.Tests
             var uri = new Uri(new Uri("http://api"), output.RequestUri);
 
             Assert.Equal("/foo?test-query-alias=one&TestAlias2=two", uri.PathAndQuery);
+        }
+
+        [Fact]
+        public void ObjectQueryParameterWithInnerCollectionHasCorrectQuerystring()
+        {
+            var fixture = new RequestBuilderImplementation<IDummyHttpApi>();
+            var factory = fixture.BuildRequestFactoryForMethod(nameof(IDummyHttpApi.ComplexTypeQueryWithInnerCollection));
+
+            var param = new ComplexQueryObject { TestCollection = new[] { 1, 2, 3 } };
+            var output = factory(new object[] { param });
+            var uri = new Uri(new Uri("http://api"), output.RequestUri);
+
+            Assert.Equal("/foo?TestCollection=1%2C2%2C3", uri.PathAndQuery);
         }
 
         [Fact]
@@ -708,6 +723,9 @@ namespace Refit.Tests
 
         [Post("/foo")]
         Task PostWithComplexTypeQuery([Query]ComplexQueryObject queryParams);
+
+        [Get("/foo")]
+        Task ComplexTypeQueryWithInnerCollection([Query]ComplexQueryObject queryParams);
 
         [Get("/api/{obj.someProperty}")]
         Task QueryWithOptionalParametersPathBoundObject(PathBoundObject obj, [Query]string text = null, [Query]int? optionalId = null, [Query(CollectionFormat = CollectionFormat.Multi)]string[] filters = null);
--- a/Refit/RequestBuilderImplementation.cs
+++ b/Refit/RequestBuilderImplementation.cs
@@ -592,45 +592,14 @@ namespace Refit
                         var attr = queryAttribute ?? new QueryAttribute();
                         if (DoNotConvertToQueryMap(param))
                         {
-                            if (!(param is string) && param is IEnumerable paramValues)
-                            {
-                                switch (attr.CollectionFormat)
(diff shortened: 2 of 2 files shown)


ahmedalejo:
There is more
#847 

**Work Arround**:
 For your specific case
Use `IEnumerable ` instead of `IEnumerable<int>`

```csharp
class QueryObject
{
    IEnumerable TestIntCollection { get; set; }
}
```

Refit.Tests/RequestBuilder.cs
```diff
@@ -221,6 +223,19 @@ public void PostWithObjectQueryParameterHasCorrectQuerystring()
             Assert.Equal("/foo?test-query-alias=one&TestAlias2=two", uri.PathAndQuery);
         }
 
+        [Fact]
+        public void ObjectQueryParameterWithInnerCollectionHasCorrectQuerystring()
+        {
+            var fixture = new RequestBuilderImplementation<IDummyHttpApi>();
+            var factory = fixture.BuildRequestFactoryForMethod(nameof(IDummyHttpApi.ComplexTypeQueryWithInnerCollection));
+
+            var param = new ComplexQueryObject { TestCollection = new[] { 1, 2, 3 } };
```
ahmedalejo: What about `new []{1.1,2.1,3.0}`?
in some locale like `pt ` and `pt-br` `(1.2).ToString()` results in `1,2`.
We need to make sure that CultureInfo.InvariantCulture is used

FolkCoder:
@ahmedalejo You've certainly provided some food for thought. Regarding the related issue in https://github.com/reactiveui/refit/issues/847, the current behavior, which continues in my PR, is that the query attribute from the parent query object is used; the request builder does not even look for an overriding query attribute on any child properties. As you've noticed, this can certainly cause some issues as when, for example, you want a custom string format for a date property that should not be applied to an int property.

It's easy enough to add logic that will apply the most specific query attribute (e.g., use the query attribute on the parent class unless a child property defines its own). I'm going to defer to the project maintainers on whether or not this fits with their intent for query objects. If so, I'm happy to implement that behavior.

I'm not a huge fan of adding Refit-specific attributes to objects that, for my purposes, are distributed across various layers of our application (e.g., to Refit clients, to ASP.NET controllers, etc.), but it seems unavoidable if we want to encapsulate our query requests within an object. A more robust long-term solution may be configuration classes similar to how libraries such as AutoMapper or EF Core map object properties without muddying up the objects themselves. This is, obviously, a much bigger lift.

FolkCoder:
> **Work Arround**:
> For your specific case
> Use `IEnumerable ` instead of `IEnumerable<int>`

In my particular use case, the query object is passed as part of a library to .NET consumers and we want to enforce the type safety of the provided collection, so this workaround doesn't well suit our purposes. Thanks for the idea though!

Refit.Tests/RequestBuilder.cs
FolkCoder: Hmm. I didn't modify any code that actually converts properties to their string representation, but it looks like `CultureInfo.InvariantCulture` is already being used in the relevant section https://github.com/reactiveui/refit/blob/00b21bf7e25bab5f930d8b0abebdcf7d118f9c3c/Refit/RefitSettings.cs#L86 and there's a test around the culture issue as well https://github.com/reactiveui/refit/blob/00b21bf7e25bab5f930d8b0abebdcf7d118f9c3c/Refit.Tests/RequestBuilder.cs#L1564

Are you having a specific problem with this?

ahmedalejo:
> @ahmedalejo 
> I'm not a huge fan of adding Refit-specific attributes to objects that, for my purposes, are distributed across various layers of our application (e.g., to Refit clients, to ASP.NET controllers, etc.), but it seems unavoidable if we want to encapsulate our query requests within an object. A more robust long-term solution may be configuration classes similar to how libraries such as AutoMapper or EF Core map object properties without muddying up the objects themselves. This is, obviously, a much bigger lift.a

Me neither that why i´ve been favoring thing that allow to define an interface that both the client and server with implement without specific server or client decorations on the interface.

That´s why i have been considering Rpc libraries libraries like Grpc.Net and [MagicOnion](https://github.com/Cysharp/MagicOnion)

FolkCoder:
> 
> It's easy enough to add logic that will apply the most specific query attribute (e.g., use the query attribute on the parent class unless a child property defines its own). I'm going to defer to the project maintainers on whether or not this fits with their intent for query objects. If so, I'm happy to implement that behavior.
> 

Any feedback on this? I'm eager to make the changes and get this fix through (and thus avoid implementing workarounds in our current Refit clients).

ahmedalejo:
Ping @bennor

meeple142:
Thank you!!

meeple142:
Thank you!!!

PR #883, opened 2020-04-01
Usings fix

### LINKED ISSUE #876: [BUG] Stubs No Longer Compiles If Namespaces Overlap
**Describe the bug**
Using statements with names that overlap are no longer properly respected when compiling the RefitStubs.g.cs. It generates the error "The type or namespace name 'Security' does not exist in the namespace 'Project.Common' (are you missing an assembly reference?)". This was working just fine with version 5.0.23.

**Steps To Reproduce**
1. Create a class that uses namespaces similar to the following:
```cs
using Common.Security;
using Project.Common;
```
A workaround currently is to do the following:
```cs
using Project.Common;
using SomeType = global::Common.Security;
```
**NOTE** `global::` is required to make the code compile correctly.

**Expected behavior**
It should compile without error and without having to add `global::` or `using SomeType = global::Common.Security.SomeType;`. Just referencing `using Common.Security;` should compile.

**Environment**
- OS: Windows 10
- Version: 5.1.27
- Working Version: 5.0.23

Dreamescaper:
Fixes #876

All usings from original file are put inside namespace declaration in generated file. That might lead to conflicts.

**What is the new behavior?**
Force `global::` qualification for namespaces from usings, that are declared outside of namespace declaration.
Leave namespaces from usings, that are declared inside namespace declaration as-is, as they might be _not_ globally qualified (relatively to parent namespace).


clairernovotny:
Will this work if the using statement is already either `global::` or `somealias::foo`? Can we add a `global::` using statement in one of the test files to verify?

Dreamescaper:
I handled `global::` case, but haven't thought about other aliases. Will update.

Dreamescaper:
Added test Api file with global aliases.
Do we need to add explicit test for those? Or ok leaving as-is, since they fail build anyway?

I don't think we support other aliases though (not my fault).
E.g. something like that:
```
using System.Threading.Tasks;
using nsAlias = Refit.Tests.SomeNamespace;
using Refit; // InterfaceStubGenerator looks for this

namespace Refit.Tests
{
    using nsAlias::AnotherNamespace;

    public interface NamespaceWithGlobalAliasApi
    {
    }
}
```

Since we put all usings inside namespace, we cannot reference alias defined on same level.

PR #895, opened 2020-04-07
Fix generated usings in inherited interfaces

### LINKED ISSUE #892: [BUG] 5.0.23 to 5.1.54 update causes build errors (missing using directives in stub)
**Describe the bug**

Upgrading from `5.0.23` to `5.1.54` causes a build error due to some `using` directives not being added to the generated stubs. I have multiple libraries using `refit` and only one of them seems to break with this upgrade, I'll do some digging to see if I can figure out the cause or create a repro.

As a test, I've added the various `using` directives manually to the stubs, and that let me successfully build and run my app. Of course, cleaning the solution and rebuilding made those errors pop right back up again. Downgrading to `5.0.23` fixed the issue for the time being.

**EDIT:** the library fails to build with `5.1.27` as well. Will try prerelease builds next.

**EDIT 2:** some more info:

- I tried the latest `5.1.60` build, same errors
- Went through all the prerelease builds available on NuGet through the `Refit` feed. The first build causing the errors is actually just `5.0.39`. Up to `5.0.32` builds just fine.

cc. @clairernovotny is there somewhere to look up the exact list of commits between releases on the NuGet feed so that I can try seeing if I can spot the change that caused this to break? 😄
I've tried searching for them on the Azure Pipeline but I can't see packages before `5.0.59` there.

**Steps To Reproduce**
Not available yet, I'll first try all the intermediate versions from `5.0.23` to `5.1.54` to see if I can spot the first that's causing this crash to occur. Once I have that, I'll see if I can create a repro.

**Expected behavior**

Libraries using `refit` should build correctly.

**Screenshots**

image

**Environment**

- OS:       Windows 10 Pro x64 18363.720
- Device:   Desktop
- Version:   `5.1.54`
- Working Version: `5.0.23`

**Additional context**

I'm trying to build a .NET Standard 2.0 library.

Dreamescaper:
Fixes #892

If we have `InterfaceB` inheriting `InterfaceA`, during generation of `InterfaceB` we include all methods from `InterfaceA`. However - usings are not accounted, therefore if `InterfaceA` requires some using directive, that is not present in `InterfaceB` file, generated file will be broken.

**What is the new behavior?**
Take all using directives from `InterfaceA` file to generated `InterfaceB` implementation.

**What might this PR break?**
Technically it might lead to similar issue as https://github.com/reactiveui/refit/issues/784. Including usings from several files might lead to conflicts.
However, in this case it applies only to inherited types, so probability is much lower.
Not sure if we can fully handle that without full semantic analysis.


InterfaceStubGenerator.Core/InterfaceStubGenerator.cs
```diff
@@ -370,6 +372,16 @@ string GetInterfaceName(SyntaxToken identifier)
     public class UsingDeclaration
     {
         public string Item { get; set; }
+
+        public override bool Equals(object obj)
```
Dreamescaper, line 376: Added Equals and GetHashCode so that `Union` above would work properly.

Sergio0694:
This was super quick, thank you @Dreamescaper!
Also it's great to see the fix itself was indeed very small too 😄

PR #900, opened 2020-04-11
Add ability to provide custom ExceptionFactory

### LINKED ISSUE #272: Custom ApiException
Is there a way to inject a custom implementation of Refit.ApiException ?

### LINKED ISSUE #562: Add hook for exception handling and transforming?
Hi,
I am using Refit a lot, and noticed a pattern in my libraries which I dislike:

1. Use Refit
2. Notice that exceptions need to be handled somewhat differently than Refits generic mechanism (e.g. instead of throwing a generic ApiException throw an UnauthorizedAccessException if the StatusCode is a 401 or 403).
3. Write a wrapper interface around refit to statisfy #2

While this is not a large amount of code, it is annoying to write. 
Would it be possible to add some attributes to Refit which can be used to instpect a Response and transform it, if necessary?

e.g.

` [Post("/api/v1")]
 [OnException(ExceptionHandlingMethod)]

   Task<Model> CreateAsync([Header("Authorization")] string authorization, [Body] CreateModel);
`

### LINKED ISSUE #844: feature: Exceptions should be exceptional

When error codes are encountered, Refit throws. While I could catch this and handle it, the cost of the throw has already been incurred. 

Provide a means to handle the response without the exception being thrown. This could also open a door for project that would like to handle different codes in different ways.

Creating an extensible response pipeline could do it. If I could insert a handler that allows me to handle the response and short-circuit the exception, that should suffice. Following a similar pattern as ASP's exception filter pipeline would feel natural, I think.

**Additional context**

### LINKED ISSUE #882: Give a choice whether to throw an exception or not when response status code is not successful
Hi,

Would it be a big breaking change or totally nonsense thing to include a simple boolean flag in `RefitSettings` that will allow us to skip throwing exception when response status code is not successful? What I mean is that my existing api is already using internal api response wrapper and I would like to always use it from Refit and don't use built in refit `ApiResponse` which for me at least don't give any benefits because when I use it I need to double wrap that response and that doesn't make sense and complicate things a little bit.

https://github.com/reactiveui/refit/blob/77437aef0d71b9f2ff3fe790ac6ae3f6b70cdc46/Refit/RequestBuilderImplementation.cs#L250

So the question is simple, is it doable? If not why this is a bad idea because I can't see any cons of it.

Dreamescaper:
fixes #844, fixes #882, fixes #272, fixes #562

Refit throws exception if response status is not successful. No (easy) way to override that behavior.

**What is the new behavior?**
Add property to RefitSettings: 
`public Func<HttpResponseMessage, Task<Exception>> ExceptionFactory { get; set; }`

Throw exception created by this Func. If it returns null - do not throw anything.

Also, added overload to `ApiException.Create` method to allow to provide custom exception message.

**Other information**:

Open question: I wanted to allow to create custom exceptions (i.e. non `ApiException`), however that leads to issue when Api method returns `ApiResponse`. It has property `ApiException Error { get; }`, and I cannot put custom Exception type there. Currently I simply ignore exception, if it is not ApiException. 
Not sure whether there is a better way.


clairernovotny:
Sorry for taking so long to get to this. If you can please rebase this, I'll get it merged.

Dreamescaper:
@clairernovotny 
Rebased.

PR #902, opened 2020-04-12
latest coverlet

clairernovotny:


clairernovotny:
@MarcoRossignoli @tmat

Getting errors here using the 1.3.x nightly of the collectors

clairernovotny:
I wonder if there's a dependency/ordering issue when the `VSTest` target is used:

image

The `InitializeSourceControlInformation` target shows as grey in that call path.

Directory.Build.targets
```diff
@@ -7,4 +7,13 @@ https://github.com/dotnet/sourcelink/issues/572 -->
   <ItemGroup>
     <EmbeddedFiles Include="$(GeneratedAssemblyInfoFile)"/>
   </ItemGroup>
+  
+  <Target Name="CoverletGetPathMap"
```
MarcoRossignoli, line 11: Are you missing?
```xml
  <ItemGroup>
    <SourceRoot Include="$(NuGetPackageRoot)" />
  </ItemGroup>
```

clairernovotny: Shouldn't need that. That workaround is only needed for packages that contain source files.

PR #904, opened 2020-04-15
Support Xml default namespace

yhvicey:
Feature

`XmlContentSerializer` won't pass default namespace to `XmlSerializer`'s constuctor

**What is the new behavior?**
`XmlContentSerializer` will pass default namespace to `XmlSerializer`'s constuctor. If there's no default namespace specified, then default behavior (using `null` as the default namespace) will be used.

**What might this PR break?**
It won't break any existing code in theory.


clairernovotny:
Can you please add a unit test that shows it working? It will help prevent regressions.

PR #927, opened 2020-05-19
Custom error handling

yamaoto:
feature

**What is the new behavior?**
Provide the ability to independently manage errors instead of the current set of ApiException and ValidationApiException.

How doies it works?
I provided IErrorHandler that presented in RefitSettings and RestMethodInfo, by default it set to DefaultErrorHandler with current logic.

RequestBuilderImplementation for now is calling RestMethodInfo.IErrorHandler.HandleErrorAsync instead of ApiException.Create.

**What might this PR break?**
There is nothing to break. I provided `Refit.DefaultErrorHandler` with current exception management logic that used by default.

**Other information**:

### DIFF
--- a/Refit/ApiResponse.cs
+++ b/Refit/ApiResponse.cs
@@ -8,7 +8,7 @@ namespace Refit
 {
     static class ApiResponse
     {
-        internal static T Create<T, TBody>(HttpResponseMessage resp, object content, ApiException error = null)
+        internal static T Create<T, TBody>(HttpResponseMessage resp, object content, Exception error = null)
         {
             return (T)Activator.CreateInstance(typeof(ApiResponse<TBody>), resp, content, error);
         }
@@ -19,7 +19,7 @@ namespace Refit
         readonly HttpResponseMessage response;
         bool disposed;
 
-        public ApiResponse(HttpResponseMessage response, T content, ApiException error = null)
+        public ApiResponse(HttpResponseMessage response, T content, Exception error = null)
         {
             this.response = response ?? throw new ArgumentNullException(nameof(response));
             Error = error;
@@ -34,7 +34,7 @@ namespace Refit
         public HttpRequestMessage RequestMessage => response.RequestMessage;
         public HttpStatusCode StatusCode => response.StatusCode;
         public Version Version => response.Version;
-        public ApiException Error { get; private set; }
+        public Exception Error { get; private set; }
 
 
         public void Dispose()
@@ -81,6 +81,6 @@ namespace Refit
         HttpRequestMessage RequestMessage { get; }
         HttpStatusCode StatusCode { get; }
         Version Version { get; }
-        ApiException Error { get; }
+        Exception Error { get; }
     }
 }
--- a/Refit/IErrorHandler.cs
+++ b/Refit/IErrorHandler.cs
@@ -1,0 +1,14 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Net.Http;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace Refit
+{
+    public interface IErrorHandler
+    {
+        Task<Exception> HandleErrorAsync(HttpRequestMessage message, HttpMethod httpMethod, HttpResponseMessage response,
+            RefitSettings refitSettings = null);
+    }
+}
--- a/Refit/RequestBuilderImplementation.cs
+++ b/Refit/RequestBuilderImplementation.cs
@@ -244,13 +244,13 @@ namespace Refit
                 {
                     resp = await client.SendAsync(rq, HttpCompletionOption.ResponseHeadersRead, ct).ConfigureAwait(false);
                     content = resp.Content ?? new StringContent(string.Empty);
-                    ApiException e = null;
+                    Exception e = null;
                     disposeResponse = restMethod.ShouldDisposeResponse;
 
                     if (!resp.IsSuccessStatusCode && typeof(T) != typeof(HttpResponseMessage))
                     {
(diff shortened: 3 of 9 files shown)


Dreamescaper:
Seems like same as https://github.com/reactiveui/refit/pull/900

Refit/IErrorHandler.cs
```diff
@@ -0,0 +1,14 @@
...
+using System.Net.Http;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace Refit
+{
+    public interface IErrorHandler
+    {
+        Task<Exception> HandleErrorAsync(HttpRequestMessage message, HttpMethod httpMethod, HttpResponseMessage response,
+            RefitSettings refitSettings = null);
```
clairernovotny, line 12: `refitSettings` shoudln't default to `null` as it's called by library code, which we control. `refitSettings` should never be null.

Refit/RequestBuilderImplementation.cs
```diff
@@ -244,13 +244,13 @@ async Task AddMultipartItemAsync(MultipartFormDataContent multiPartContent, stri
...
                     content = resp.Content ?? new StringContent(string.Empty);
-                    ApiException e = null;
+                    Exception e = null;
                     disposeResponse = restMethod.ShouldDisposeResponse;
 
                     if (!resp.IsSuccessStatusCode && typeof(T) != typeof(HttpResponseMessage))
                     {
                         disposeResponse = false; // caller has to dispose
-                        e = await ApiException.Create(rq, restMethod.HttpMethod, resp, restMethod.RefitSettings).ConfigureAwait(false);
+                        e = await restMethod.ErrorHandler.HandleErrorAsync(rq, restMethod.HttpMethod, resp, restMethod.RefitSettings).ConfigureAwait(false);
```
clairernovotny, line 253: What happens if the error handler returns null?

Refit/RequestBuilderImplementation.cs
```diff
@@ -788,7 +788,7 @@ IEnumerable<string> ParseEnumerableQueryParameterValue(IEnumerable paramValues,
                 using var resp = await client.SendAsync(rq, ct).ConfigureAwait(false);
                 if (!resp.IsSuccessStatusCode)
                 {
-                    throw await ApiException.Create(rq, restMethod.HttpMethod, resp, settings).ConfigureAwait(false);
+                    throw await restMethod.ErrorHandler.HandleErrorAsync(rq, restMethod.HttpMethod, resp, settings).ConfigureAwait(false);
```
clairernovotny, line 791: Same, what happens if the error handler returns null?

Refit/ApiResponse.cs
```diff
@@ -81,6 +81,6 @@ public interface IApiResponse : IDisposable
         HttpRequestMessage RequestMessage { get; }
         HttpStatusCode StatusCode { get; }
         Version Version { get; }
-        ApiException Error { get; }
+        Exception Error { get; }
```
Dreamescaper, line 84: It is a breaking change.
Not sure it could be done.

clairernovotny: Good point. We could make it so that the custom exception has to derive from ApiException. That would allow people to have custom exception types. 

It's not clear to me what other exceptions people are looking to throw here?

Dreamescaper: Ability to throw `UnauthorizedAccessException` was requested here:
https://github.com/reactiveui/refit/issues/562

Refit/IErrorHandler.cs
yamaoto: This is due to the fact that in `RestMethodInfo` constructor it may be null https://github.com/reactiveui/refit/pull/927/files#diff-12d38083c42706a6ad12f6b11083cea6R43.
Should i do refitSettings not null in `IErrorHandler` and `RestMethodInfo`? This will require changing tests in `Refit.Tests\RequestBuilder.cs`

Refit/RequestBuilderImplementation.cs
yamaoto: Now it will go the way of deserializing the response and the Refit code will return `default` for this type.

It is possible to alter on when the user implementation considers that this is not an error scenario, they can return null and the refit code can deserialize the response.

For example:
Normal response
```json
{
    "result": true,
    "data" : { ... }
}
```
Error response
```json
{
    "result": false,
    "error": "error description"
}
```

and model for this
```c#
public class ApiMethodResponse
{
    public bool Result { get; set; }
    public string Error { get; set; }
    public SomeDataModel Data { get; set; }
}
```

Refit/RequestBuilderImplementation.cs
yamaoto: this will now throw a NullReferenceException
it can be redone with
```diff
- throw await restMethod.ErrorHandler.HandleErrorAsync(rq, restMethod.HttpMethod, resp, settings).ConfigureAwait(false);
+ var exception = await restMethod.ErrorHandler.HandleErrorAsync(rq, restMethod.HttpMethod, resp, settings).ConfigureAwait(false);
+ if (exception != null)
+ 
+    throw exception;
+ }
```
and go according to the scenario described above

anaisbetts:
@clairernovotny has sold me on this idea, great work! Though I agree with her that we should not change ApiException => Exception, because this means that nearly everyone will have to cast this Exception to be able to get the details of the HTTP response out.

clairernovotny:
We also now have https://github.com/reactiveui/refit/pull/900 that was merged -- that enables logic to happen to create custom exceptions. If there are additional updates to that, we'd be happy to consider them.

PR #943, opened 2020-06-19
fix: Allow for GenerateRefitStubs to be disabled with RefitDisableGenerateRefitStubs

### LINKED ISSUE #942: Refit 5.x loads Roslyn in the current MSBuild node and locks the loaded Microsoft.CodeAnalysis version
**Describe the bug**

When the following MSBuild task is loaded: 

https://github.com/reactiveui/refit/blob/477ddf0b8f5cb81e9a938d144ff8294c414382ca/Refit/targets/refit.targets#L33

The current MSBuild node loads the Microsoft.CodeAnalysis assemblies in memory, and locks the version of roslyn to the first loaded version. If it happens that there's another build task that does the same, one or the other may fail if the Roslyn versions are not the same.

This issue will get easily fixable by using the C# 9.0 source generators, but in the meantime, this task causes trouble with other source generators, particularly with [Uno.SourceGeneration](https://github.com/unoplatform/Uno.SourceGeneration).

The issue has been reported here: https://github.com/unoplatform/uno/issues/3408, and here https://stackoverflow.com/questions/61599385/refit-5-0-with-uno-platform-fails-with-missingmethodexception

**Suggested workaround**

To avoid any kind of dependency on Refit to other frameworks, adding an escape hatch to disable the Refit generation during the Uno.SourceGeneration generators' pass (with is done out of the msbuild process), the following change could be made in Refit, where:

https://github.com/reactiveui/refit/blob/477ddf0b8f5cb81e9a938d144ff8294c414382ca/Refit/targets/refit.targets#L27

would become:

```csharp
<Target Name="GenerateRefitStubs" BeforeTargets="BeforeCompile;CoreCompile" Inputs="@(Compile)" Outputs="$(RefitGeneratedFile)" Condition="$(Language) == 'C#' and '$(RefitDisableGenerateRefitStubs)'==''">
```

This would allow users to set `RefitDisableGenerateRefitStubs` to true where appropriate (in Directory.Build.Props, when `BuildingInsideUnoSourceGenerator` is set to `true`, for instance).

Would this change be appropriate in Refit ? I can make a PR if that's the case.

jeromelaban:
Fixes #942

This PR introduces the `RefitDisableGenerateRefitStubs` MSBuild property to allow for the `GenerateRefitStubs` to be skipped when other source generators based on roslyn are running.

- ~~[ ] Tests for the changes have been added (for bug fixes / features)~~

### DIFF
--- a/README.md
+++ b/README.md
@@ -919,3 +919,7 @@ catch (ApiException exception)
 }
 // ...
 ```
+
+### MSBuild configuration
+
+- `RefitDisableGenerateRefitStubs`: This property allows for other Roslyn-based source generators to disable the Refit's stubs generation during they own generation phase.
--- a/Refit/targets/refit.targets
+++ b/Refit/targets/refit.targets
@@ -24,7 +24,7 @@
     BH: Checking that the project type is a csproj, otherwise don't execute
     https://github.com/reactiveui/refit/issues/811
   -->
-  <Target Name="GenerateRefitStubs" BeforeTargets="BeforeCompile;CoreCompile" Inputs="@(Compile)" Outputs="$(RefitGeneratedFile)" Condition="$(Language) == 'C#'">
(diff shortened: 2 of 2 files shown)


jeromelaban:
Thanks @clairernovotny !

PR #945, opened 2020-07-01
refactor: change verify url method exception message

InRedikaWB:
Add more information about exception in verify url method

If url don't start with '/', refit throw exception that "URL path {relativePath} must be of the form '/foo/bar/baz'". 

**What is the new behavior?**
The only one thing that this method check - if path start with '/'. If it's start with '/' than we will have at least two parts, so second check make no sense.


bennor:
Second check was always wrong - `string.Split()` always returns at least one element.

Issue #636, opened 2019-03-15
Content in ValidationApiException hides Content in derived ApiException


SWarnberg:
Hi

I think it's a bad design to hide the **Content** property of the derived class **ApiException**. The reason is that if you don't catch **ValidationApiException** but catch **ApiException**, you expect to get the **Content** property as string. However, if the real content is a problem+json document, it's null in the ApiException.
A better approach would be to have a new property in the ValidationApiException class that returns the ProblemDetails and keep Content as text, giving the client a chance to deserialize the problem itself.

Staffan

DrakeLambert:
I'm having this exact issue right now. Took me 30 minutes to finally track down this issue.

benjaminhowarth1:
It's because the request object is disposed when catching the exception. I'll look at how to do this in a better fashion.

DrakeLambert:
@jamiehowarth0 See my PR!

winterdouglas:
I'm facing this exact same issue.

PR #953, opened 2020-07-31
Support ApiException.Content when actual type is ValidationApiException

DrakeLambert:
Fixes reactiveui/refit#636

Consider this usage of a Refit client:
```csharp
try
{
    await refitClient.GetAsync(123);
}
catch (ApiException e)
{
    Console.WriteLine(e.Content);
}
```

If the response follows [RFC 7808](https://tools.ietf.org/html/rfc7807), then `e.Content` will unexpectedly be `null`.

This PR would change the behavior so that `e.Content` would still be set to the string content of the response message. If the exception is caught as `ValidationApiException`, the `public ProblemDetails Content { get; }` will still function as normal.

Enhancement: Support use of `ApiException.Content` when hidden by `ValidationApiException.Content`.

`ApiException.Content` is left as default, `null`, when hidden by `ValidationApiException.Content`.

**What is the new behavior?**

`ApiException.Content` is set to the string content of the response message.

**What might this PR break?**

If consumers rely on `ApiException.Content` to be `null` when the actual type is `ValidationApiException`, then their code may break. As far as I can tell, the alternative is more of an issue than this fix.

What docs should be added for this? I neglected this step because I considered the original behavior to be unexpected.


jamiehowarth0:
Brilliant, thanks for this! I'll review shortly

DrakeLambert:
@jamiehowarth0 who else could give a review here? I'd like to see what y'all think of this change.

jamiehowarth0:
> @jamiehowarth0 who else could give a review here? I'd like to see what y'all think of this change.

Sorry for the delay, I'll check this in the next 24hrs

DrakeLambert:
@jamiehowarth0 ping!

Refit.Tests/ResponseTests.cs
```diff
@@ -221,6 +221,31 @@ public async Task BadRequestWithEmptyContent_ShouldReturnApiException()
...
+            var expectedResponse = new HttpResponseMessage(HttpStatusCode.BadRequest)
+            {
+                Content = new StringContent(expectedContent)
+            };
+            expectedResponse.Content.Headers.ContentType = new MediaTypeHeaderValue("application/problem+json");
+            mockHandler.Expect(HttpMethod.Get, "http://api/aliasTest")
+                .Respond(req => expectedResponse);
+
+            var actualException = await Assert.ThrowsAsync<ValidationApiException>(() => fixture.GetTestObject());
+            var actualBaseExcpetion = actualException as ApiException;
```
jamiehowarth0, line 246: Please can you rename these (typo in "exception")? I'll accept the PR

jamiehowarth0 (changes requested):

Refit.Tests/ResponseTests.cs
DrakeLambert: Done!

PR #964, opened 2020-08-17
Disposable interfaces

### LINKED ISSUE #660: IDisposable Interfaces
using (var client = RestService.For<IInterface>(Url))
{
// code here
}

this requires the Interface to be IDisposable which makes sense , but when you make It IDisposable, it asks for a Dispose() Method..  but it's an interface .... the code generation needs to generate the appropriate Dispose call when an interface inherits from IDisposable.

trejjam:
Feature, resolve #660. A generated client has an implementation of `IDisposable.Dispose` (if interface is marked as `IDisposable`)

`IDispose` is not generated, when necessary.

**What is the new behavior?**
Interface marked with `IDisposable` has implemented `IDispose` method
```cs
void IDisposable.Dispose()
{
    Client?.Dispose();
}
```

**What might this PR break?**
Nothing

**Other information**:
This feature is handling following messages when `AddRefitClient` is used for registration (`HttpClient` is created using `IHttpClientFactory`)
```
[15:18:02 DBG] Microsoft.Extensions.Http.DefaultHttpClientFactory
        HttpMessageHandler expired after 120000ms for client '...'
[15:18:12 DBG] Microsoft.Extensions.Http.DefaultHttpClientFactory
        Starting HttpMessageHandler cleanup cycle with 2 items
[15:18:12 DBG] Microsoft.Extensions.Http.DefaultHttpClientFactory
        Ending HttpMessageHandler cleanup cycle after 0.6748ms - processed: 0 items - remaining: 2 items
```


bennor:
Weirdly, `HttpClient` is not supposed to be disposed. From the [official docs](https://docs.microsoft.com/en-us/dotnet/api/system.net.http.httpclient?redirectedfrom=MSDN&view=netcore-3.1#remarks):

>  HttpClient is intended to be instantiated once and re-used throughout the life of an application. Instantiating an HttpClient class for every request will exhaust the number of sockets available under heavy loads. This will result in SocketException errors.

trejjam:
Hi, yes the default (empty) constructor should not. HttpClient created through HttpClientFactory has shorter lifetime with different internal behaviour.

iansteed:
@trejjam is correct, a generated client is injected with a new HttpClient instance each time it's created. The instance contains client specific base address, timeout, headers, and pipeline handlers. The underlying HttpMessageHandler is performing the actual requests and is managed by the HttpClientFactory.

https://docs.microsoft.com/en-us/dotnet/architecture/microservices/implement-resilient-applications/use-httpclientfactory-to-implement-resilient-http-requests

bennor:
Looks like I have some reading to do. 😳

Issue #971, opened 2020-09-14
[BUG] Dictionary contained inside ComplexQueryParam not properly transformed in QueryString


ThomasSmeets:
**Describe the bug**
Dictionaries contained inside object for GET methods aren't managed properly. While they are managed properly when directly provided as parameters

```
 public class MyComplexQueryParams
 {
       //Other properties omitted for example purpose

        public Dictionary<string, object> MetaData { get; set; } = new Dictionary<string, object>();

       //Other properties omitted for example purpose
 }
```

Instead of transforming properly the dictionary as Query parameter with the following format
`&{prefix}{delimiter}{keyvalue.Key}={keyvalue.Value}`

it consider it as an IEnumerable and  use the following format 
` &{prefix} = [keyvalue.Key, keyvalue.Value]`

**Expected behavior**

Use the format : ` &{prefix}{delimiter}{keyvalue.Key}={keyvalue.Value}
`

**Additional Information**

I've checked, the code to see what's happening.
The problem is located in the **RequestBuilderImplementation.cs**

Some treatments are done to handle the **IEnumerable**  before the specific test to the **IDictionary** => But the **IDictionary** Implement **IEnumerable** so it means the **IDictionary** is not handled anymore , but instead is always handled as **IEnumerable**.

```
                if (!(obj is string) && obj is IEnumerable ienu)
                {
                    foreach (var value in ParseEnumerableQueryParameterValue(ienu, propertyInfo, propertyInfo.PropertyType, queryAttribute))
                    {
                        kvps.Add(new KeyValuePair<string, object>(key, value));
                    }

                    continue;
                }

 `              if (DoNotConvertToQueryMap(obj))
                {
                    kvps.Add(new KeyValuePair<string, object>(key, obj));
                    continue;
                }

                switch (obj)
                {
                    case IDictionary idict:
                        foreach (var keyValuePair in BuildQueryMap(idict, delimiter))
                        {
                            kvps.Add(new KeyValuePair<string, object>($"{key}{delimiter}{keyValuePair.Key}", keyValuePair.Value));
                        }

                        break;
```
**Suggested Fix**
A possible fix would be to simple change the condition
` if (!(obj is string) && obj is IEnumerable ienu)`
by
` if (!(obj is string) && obj is IEnumerable ienu &&  !(obj is IDictionary) )`

If you are fine with that solution, I can make a PR

**Problem with Unit tests**

[...]

PR #976, opened 2020-09-29
Fix dictionary query request mapping

FolkCoder:

Bug fix

In the current version (5.2.1), dictionary query parameters with non-string keys throw an exception: `'Unable to cast object of type 'Example.Enum' to type 'System.String'.'`.

Additionally, if the dictionary parameter is part of a complex object, the dictionary is serialized as `?DictionaryPropertyName=%5BKey%2C%20Value%5D`.

This latter issue is reported [here](https://github.com/reactiveui/refit/issues/971).

**What is the new behavior?**

The current query mapping code for IDictionary properties assumes that the dictionary key type is a string. This PR explicitly calls `ToString()` on the key before storing its value to the query map.

For dictionaries that are a property on a complex query object, the current behavior treats them the same as ordinary collections (e.g., implements `IEnumerable<T>`). This PR implements the quick fix offered by @ThomasSmeets to skip this handling and be handled by the appropriate query map method.

**What might this PR break?**

The change to handle IDictionary query parameters fixes an unhanded exception, this change is unlikely to break any existing assumptions.

The handling for IDictionary query properties that are part of a complex query objects may cause breaking changes for consumers who are expecting the previous serialization format (documented in [this issue](https://github.com/reactiveui/refit/issues/971)). However, prior to version 5.1.67, dictionary properties on query objects serialized as expected so this is likely a recent regression introduced by [issue #896](https://github.com/reactiveui/refit/pull/896) and consumers are unlikely to be relying on this strange serialization behavior.

### DIFF
--- a/Refit/RequestBuilderImplementation.cs
+++ b/Refit/RequestBuilderImplementation.cs
@@ -416,9 +416,7 @@ namespace Refit
 
                 if (DoNotConvertToQueryMap(obj))
                 {
-                    var objType = obj.GetType();
-                    var formattedValue = settings.UrlParameterFormatter.Format(obj, objType, objType);
-                    kvps.Add(new KeyValuePair<string, object>(formattedKey, formattedValue));
+                    kvps.Add(new KeyValuePair<string, object>(formattedKey, obj));
                 }
                 else
                 {
--- a/Refit/RequestBuilderImplementation.cs
+++ b/Refit/RequestBuilderImplementation.cs
@@ -2,6 +2,7 @@
 using System.Collections;
 using System.Collections.Concurrent;
 using System.Collections.Generic;
+using System.ComponentModel.Design;
 using System.IO;
 using System.Linq;
 using System.Net.Http;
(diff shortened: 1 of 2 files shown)


Refit/RequestBuilderImplementation.cs
```diff
@@ -404,16 +404,15 @@ private async Task<T> DeserializeContentAsync<T>(HttpResponseMessage resp, HttpC
...
+            foreach (var key in dictionary.Keys)
             {
                 var obj = dictionary[key];
                 if (obj == null)
                     continue;
 
                 if (DoNotConvertToQueryMap(obj))
                 {
-                    kvps.Add(new KeyValuePair<string, object>(key, obj));
+                    kvps.Add(new KeyValuePair<string, object>(key.ToString(), obj));
```
clairernovotny, line 415: In general this looks good -- but we should not call ToString() as that makes it harder to override the value for types you don't control. Instead it should probably delegate to the `settings.UrlParameterFormatter` to do the format.

https://github.com/reactiveui/refit/blob/master/Refit/RefitSettings.cs#L24

FolkCoder: Thanks for the helpful feedback! I'll dig into this and hopefully have changes to this PR out sometime tomorrow.

FolkCoder: @clairernovotny I've made updates to the PR to ensure that the query map respects provided formatters. My unit tests match my assumptions on what the generated query string is supposed to look like with the formatters, but please let me know if I'm missing something. Thanks again!

Issue #809, opened 2019-12-07
feature: Add the possibility to set Properties in HttpRequestMessage


unleashed-jamest:
**Is your feature request related to a problem? Please describe.**

Currently there isn't a (nice) way to pass data to your pipeline of DelegatingHandler classes. 

**Describe the solution you'd like**

What I would like to see is the ability to place data into HttpRequestMessage.Properties dictionary on a per-request basis. It should function basically the exact same way the Headers and Header attributes do, except you pass in key value pairs and they get stored in the HttpRequestMessage.Properties. 

**Describe alternatives you've considered**

I currently am using the Headers and Header attributes to accomplish this goal, but it's a nasty hack. It works, but it's not a nice solution. It's not what HTTP headers are intended for. I currently have 3 very different logging concerns that I take care of using a series of DelegatingHandler classes and each pulls out the header(s) its looking for and logs the contents if certain criteria are met. For example I need to log a friendly error message in a user facing log so there is one handler for that and the exception message (if the result is non 200) is stored in a header that I pull out and throw an exception which bubbles up to the client. I also need to log a success message in a user facing log if the request succeeded, so there is a header for that. Plus some diagnostics information for instrumenting performance and tracing metrics on the whole pipeline etc.

**Describe suggestions on how to achieve the feature**

It looks like it's not too hard to do, and if you agree with the approach I'd be happy to submit a PR. 
How I would implement it is add two new attributes `PropertiesAttribute` and `PropertyAttribute` and in the constructor of `RestMethodInfo.cs` extract the values present on the MethodInfo and MethodParams pretty much exactly the same as Headers and Header attributes do it and store them on the `RestMethodInfo` class.  

Then inside `RequestBuilderImplementation` take the Properties we saved in `RestMethodInfo` and populate them into the `HttpRequestMessage.Properties`

The one problem I can see is that all the building of those request objects is currently cached.  So, you'd need to be able to opt-out of caching either manually (with an attribute) or automatically (if Property attribute is detected) or update the cache key to include Property parameters key AND value pairs.

clydevassallo:
Was this ever considered?

[...]

PR #1001, opened 2020-11-22
809 set properties in http request message

nicostouch:
This PR introduces a new feature to Refit. Being able to pass runtime state into the `HttpRequestMessage.Properties` dictionary in order to access it inside a `DelegatingHandler` by adding a `Property` attribute to a parameter.

    public interface IGitHubApi
    {
        [Post("/users/new")]
        Task CreateUser([Body] User user, [Property("SomeKey")] string someValue);

        [Post("/users/new")]
        Task CreateUser([Body] User user, [Property] string someOtherKey);
    }

    class RequestPropertyHandler : DelegatingHandler
    {
        public RequestPropertyHandler(HttpMessageHandler innerHandler = null) : base(innerHandler ?? new HttpClientHandler()) {}

        protected override async Task<HttpResponseMessage> SendAsync(HttpRequestMessage request, CancellationToken cancellationToken)
        {
            // See if the request has a the property
            if(request.Properties.ContainsKey("SomeKey")
            {
                var someProperty = request.Properties["SomeKey"];
                //do stuff
            }

            if(request.Properties.ContainsKey("someOtherKey")
            {
                var someOtherProperty = request.Properties["someOtherKey"];
                //do stuff
            }

            return await base.SendAsync(request, cancellationToken).ConfigureAwait(false);
        }
    }

PR also includes:
- full documentation on 
  - how to use the feature
  - behavior of edge cases / internals 
- comprehensive set of unit tests covering:
  - adding a property with a key
  - adding a property without specifying its key
  - adding properties with duplicate keys
  - how the feature interacts with Query, Header, and Multipart parameters

I opened issue https://github.com/reactiveui/refit/issues/809 for it about a year ago.

Currently Refit doesn't support doing this natively, and it can only be worked around in a pretty hacky way with strings only.
That issue has seen some interest since opening it. The feature is available in RestEase and on the roadmap for Flurl.

**What might this PR break?**
The excellent code-coverage of the unit tests has ensured it hasn't broken anything fundamentally. 
Additionally I've taken great care to add enough coverage to tease out any subtle bugs the feature might have introduced into Query, Header or Multipart functionality. Happy to add any additional test cases if deemed necessary.

**Other information**:
I love Refit and this closes a functionality gap making Refit the only one out of Refit, RestSharp, RestEase, and Flurl to support `HttpClient`, `IHttpClientFactory`, and populating properties into `HttpRequestMessage.Properties`


dnfadmin:
[CLA assistant check](https://cla.dotnetfoundation.org/reactiveui/refit?pullRequest=1001) <br/>Thank you for your submission, we really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla.dotnetfoundation.org/reactiveui/refit?pullRequest=1001) before we can accept your contribution.<br/><br/>:x: nicostouch [sign now](https://cla.dotnetfoundation.org/reactiveui/refit?pullRequest=1001)<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](https://cla.dotnetfoundation.org/check/reactiveui/refit?pullRequest=1001) it.</sub>

Refit/Attributes.cs
```diff
@@ -206,6 +206,19 @@ public HeaderAttribute(string header)
         public string Header { get; }
     }
 
+    [AttributeUsage(AttributeTargets.Parameter)]
```
clairernovotny, line 209: This type should have doc comments on it and its members

Refit/RequestBuilderImplementation.cs
```diff
@@ -666,6 +675,11 @@ private async Task<T> DeserializeContentAsync<T>(HttpResponseMessage resp, HttpC
                     }
                 }
 
+                foreach (var property in propertiesToAdd)
+                {
+                    ret.Properties.Add(property.Key, property.Value);
```
clairernovotny, line 680: I know there's a test for the Dup, but Add won't throw here if the key is already present?

clairernovotny (changes requested):
This PR looks great, just one small thing and I can merge it.

Refit/Attributes.cs
nicostouch: I have added some docs to it :)

Refit/RequestBuilderImplementation.cs
nicostouch: I've updated this to use the indexer rather than Add method. In the first instance though propertiesToAdd never contains duplicate keys. Also since Refit itself is constructing the HttpRequestMessage before any DelegatingHandlers get to run it's going to be empty, so I think it's likely OK, but have changed it to more of a 'last write wins' in keeping with how the rest of the feature works and erring on the side of not blowing up if it ever happens.

PR #973, opened 2020-09-24
add a new property Name to MultipartItem class

javirszhang:
consider following suituation, my webapi get upload files from Request.Form.Files so that IFormFile.Name is very important.
```
[HttpPost("upload/{bucket}")]
public ApiResult<List<FileUploadResponseDTO>> FileUpload(string bucket)
{
    List<FileUploadResponseDTO> list = new List<FileUploadResponseDTO>();
    MediaFileUploadProvider mediaFile = new MediaFileUploadProvider(bucket);
    foreach (var file in Request.Form.Files)
    {
        var fileDTO = mediaFile.Execute(file.Name, file.FileName, file.OpenReadStream());
        if (fileDTO != null)
        {
            list.Add(fileDTO);
        }
    }
    return list;
}
```
IFormFile.Name determins file means, it can not use a fixed parameter name.

**What kind of change does this PR introduce?**
enhancement, let users can set Content-Disposition name  attribute when necessary.

**What is the current behavior?**
Always use the parameter name as the Content-Disposition name attribute.

**What is the new behavior?**
Add a new property Name to MultipartItem  class, so we can set it when necessary.

**What might this PR break?**
nothing

**Please check if the PR fulfills these requirements**
- [x] Tests for the changes have been added (for bug fixes / features)
- [ ] Docs have been added / updated (for bug fixes / features)

PR #1002, opened 2020-11-23
Feature- multipart data name to be specified at the moment of execution.

farcasclaudiu:
Feature - multipart data name to be specified at the moment of execution.

Feature

On current version parameter name or [AliasAs] attribute is used to specify multipart data name. This behavior is not allowing to have multipart data name that is known only at the execution time.

**What is the new behavior?**
The multipart data name will be taken from streamPart.FileName and can be specified at runtime (dynamically) when the multipart data name is not know ahead of time. If streamPart.FileName is empty ("") it would behave like in previous version (parameter name or [AliasAs]).

**What might this PR break?**
Because streamPart.FileName will take priority it could break any usage where both streamPart.FileName and [AliasAs] (or parameter name) are used.

**Other information**:
I needed this change into a project where multipart was containing a presentation part and multiple binary parts. 
Names of the binary parts needed to be dynamically constructed at runtime and the presentation part has dynamic references to the binary parts.


dnfadmin:
[CLA assistant check](https://cla.dotnetfoundation.org/reactiveui/refit?pullRequest=1002) <br/>Thank you for your submission, we really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla.dotnetfoundation.org/reactiveui/refit?pullRequest=1002) before we can accept your contribution.<br/><br/>:x: farcasclaudiu [sign now](https://cla.dotnetfoundation.org/reactiveui/refit?pullRequest=1002)<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](https://cla.dotnetfoundation.org/check/reactiveui/refit?pullRequest=1002) it.</sub>

clairernovotny:
Overall, it makes sense to be able to specify the data name dynamically. I'm not sure we want to conflate the filename with the data name. I think it would be better to add a new property to `StreamPart`, along with a new constructor overload, where the user can specify it if desired. The default would be empty, so the existing tests and behavior wouldn't change.

clairernovotny (changes requested):
See my previous comment

clairernovotny:
Can we combine this with the PR (https://github.com/reactiveui/refit/pull/973) from @javirszhang where the change is on `MultiartItem` so that it works for more than Streams?

farcasclaudiu:
My intention was to have a minimal impact possible on the code, but I can have a look on PR (#973)  and eventually adapt it (to cover the need for dynamic datapart name). 
Either way, there needs to be created a couple of unit tests to cover this new property usage scenario and the right priority between these various ways of datapart naming. I guess that this new property will have top priority, second will be [AliasAs] and lastly the parameter name.
Plus updating the main documentation (README.md file) to acknowledge this new behavior and priorities.

PR #1005, opened 2020-11-25
Use System.Text.Json by default

### LINKED ISSUE #981: Split out serialization format from main library
Now that we have abstracted the serialization formatter, we should split out the implementation from the main library. That is, we should have a Newtonsoft.Json implementation and a System.Text.Json implementation that can be opted-in to.

This will need some design. Things to consider:
1. There should be a default. Perhaps `System.Text.Json` since it is a framework reference in .NET Core and doesn't need an explicit dependency there.
2. This will be a breaking change no matter what we do. Flipping the default, moving code to another package, etc.
3. Naming of packages -- what do people install? `Refit.Newtonsoft.Json`, `Refit.Xml`, `Refit.System.Text.Json?` What is in `Refit`, do people install `Refit` directly anymore?

See also: https://github.com/reactiveui/refit/issues/980

clairernovotny:
Implements and fixes #981

Refit now uses System.Text.Json by default and provides a new package containing support for Newtonsoft.Json.

Todo: Update docs.

@anaisbetts @bennor

### DIFF
--- a/Refit.Tests/RequestBuilder.cs
+++ b/Refit.Tests/RequestBuilder.cs
@@ -1510,8 +1510,16 @@ namespace Refit.Tests
             var factory = fixture.BuildRequestFactoryForMethod(nameof(IDummyHttpApi.FetchSomeStuffWithDynamicRequestProperty));
             var output = factory(new object[] { 6, someProperty });
 
+#if NET5_0
+            Assert.NotEmpty(output.Options);
+            Assert.Equal(someProperty, ((IDictionary<string, object>)output.Options)["SomeProperty"]);
+#endif
+
+#pragma warning disable CS0618 // Type or member is obsolete
             Assert.NotEmpty(output.Properties);
             Assert.Equal(someProperty, output.Properties["SomeProperty"]);
+#pragma warning restore CS0618 // Type or member is obsolete
+
         }
 
         [Fact]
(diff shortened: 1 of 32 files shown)


Refit.Tests/RequestBuilder.cs
```diff
@@ -1058,7 +1062,7 @@ public void MethodsShouldBeCancellableDefault()
         {
             var fixture = new RequestBuilderImplementation<ICancellableMethods>();
             var factory = fixture.RunRequest("GetWithCancellation");
-            var output = factory(new object[0]);
+            var output = factory(Array.Empty<object>());
```
bennor, line 1065: Did not know this exists, but it looks like it's not even new. I assume it's cheaper than creating an empty array every time?

Refit/CachedRequestBuilderImplementation.cs
```diff
@@ -20,7 +20,7 @@ public CachedRequestBuilderImplementation(IRequestBuilder innerBuilder)
         }
 
         readonly IRequestBuilder innerBuilder;
-        readonly ConcurrentDictionary<string, Func<HttpClient, object[], object>> methodDictionary = new ConcurrentDictionary<string, Func<HttpClient, object[], object>>();
+        readonly ConcurrentDictionary<string, Func<HttpClient, object[], object>> methodDictionary = new();
```
bennor, line 23: 😻

Refit.Tests/RequestBuilder.cs
clairernovotny: Yup, removes allocations

Issue #1018, opened 2021-01-06
feature: Add support for interface default methods


MariusVolkhart:
**Is your feature request related to a problem? Please describe.**
I have a service that exposes an API that isn't very ergonomic. We see callers repeatedly writing the same 5 lines to interact with the service. We'd like to make it easier by adding a default method to our Refit interface that does the logic of those 5 lines, which includes calling a Refit-annotated method.

```c#
interface Mine {
  [Get("/events.json")]
  Task<Events> GetEvents(string query);
  
  Task<Events> GetEventsOrDefault(string query) {
    try {
      GetEvents(query);
    } catch (ApiException e) {
        // Look at exception for Reasons
    }
}
```

**Describe the solution you'd like**
Right now, Refit errors with the message, "System.NotImplementedException : Either this method has no Refit HTTP method attribute or you've used something other than a string literal for the 'path' argument."

#845 Had a fix for static default methods, but it does not appear to address default instance methods.

**Describe alternatives you've considered**
For now we're using an extension method

MariusVolkhart:
🥳  thanks for doing this @clairernovotny!

PR #1023, opened 2021-01-13
Add support for internal and non-abstract interface members

tkp1n:
Adds support for:
* `internal` interface members
* non-abstract interface members (as requested in #1018)

This allows wrappers around "Refit methods" to be written directly in the interface with the direct REST calls potentially hidden as `internal` methods.

```csharp
public interface IApiClient {

    // This method will be generated by Refit but hidden from consumers outside of the current package
    [Get("")]
    internal Task<string> GetInternal();

    // Wraps the direct REST call above with some customization logic and is exposed to external consumers
    public async Task<string> Get()
    {
        return $"Result: {await GetInternal()}";
    }
}
```

Previously, this was only possible by defining the IApiClient as internal, defining another public interface and an implementation class using the IApiClient.

* Internal abstract members on Refit interfaces are not fully supported.
* Non-abstract members on Refit interfaces are not supported.

**What is the new behavior?**
See above...

**What might this PR break?**
Nothing I could think of. Feel free to prove me wrong, however.

--> Please excuse opening this PR before any discussion was held on the issue mentioned above. Feel free to close this, if it doesn't fit Refits roadmap.


clairernovotny:
Thanks for this! I had to recreate it on top of the semantic generation that I just checked in. The new PR is here: https://github.com/reactiveui/refit/pull/1031

PR #1045, opened 2021-01-28
Enable first unit tests for InterfaceStubGenerator

sharwell:


Refit.Tests/Verifiers/CSharpSourceGeneratorVerifier`1.cs
```diff
@@ -0,0 +1,9 @@
+﻿using Microsoft.CodeAnalysis;
+
+namespace Refit.Tests
+{
+    public static partial class CSharpSourceGeneratorVerifier<TSourceGenerator>
```
sharwell, line 5: 📝 This file would normally contain helpers for the most common cases, but we haven't identified those yet for source generators.

Refit.Tests/InterfaceStubGenerator.cs
```diff
@@ -3,22 +3,29 @@
...
 using Task = System.Threading.Tasks.Task;
+using VerifyCS = Refit.Tests.CSharpSourceGeneratorVerifier<Refit.Generator.InterfaceStubGenerator>;
 
 namespace Refit.Tests
 {
     public class InterfaceStubGeneratorTests
     {
+        static readonly MetadataReference RefitAssembly = MetadataReference.CreateFromFile(
+            typeof(GetAttribute).Assembly.Location,
+            documentation: XmlDocumentationProvider.CreateFromFile(Path.ChangeExtension(typeof(GetAttribute).Assembly.Location, ".xml")));
```
sharwell, line 27: 📝 This would be better if we could reference a NuGet package (where it would automatically detect the proper TFM to reference and include the corresponding dependencies), but that would restrict the tests to using a Refit build available on NuGet.

clairernovotny:
Looks like there are test failures here?

sharwell:
Do you have a preference regarding https://github.com/reactiveui/refit/pull/1045#discussion_r565753379? The options are:

1. Run the SG test against a build of Refit from NuGet, and have it automatically configure all the necessary references
1. Run the SG test against the actual build of Refit from the current repository, but need to set up the references as part of each test configuration

clairernovotny:
Seems like 2 is better since we the CI feed would always be N-1 of the current, so there'd be no way to validate things before a commit?

sharwell:
The tests are now updated.

clairernovotny:
I"m seeing two test failures?

https://dev.azure.com/dotnet/ReactiveUI/_build/results?buildId=44823&view=ms.vss-test-web.build-test-results-tab

sharwell:
> I'm seeing two test failures?

I'm not seeing this locally, and the CI output isn't public so I'm not sure yet where the difference lies.

> Are we referencing NuGet.Protocol anywhere here in the closure where this matters?

Yes, the test library uses NuGet to resolve package dependencies.

clairernovotny:
The build/test output is public but ADO is being silly when you've already logged into a different ADO tenant. If you open the ADO link with an InPrivate window, you'll be able to see it.

Refit.Tests/InterfaceStubGenerator.cs
```diff
@@ -1,24 +1,50 @@
...
+
+        static readonly ReferenceAssemblies ReferenceAssemblies;
+
+        static InterfaceStubGeneratorTests()
+        {
+#if NET5_0_OR_GREATER
```
sharwell, line 34: 🤦‍♂️ Looks like this was added for the 5.0.2xx SDK, but the CI is using 5.0.1xx.

clairernovotny:
Good catch, we both have the 200 preview SDK's locally 😂

PR #1069, opened 2021-02-09
Add logic to emit #nullable in the generated files if required

clairernovotny:
@sharwell I'd appreciate a code review

### DIFF
--- a/InterfaceStubGenerator.Core/InterfaceStubGenerator.cs
+++ b/InterfaceStubGenerator.Core/InterfaceStubGenerator.cs
@@ -163,8 +163,7 @@ namespace Refit.Implementation
                         // Add the interface to the generation list with an empty set of methods
                         // The logic already looks for base refit methods
                         interfaces.Add(ifaceSymbol, new List<IMethodSymbol>() );
-                        var isAnnotated = context.Compilation.Options.NullableContextOptions == NullableContextOptions.Enable ||
-                                            model.GetNullableContext(iface.SpanStart) == NullableContext.Enabled;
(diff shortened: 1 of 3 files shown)


InterfaceStubGenerator.Core/InterfaceStubGenerator.cs
```diff
@@ -114,47 +114,65 @@ internal static partial class Generated
...
+                                                     .Where(m => IsRefitMethod(m, httpMethodBaseAttributeSymbol))
+                                                     .Any();
+
+                    if (hasDerivedRefit)
+                    {
+                        // Add the interface to the generation list with an empty set of methods
+                        // The logic already looks for base refit methods
+                        interfaces.Add(ifaceSymbol, new List<IMethodSymbol>() );
+                        var isAnnotated = context.Compilation.Options.NullableContextOptions == NullableContextOptions.Enable ||
+                                            model.GetNullableContext(iface.SpanStart) == NullableContext.Enabled;
```
sharwell, line 167: Is there a reason we aren't just checking the current context?

```suggestion
                        var isAnnotated = model.GetNullableContext(iface.SpanStart) == NullableContext.Enabled;
```

sharwell (commented):
Generally looks good :+1:

Would like to see tests, if for no other reason than knowing it supports this case. 😄

InterfaceStubGenerator.Core/InterfaceStubGenerator.cs
clairernovotny: I was copying the sample code you had :)

PR #1097, opened 2021-03-01
Update Meow Sample To .NET 5 and .NET Standard 2.1

khalidabuhakmeh:
This also adds a .NET Standard 2.1 project
where it can be shared.


samples/global.json
```diff
@@ -0,0 +1,6 @@
+{
```
clairernovotny, line 1: Lets get rid of the global.json; we shouldn't need it.

khalidabuhakmeh: I can delete it though since I was at least able to isolate the issue. Let me know and I'll take the appropriate action.

clairernovotny: I think it'd be best to remove it

Issue #1099, opened 2021-03-02
Refit 6.0.x: Content-Length not set for PATCH requests


softmatters:
Hi

It seems like the following issue is related to a slightly old issue i.e. #885 and probably #372 

**Description**

I am consuming a third-party REST API using .NET 5/C# 9.0 and Refit version 6.0.24. The third-party system doesn't support chunked encoding. When making a PATCH call the `Transfer-Encoding: chunked` header is getting added, and the API returns with 400 status code.

The interface method looks like below:
```
[Patch("/states/{Id}"), Headers("Content-Type: application/json")]
Task<ApiResponse<State>> ChangeState([Body] StateControl stateControl, string id);
```
I have tried changing the `[Body]` attribute to all possible documented values without any success

```
- [Body]
- [Body(true)]
- [Body(BodySerializationMethod.Default, true)]
- [Body(BodySerializationMethod.Serialized)]
- [Body(BodySerializationMethod.Serialized, true)]
```
Following is the request and response captured by Fiddler

Request
---
```
PATCH http://thidparty.api.com:5000/states/0 HTTP/1.1
Host: thidparty.api.com:5000
X-Correlation-ID: 3edcc52b-aec1-4a8c-b725-c646fc269e95
traceparent: 00-4f869cf85a015f4fae43466700c2a395-5effe2f12308e444-00
Transfer-Encoding: chunked
Content-Type: application/json

1C
{"state":"new","force":true}
0
```
Response
---
```
HTTP/1.1 400 BAD REQUEST
server: gunicorn/19.4.5
date: Mon, 01 Mar 2021 23:29:51 GMT
content-type: application/json
content-length: 92
x-sentry-id: 3b2b0bad862a4eed9d62e8e76c96a7f0

{"message": "The browser (or proxy) sent a request that this server could not understand."}
```

**Expected behavior**
The expected behavior should be how it was in the version prior to 6.0.x. i.e. Content-Length header should be set. It works fine in version 5.2.4, without changing the [Body] attribute. Following is the request and response captured by Fiddler when using Refit version 5.2.4

Request
---
```
PATCH http://thidparty.api.com:5000/states/0 HTTP/1.1
Host: thidparty.api.com:5000
X-Correlation-ID: 3edcc52b-aec1-4a8c-b725-c646fc269e95
traceparent: 00-1b3b2ddae1c35c4f9ae2b6c6e7c72093-ce1d3605b1a6b241-00
Content-Length: 28
Content-Type: application/json

[...]

PR #1115, opened 2021-03-11
Really Buffered JsonContent when buffered is set

vbornand:

If fixes the bug #1099.

When the body is serialized (with JsonContent), the data is not buffered because the JsonContent doesn't implement the `TryComputeLength` method (https://github.com/dotnet/runtime/blob/9c3fc566eb916434f8bfe6974e4ddc21ba6e7ff6/src/libraries/System.Net.Http.Json/src/System/Net/Http/Json/JsonContent.cs#L55)

The settings `RefitSettings.Buffered` is only used when the `Body` attribute is set without the property `Buffered` defined and has no effect when the `Body` attribute missing. The buffering is always disabled when the attribute is missing.

**What is the new behavior?**

The request content is buffered, if the buffering is active.

The `RefitSettings.Buffered` is by default equals to false, and have effect even if the Body property is missing.

**What might this PR break?**
The `RefitSettings.Buffered` have now false as default value, so the attribute `Body` is without buffering by default.
If `RefitSettings.Buffered` is explicitely set to true, all the body are buffered even if the Body attribute is missing.

Some users can belived they buffered and it was not the case, so with this fix the memory usage can increase.

- [ ] Tests for the changes have been added (for bug fixes / features) => not done. Hard to find how to test it but all the current tests are not broken.
- [ X] Docs have been added / updated (for bug fixes / features) => The doc was incorrect. "By default, Refit streams the body content without buffering it." It was not totally true.

**Other information**:
The use of  `PushStreamContent` (https://github.com/reactiveui/refit/blob/64d91012c64057ea0fc6c3cb308bf3468a392973/Refit/RequestBuilderImplementation.cs#L576) when buffering is disabled is maybe useless when it is a `JsonContent` because it already streams the data. I didn't change this because I don't know Refit enough well to know the other implication.

### DIFF
--- a/Refit.Tests/RequestBuilder.cs
+++ b/Refit.Tests/RequestBuilder.cs
@@ -104,7 +104,7 @@ namespace Refit.Tests
         Task<string> FetchSomeStuffWithDynamicRequestPropertyWithoutKey(int id, [Property] object someValue, [Property("")] object someOtherValue);
 
         [Post("/foo/{id}")]
-        Task<bool> OhYeahValueTypes(int id, [Body] int whatever);
+        Task<bool> OhYeahValueTypes(int id, [Body(buffered: true)] int whatever);
 
         [Post("/foo/{id}")]
         Task<bool> OhYeahValueTypesUnbuffered(int id, [Body(buffered: false)] int whatever);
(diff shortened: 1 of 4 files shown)


dnfadmin:
[CLA assistant check](https://cla.dotnetfoundation.org/reactiveui/refit?pullRequest=1115) <br/>Thank you for your submission, we really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla.dotnetfoundation.org/reactiveui/refit?pullRequest=1115) before we can accept your contribution.<br/><br/>:x: vbornand [sign now](https://cla.dotnetfoundation.org/reactiveui/refit?pullRequest=1115)<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](https://cla.dotnetfoundation.org/check/reactiveui/refit?pullRequest=1115) it.</sub>

Issue #1160, opened 2021-04-30
feature: support returning ApiResponse in case of deserialization exception


toanxyz:
**Is your feature request related to a problem? Please describe.**

We're evaluating the usage of Refit in our project. One of our concerns is how to have general exception handling when using Refit. Refit will not throw the exception in the case of ApiResponse, this is good. However, in the case of the Deserialization process, Refit throws the exception directly to the application ([refer here](https://github.com/reactiveui/refit/blob/0baec2ea6f462264696fe476a88dfc267b527ad7/Refit/RequestBuilderImplementation.cs#L299)) so it could make us have try/catch everywhere or we have to create a wrapper class that handles the exception and can lead to many boilerplate or unnecessary code.

**Describe the solution you'd like**

I would like to have something similar in another method that will check the response is ApiResponse, if yes we can create that instead of throwing the exception.

**Describe alternatives you've considered**

N/A

**Describe suggestions on how to achieve the feature**

N/A

**Additional context**

N/A

james-s-tayler:
Looking at the code I see exactly the problem, and you're right.

      if (typeof(T) != typeof(HttpResponseMessage))
                    {
                        e = await settings.ExceptionFactory(resp).ConfigureAwait(false);
                    }

                    
                    if (restMethod.IsApiResponse)
                    {
                        // Only attempt to deserialize content if no error present for backward-compatibility
                        var body = e == null
                            ? await DeserializeContentAsync<TBody>(resp, content, ct).ConfigureAwait(false)
                            : default;

                        return ApiResponse.Create<T, TBody>(resp, body, settings, e as ApiException);
                    }
                    else if (e != null)
                    {
                        disposeResponse = false; // caller has to dispose
                        throw e;
                    }
                    else
                        return await DeserializeContentAsync<T>(resp, content, ct).ConfigureAwait(false);


Both of these call `DeserializeContentAsync` and that method wraps a `try/catch` around the deserialization attempt and even creates an ApiException in it's catch block that it then simply just always rethrows, which is not ideal. 

Probably what we can do is something like this:

       if (restMethod.IsApiResponse)
                    {
                        // Only attempt to deserialize content if no error present for backward-compatibility
                        var body = default(TBody);
                        try
                        {
                            body = e == null
                            ? await DeserializeContentAsync<TBody>(resp, content, ct).ConfigureAwait(false)
                            : default;
                        }
                        catch(ApiException ex)
                        {
                            e = ex;
                        }

[...]

Issue #1017, opened 2021-01-02
feature: Access Full Response Message with Serialization Exceptions


j-bbr:
**Is your feature request related to a problem? Please describe.**

When using refit to access an api that is not well documented or changes often results in Serialization Exceptions. Imagine the following Refit interface
`[Get("/myclass/{id}")]
     Task<ApiResponse<MyClass>> Get( Guid id);`
When the api changes or a property in MyClass has a typo a serialization exception will occur. Even if I catch the Exception I can't access the full response message to find the errors in either my code or with the Api.

**Describe the solution you'd like**

I would like the serialization exception to be handled in such a way that the full response message is preserved and passed along. Arguably this should be with ApiResponse as it might be due to an error with the remote api although it see the problem with missing status code etc. 

**Describe alternatives you've considered**

- Change the return type to HttpResponseMessage (Real return type becomes unclear for people that haven't written the client)
- A "wrapper" content serializer that puts a try-catch around the actual  serialization (My current solution)

```
        private class WrappingSerializer : IContentSerializer
        {
            private readonly IContentSerializer _serializer;

            public WrappingSerializer(IContentSerializer serializer)
            {
                _serializer = serializer;
            }

            //nothing useful to add to the serialization exception in this case
            public async Task<HttpContent> SerializeAsync<T>(T item) => await _serializer.SerializeAsync(item);

            public async Task<T> DeserializeAsync<T>(HttpContent content)
            {
                var stringContent = await content.ReadAsStringAsync();
                try
                {
                    return await _serializer.DeserializeAsync<T>(content);
                }
                catch (Exception serializerException)
                {
                    var message =
                        new StringBuilder($"Serialization failed with a {serializerException.GetType()}")
                            .AppendLine("Original Content was")
                            .Append(stringContent)
                            .ToString();
                    throw new SerializationException(message, serializerException);
                }
            }
        }
```

**Describe suggestions on how to achieve the feature**

I think the above solution with the wrapping serializer would be ok if we didn't have to read the content twice in the no error case. Unfortunately the content is already consumed when the exception occurs but maybe accessing the original Response is easier from within Refit

**Additional context**

glennawatson:
[...]

PR #1167, opened 2021-05-08
Support returning api response in case of deserialization exception

james-s-tayler:
This is a bug fix for #1160 and #1017 - ApiResponse<T> wasn't trapping exceptions that occurred while attempting to deserialize the response, which violated the expectations of the ApiResponse<T> programming model.

The current behavior is that an ApiResponse<T> will throw a Refit.ApiException if an exception is thrown while attempting to deserialize the response.

**What is the new behavior?**
The deserialization exception is properly wrapped by a Refit.ApiException and returned in the ApiResponse<T>

**What might this PR break?**
I've added tests for this and ensured the existing tests for the non ApiResponse<T> case still pass. If callers knew about the current behavior and had explicit try/catch blocks those will no longer be invoked anymore. Though, all callers will almost certainly have existing exception handling logic for the ApiResponse<T> which should pick up the deserialization exceptions now too. I, for one, didn't realize the library behaved as it does at present and expected every and all exception to come back via ApiResponse.Error.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1131,8 +1131,43 @@ public class HomeController : Controller
 ```
 
 ### Handling exceptions
+Refit has different exception handling behavior depending on if your Refit interface methods return `Task<T>` or if they return `Task<ApiResponse<T>>`.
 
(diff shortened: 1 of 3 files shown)


james-s-tayler:
I will add one more commit to this PR to update the docs with instructions on how the ApiResponse<T> programming model behaves with respect to trapping errors and wrapping them in ApiException and populating them onto the ApiResponse.Error field.

PR #898, opened 2020-04-08
BenchmarkDotNet profiler

Sergio0694:
**What kind of change does this PR introduce?**

This PR adds a profiler project using `BenchmarkDotNet`.

**What is the current behavior?**

There is no built-in profiler.

**What is the new behavior?**

There is a profiler to benchmark some REST calls on .NET Core 2.1 and 3.1.

**What might this PR break?**
Nothing.

**Please check if the PR fulfills these requirements**
- [X] Tests for the changes have been added (for bug fixes / features)
- [X] Docs have been added / updated (for bug fixes / features)

**Other information**:
cc. @clairernovotny, this PR is still marked as draft for now. I can't figure out how to make the project build with `BenchmarkDotNet` when `refit` is directly referenced from the local projects and not through NuGet, could you help with that? I copied/modified the `.csproj` code from the test project and I can get the profiler to build and run correctly if I run it directly from VS and try to execute one of the benchmarking methods, but if I use just `dotnet Refit.Profiler.dll` and let `BenchmarkDotNet` do its thing, it fails to run the benchmark with an error saying the interface in use is not a valid `refit` interface, which shows the build task is not run correctly I guess? Any help here is welcome 😄

cc. @Dreamescaper - feel free to help out with new benchmarks to add!

**NOTE:** this PR needs to be merged in before https://github.com/reactiveui/refit/pull/897 so that we can then run comparative benchmarks between the two versions of `refit` in that other branch before adding those optimizations.

PR #1175, opened 2021-05-16
Add benchmark dotnet and comprehensive set of end to end benchmarks

james-s-tayler:
This PR adds a full set of end-to-end benchmarks using `BenchmarkDotNet` for all of the return types Refit supports parameterized with different payload sizes (several orders of magnitude), serialization methods, http verbs, and status codes.

Inspiration was taken from the stale PR already open for this here https://github.com/reactiveui/refit/pull/898 which helped in getting passed the initial hurdles of getting it running.

Currently there are no benchmarks.

**What is the new behavior?**
We now benchmark all the things! Yay. 

**What might this PR break?**
Nothing. It's a new `csproj` added to the solution.

**Other information**:
@Sergio0694 With the original PR for this - turns out you can't set the mock expectation in the `[GlobalSetup]` method as it seems it gets garbage collected, which causes Refit to return `HTTP 404`. Once I moved that into the benchmarks themselves, it all worked. I profiled just that setup to understand what overhead it adds and it's negligible.

@clairernovotny I didn't update the docs yet with the results of the benchmarks, they are pretty hardware specific and thought you might like to select some sort of standard hardware configuration to run it on. Though, I will certainly post my set as a comment here once they finish running. Takes a fairly long time.


james-s-tayler:
hmm seems I need to make some adjustments as it is complaining about the benchmark that returns 'Task<IEnumerable<T>>'

james-s-tayler:
Interesting. I found a memory leak in the benchmarks themselves. Turns out when I moved the `mockHttp` handler expectation setup into the benchmark itself that caused a memory leak and massively, massively slowed down the benchmarks. After reading through the docs of BenchmarkDotNet I found there are `[IterationSetup]` and `[IterationCleanup]` methods. Turns out doing it in there is the correct way to do it. Benchmark is blazingly fast in comparison.

james-s-tayler:
hmm reading the docs more I'm looking at adding the necessary things in order to be able to easily control which sets of benchmarks you want to run, as the supplied set of [Params] runs an absolute tonne of benchmarks.

Refit.Benchmarks/Refit.Benchmarks.csproj
```diff
@@ -0,0 +1,31 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+    <PropertyGroup>
+        <OutputType>Exe</OutputType>
+        <TargetFramework>netcoreapp3.1</TargetFramework>
```
clairernovotny, line 5: Can this be multi-targeted to support benchmarks on .NET Core 3.1, .NET 5 & 6?

Refit.Benchmarks/StaticFileHttpResponseHandler.cs
```diff
@@ -0,0 +1,34 @@
...
+    {
+        private readonly HttpStatusCode responseCode;
+        private readonly string responsePayload;
+
+        public StaticFileHttpResponseHandler(string fileName, HttpStatusCode responseCode)
+        {
+            if (string.IsNullOrEmpty(fileName))
+                throw new ArgumentNullException(nameof(fileName));
+
+            responsePayload = File.ReadAllText(fileName);
```
clairernovotny, line 20: Should this handler be reading the file from disk each time? Or should we preload the files into memory during the setup stage and returned cached bytes?

That would prevent the tests from being affected by disk I/O.

clairernovotny (changes requested):
Looks good but a few notes

Refit.Benchmarks/Refit.Benchmarks.csproj
james-s-tayler: BenchmarkDotNet has excellent support built in for running benchmarks on different runtimes. I don't think it actually requires that we multi-target it here. I believe the two ways to do it are either configure the benchmark via code either through a custom config or simply by adding an attribute to the benchmark class for each runtime you want it to run the benchmarks for. Or it lets you supply a command line argument for which runtime you want to run the benchmarks against. The docs for the latter are here https://benchmarkdotnet.org/articles/guides/console-args.html#runtimes

So, that's supported as is. Just need to pass the right arguments I think. I originally had it hardcoded as an attribute on the benchmark class itself, but then that mandates running the benchmarks for all the specified runtimes which can take quite a while if you're targeting a few of them.

Refit.Benchmarks/StaticFileHttpResponseHandler.cs
james-s-tayler: This runs as part of [GlobalSetup], and this disk access happens in the constructor, so it runs only a single time, and isn't having to hit the disk and do I/O on each iteration during the benchmark. During the benchmarks it's just taking the string which it's got in memory and returning it.

clairernovotny:
Hi, just looking to see where this is at -- would love to get some benchmarks into Refit! :)

james-s-tayler:
Hi @clairernovotny sorry it's taken a while to respond. If you can provide some additional guidance on recommended changes I can look at finishing this off as it's pretty close. I responded to your initial comments but didn't get any subsequent feedback, so wasn't sure what to change to get it approved and then life happened. 

I suspect possibly including a set of shell scripts for each runtime to run the benchmarks against might do it? The framework itself handles the multi-targeting and you just need to provide the runtime as an argument, hence if there were a set of scripts we could run for benchmarking against each runtime then that would cover the multi-targeting aspect.

Refit.Benchmarks/Refit.Benchmarks.csproj
james-s-tayler: Upon playing with this it did indeed require me to multi-target it. My initial understanding of it wasn't quite right. I've added the multi-targeting and subsequent scripts to run the benchmarks for each runtime.

james-s-tayler:
@clairernovotny I've added added the multi-targeting and subsequent scripts in, so this should be good to review again. I only included `net6.0`, `net5.0` and `netcoreapp3.1` as `net461` is of course quite different from the others and I just don't have the bandwidth to work through the differences right now, but wanted to at least get the .NET Core / .NET 5+ stuff over the line and finally get some benchmarks into Refit :) 

The results are pretty interesting.

laurencee:
This was one of the first things I went to try and find to see if this would be good for my use case, as without this sort of information, it's difficult to figure out what the trade-offs are compared to hand creating a client.

Ideally the Readme would include a little table summary of the output of these sorts of tests comparing with the hand crafted or other libraries that do a similar thing (if there are any).

Issue #1156, opened 2021-04-22
feature: Allow decoration of interface methods to be used by DelegatingHandlers


dejvid-smth:
**Problem:**

I want to be able to make decisions in `DelegatingHandler` based on the client
interface method being executed. 

Specifically I want to apply retry logic to some requests. I would like to be able
to have one `DelegatingHandler` type for that and be able to somehow mark
interface methods where retry should be allowed. Deciding only based on request
method is not enough.

**Describe the solution you'd like**

One possible solution is to add another property to `HttpRequestMessage.Properties/Options`
that would contain `MethodInfo` of the method. Then I could define custom attribute that can
mark methods where retry is enabled and my `DelegatingHandler` could look for that attribute.

Another solution could be to add a new attribute like `ConstantProperty` that could be used
on method. It would work like existing `PropertyAttribute` with the difference that the value 
of the property would be specified in constructor and would be constant. Again, I could look
for that property in the `DelegatingHandler`

**Describe alternatives you've considered**

I don't really see how could this be done today in a clean way. The only way to pass additional
information to `DelegatingHandler` is through `PropertyAttribute` but that requires adding
method parameter which is not right. Caller shouldn't have to pass another value to each
method to manage how the client works internally.

**Additional context**

#1150 seems like similar request

james-s-tayler:
Hmm... I currently solve exactly this problem today using https://github.com/App-vNext/Polly which in behind the scenes actually leverages DelegatingHandlers to do its magic. 

I have some endpoints in a 3rd party API I need to integrate with that have different needs with respect to when/what to retry on, and I simply grouped stuff with the same needs into one Refit client, and grouped stuff with a different set of needs into a different Refit client, and then applied the appropriate Polly policies to each Refit client separately. That worked pretty well.

As for adding a new attribute, the additional complexity and effort required to add a new attribute is kind of high in conjunction with that not really being a very intuitive programming model, so I wouldn't think that's the right way to solve it. 

[...]

Issue #1150, opened 2021-04-09
feature: Add method name to HttpRequestMessage Properties


JustinBerger:
**Is your feature request related to a problem? Please describe.**
HttpRequestMessage.Properties is useful, but only including the interface name by default doesn't provide a lot of detail about the specific request being made.

**Describe the solution you'd like**
Include the path in the Get/Post/Etc attribute ("/path/{path}/{id}") and/or the Method name in HttpRequestMessage.Properties.
This makes it easier to add things like metrics and telemetry via DelegatingHandler without having to parse out the dynamic parts of the url.

Issue #1168, opened 2021-05-09
[BUG] Design of ExceptionFactory doesn't support Task<ApiResponse<T>> programming model


james-s-tayler:
**Describe the bug**
I noticed too when working on  #1167 that the design of `ExceptionFactory` is fundamentally broken. It's only designed to work well when interface methods return Task<T>, but it has very unexpected behavior when used with Task<ApiResponse<T>> in that if the exception factory returns null it will attempt to deserialize the response, but that may well fail with a deserialization error. Prior to #1167 it would have actually thrown that exception too. 

It seems the main use case of that feature was to override Refit's behavior of throwing exceptions on HTTP 404, which while technically an "error" code has semantics loose enough to only be an error in some cases, and not an error in others. So, being able to chose how that behaves such that it lines up with the semantics of the particular service you're consuming is *great*, but the docs and developer experience aren't the best for it. 

A couple of things wrong with it are as follows:

- Returning null from the `ExceptionFactory` will attempt to deserialize the response and technically the user has specified that there isn't an error! Except, that just isn't a strong enough guarantee the deserialization can occur successfully. So, the deserialization can fail and populate an unexpected result into `ApiException.Error` which may have knock-on effects to the calling code which may decide to log any errors present there which could generate a lot of noisy logs.
- The exception factory allows for returning any arbitrary exception class, but `ApiResponse.Create()` will try to cast it like `e as ApiException` which surely isn't correct. 
- There aren't any unit tests for the feature that make use of Task<ApiResponse<T>>
- The docs for the feature don't show a clear example of how to get HTTP 404 to behave as you need to which is the main reason I can see for the existence of the feature
- The docs show an example which simply never returns an exception ever, which I have a hard time imagining would actually be useful in a real world scenario as there are plenty of transient HTTP errors that happen and you would simple get back `null` and not really know why. 

**Steps To Reproduce**
Provide an `ExceptionFactory` that returns null on `HTTP 404` and use it in conjunction with a method that returns `Task<ApiResponse<T>>` and invoke an api returning 404. 

**Expected behavior**
I'm not exactly sure what I expect to happen in this case. I just simply don't expect it to do what it does. I'm leaving this issue as a note to self, so I can have the context to address it later, unless someone else gets to it. 

[...]

Issue #1153, opened 2021-04-15
ExceptionFactory for 404


zh355245849:
Hi, guys,
How can I use the ExceptionFactory to override 404 code with null result? I have tried below but none of these can work, it always throw exception when code is not 404:
            ExceptionFactory = httpResponse => httpResponse.StatusCode != HttpStatusCode.NotFound ? Task.FromResult<Exception?>(null) : (Task<Exception?>)Task.CompletedTask,

ExceptionFactory = httpResponse => httpResponse.StatusCode != HttpStatusCode.NotFound ? Task.FromResult<Exception?>(null) : null

Can someone guide me to the correct path? 
Thanks,
Han

glennawatson:
The signature is `Func<HttpResponseMessage, Task<Exception?>>`: return an exception to throw, or `null` to suppress and let Refit deserialize the response as normal. The pattern you want:

```csharp
var settings = new RefitSettings
{
    ExceptionFactory = async response =>
    {
        if (response.StatusCode == HttpStatusCode.NotFound)
            return null; // suppress -> method returns null/default

        return response.IsSuccessStatusCode
            ? null
            : await ApiException.Create(
                response.RequestMessage!, response.RequestMessage!.Method, response, settings);
    }
};
```

Your attempts failed because of the cast/`Task.CompletedTask` mix-up - return `Task.FromResult<Exception?>(null)` (or use `async` as above), never `(Task<Exception?>)Task.CompletedTask`. This behavior is covered by tests today (`src/Refit.Tests/ExceptionFactoryTests.cs`, `ProvideFactoryWhichAlwaysReturnsNull_WithResult` returns the body on a 404).

Closing as a usage question - please use https://github.com/reactiveui/refit/discussions for follow-ups.


<blockquote><img src="https://opengraph.githubassets.com/7da521d756b94831ff31d8d5fc75ecbe2c2d69a69928da79595847192e775278/reactiveui/refit" width="48" align="right"><div><img src="https://github.githubassets.com/favicons/favicon.svg" height="14"> GitHub</div><div><strong><a href="https://github.com/reactiveui/refit/discussions">reactiveui refit · Discussions</a></strong></div><div>Explore the GitHub Discussions forum for reactiveui refit. Discuss code, ask questions & collaborate with the developer community.</div></blockquote>

PR #1180, opened 2021-05-22
Property provider factory

james-s-tayler:
This is a new feature that extends/enhances the work done around the `[Property]` attribute. 
It adds the notion of a `PropertyProvider` which allows adding properties into `HttpRequestMessage.Properties`
on every request without explicitly having to pass in a `[Property]` attribute but instead using information derived
from the `MethodInfo` of the currently executing method on the Refit instance, as well as the `Type` of the Refit target interface.

`PropertyProviderFactory` allows you to build a `List<PropertyProvider>` to populate `RefitSettings.PropertyProviders` and 
comes with built-in support for common scenarios such as adding the refit target interface type, the method info, and any custom attributes into the `HttpRequestMessage.Properties`. 

Users can write their own `PropertyProvider` implementations by simply implementing the `PropertyProvider` interface and passing into the `PropertyProviderFactory` when configuring `RefitSettings.PropertyProviders`.

The current behavior is you can only pass `[Property]` attributes as part of the method signature.

Community feedback on the `[Property]` feature seems to indicate the ability to pass properties is something
useful, but the current functionality doesn't quite support some great use cases.

See https://github.com/reactiveui/refit/issues/1156 and https://github.com/reactiveui/refit/issues/1150

**What is the new behavior?**
Users can leverage the new `PropertyProvider` implementation as well as write their own in order to extend the functionality of Refit via passing state to `HttpClient` middleware.

**What might this PR break?**
It touches a core part of `HttpRequestBuilderImplementation` as such it has been programmed defensively, so as not to blow up if anything goes wrong with a user provided implementation of `PropertyProvider`. It has a fairly extensive set of tests that include this scenario too to ensure the request can still succeed.

**Other information**:
This PR introduces a new namespace `Refit.Extensions` to aide in discoverability. It also shifts the existing `ExceptionFactory` stuff into that namespace. I hope that's not too invasive. If there's any changes you'd rather see backed out in that area, just let me know.


clairernovotny:
This is a big PR and will need some time to review.

clairernovotny:
One initial thought here is that this seems quite complicated. Why do we need more than the method name as per #1150?

james-s-tayler:
It's so that consumers of the library can grow its functionality to support their use cases without being blocked on anyone. 

Also means Refit itself doesn't have to have too strong am opinion of its own on what winds up in the request properties.

And it's done in a way that's composible such that consumers of the library don't have to accept or reject the default opinion, but can instead choose to augment it, or replace it.

Personally, I myself don't have a need for the method info, but can definitely see use cases others might have around that. I'm most excited about and in need of flowing data from custom attributes into the request properties as that opens a very powerful extension point that allows for pretty much anything.

anaisbetts:
Hey @james-s-tayler, while it's super clear that you've done a ton of work on this and definitely crossed the i's and dotted the t's, I think the design of this change is Too Complicated. It has breaking changes (changing the base class of an exception, moving types to other namespaces), and overall it introduces too much complexity.

I think that we should start instead with directly implementing this comment from #1150: 

> Include the path in the Get/Post/Etc attribute ("/path/{path}/{id}") and/or the Method name in HttpRequestMessage.Properties.
This makes it easier to add things like metrics and telemetry via DelegatingHandler without having to parse out the dynamic parts of the url.

Let's do this First, then see what scenarios are still painful to do and think about how to solve them. Parsing string URLs might not be as Elegant as a fancy MethodInfo and Provider setup but like, nearly everyone given the URL and Method will be like "I understand what to do next to Solve My Problem (even if it's writing a bunch of ifs and Regexes)".

lostincomputer:
My business problem is I want to be able to configure Polly without needing to add an argument to the method. For example:

```
interface IApiClient
{
   [MaxRetry(3)]
   string DoSomething();

   [MaxRetry(1)]
   string DoAnotherThing();
}

```

james-s-tayler:
@anaisbetts The design is targeted more at solving #1156. I've had plenty of uses for this in the past. @lostincomputer mentions one particular case, but there are others...

Perhaps you have a versioned API and want to be able to do something like:

    interface IApiClient
    {
       [Version(1)]
       [Get("/Something")]
       string GetSomethingV1(sring id);

       [Version(2)]
       [Get("/Something")]
       string GetSomethingV2(string id);
    }

Or maybe you simply want a declarative way to output messages under certain conditions on certain endpoints
without littering the code base with a lot of custom logic at each of the callsites like:

interface IApiClient
{
   [SuccessMessage("Yay - we did the thing!")]
   [ErrorMessage("Oops - looks like we couldn't do the thing :(")]
   [Post("/Something")]
   string DoSomething([Body] object something);

   [SuccessMessage("Yay - we did the other thing!")]
   [ErrorMessage("Oops - looks like we couldn't do the other thing :(")]
   [Put("/Something")]
   string DoTheOtherThing([Body] object somethingEle);
}

Could probably support https://github.com/reactiveui/refit/issues/163 too by something like

[SomeData("123")]
interface IApiClient
{
   [Post("/Something")]
   string DoSomething([Body] object something);

   [Put("/Something")]
   string DoTheOtherThing([Body] object somethingEle);
}

These particular use cases aren't something Refit would likely want to directly support,
but this makes it extensible such that the developer can solve whatever problem they have
without any specific details of that having to leak into the Refit codebase itself.

I feel like solving just #1150 doesn't help with that while at the same time making Refit
more opinionated about what winds up in the request properties. Doing so may have unintended consequences as we have no knowledge of what DelegatingHandlers are present and what they do. It's better if the developer is in control of what winds up in there. It's ok for Refit to have some default opinion, but quite a bit nicer if it is optional.

I've read through several hundred issues raised on this repo and one of the overall themes that seems to repeat is certain parts of Refit not being extensible enough and users getting blocked.
Over the years certain points of extensibility have been added and each time Refit gets a bit more useful. When I saw there was already a HttpMessageHandlerFactory and an ExceptionFactory it made me feet that a PropertyProviderFactory was in line with idioms already present in the library. I think these points of extensibility make a lot of sense. They allow Refit to be grown by its users, and they allow whole classes of problems to be solved at once.

In terms of breaking changes... Yeah I wasn't 100% sure about that. I think it's just shifting DefaultApiExceptionFactory into its own class and a different namespace. I can certainly update the PR and undo that particular change. Though I was brave enough to try it since its is the default ExceptionFactory configured internally by RefitSettings it's probably not actually a breaking change to any consumers of the library, as no one would/should
actually have any need to be referencing it directly, and was considering to do some work on
https://github.com/reactiveui/refit/issues/1168 to perhaps provide an implementation of ExceptionFactory out of the box that could live in that namespace that makes it easy to suppress errors in certain status codes so users don't wind up in a situation like https://github.com/reactiveui/refit/issues/1153. I'm not married to that particular idea though, could be a better idea to finish [this PR](https://github.com/reactiveui/refit/pull/927).
But... up to you. Just let me know. Happy to back that particular part out!

Anyway, sorry for the essay, appreciate the review and would love to hear if you have any alternative suggestions on a design that can meet the needs I've outlined here. I hope I can convince you this design, though seemingly complicated, does have its merits!

Misiu:
This would be awesome.
I have multiple interfaces, all hitting the same endpoint, but with a different suffix.
My API URLs: https://example.com/api/claims, https://example.com/api/users

```
    public interface IClaimsApi
    {
        [Get("claims/")]
        Task<IEnumerable<Claim>> GetAll();

        [Get("claims/{claimId}")]
        Task<ApiResponse<Claim>> Get(int claimId);
    }
```

```
    public interface IUsersApi
    {
        [Get("users/")]
        Task<IEnumerable<User>> GetAll();

        [Get("users/{userId}")]
        Task<ApiResponse<User>> Get(int userId);
    }
```

This could be easily simplified to this:

```
    [Prefix("claims") ]
    public interface IClaimsApi
    {
        [Get("/")]
        Task<IEnumerable<Claim>> GetAll();

        [Get("/{claimId}")]
        Task<ApiResponse<Claim>> Get(int claimId);
    }
```

```
    [Prefix("users") ]
    public interface IUsersApi
    {
        [Get("/")]
        Task<IEnumerable<User>> GetAll();

        [Get("/{userId}")]
        Task<ApiResponse<User>> Get(int userId);
    }
```

PR #1190, opened 2021-06-17
Fix Refit exception when getting 204 NoContent - Close #1128

### LINKED ISSUE #1128: Refit.Exception when deserializing using System.Text.Json
When using Refit with System.Text.Json and the SystemTextJsonContentSerializer the following error comes up:
Refit.ApiException: An error occured deserializing the response.
 ---> System.Text.Json.JsonException: The input does not contain any JSON tokens. Expected the input to start with a valid JSON token, when isFinalBlock is true. Path: $ | LineNumber: 0 | BytePositionInLine: 0.
 ---> System.Text.Json.JsonReaderException: The input does not contain any JSON tokens. Expected the input to start with a valid JSON token, when isFinalBlock is true. LineNumber: 0 | BytePositionInLine: 0.
   at System.Text.Json.ThrowHelper.ThrowJsonReaderException(Utf8JsonReader& json, ExceptionResource resource, Byte nextByte, ReadOnlySpan`1 bytes)
   at System.Text.Json.Utf8JsonReader.Read()
   at System.Text.Json.Serialization.JsonConverter`1.ReadCore(Utf8JsonReader& reader, JsonSerializerOptions options, ReadStack& state)

This happens when the API returns StatusCodeResult (eg. return NoContent(), return StatusCode(204).
When returning StatusCodeResult the HttpContentHeaders in HttpContent are empty and the MediaTypeHeaderValue is null and the content itself has no content to deserialize and an exception in System.Text.Json is thrown.
image

My Suggestion is to change the FromHttpContentAsync method to:
`public async Task<T?> FromHttpContentAsync<T>(HttpContent content, CancellationToken cancellationToken = default)
        {
            if(content.Headers.ContentType != null && content.Headers.ContentType.MediaType == "application/json")
            {
                var item = await content.ReadFromJsonAsync<T>(jsonSerializerOptions, cancellationToken).ConfigureAwait(false);
                return item;
            }

            return default;
        }`

jpbroeders:
This PR solves a bug (#1128).

 When getting back an HTTP 204 NoContent, the serializer will break because there is no body and content-type property, please see #1128

**What is the new behavior?**
New code is checking if the content-type is application/json. If not, it returns the default.

**What might this PR break?**
Nothing!

**Other information**:


dnfadmin:
[CLA assistant check](https://cla.dotnetfoundation.org/reactiveui/refit?pullRequest=1190) <br/>Thank you for your submission, we really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla.dotnetfoundation.org/reactiveui/refit?pullRequest=1190) before we can accept your contribution.<br/><br/>:x: jpbroeders [sign now](https://cla.dotnetfoundation.org/reactiveui/refit?pullRequest=1190)<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](https://cla.dotnetfoundation.org/check/reactiveui/refit?pullRequest=1190) it.</sub>

clairernovotny:
Unfortunately this fix is more complicated than here. We can't return `null` or `default` for non-nullable return types. A 204 returns null, effectively. That means we have to detect if the return type is non-nullable or nullable so we can either return null or throw. 

Another workaround is to use `ApiResponse<>` or `HttpResponseMessage` return types where you can check the status before deserializing the object.

jpbroeders:
Thanks for the answer.
Because this is my first “contribution” in someones repo, I don’t know what to do with the pull request.
Should I close it?

clairernovotny (changes requested):
See my comment above. This issue is much more complicated.

clairernovotny:
This PR can be closed, thanks.

Thorarin:
@clairernovotny That check cannot be performed at runtime, as far as I know.
At runtime, it's perfectly possible to assign null to a "non-nullable" reference type. 
Obviously, if your API can return empty content , you should use a nullable reference type in your Refit interface (if you're using #nullable enable), so the runtime behavior matches the type.

Refit.Newtonsoft.Json/NewtonsoftJsonContentSerializer.cs
```diff
@@ -45,8 +45,11 @@ public HttpContent ToHttpContent<T>(T item)
         }
 
         /// <inheritdoc/>
-        public async Task<T?> FromHttpContentAsync<T>(HttpContent content, CancellationToken cancellationToken = default)
+        public async Task<T?> FromHttpContentAsync<T>(HttpContent content, HttpResponseMessage? responseMessage, CancellationToken cancellationToken = default)
         {
+            if (responseMessage?.StatusCode == System.Net.HttpStatusCode.NoContent)
```
AbdullahAlRana, line 50: Interesting, I am wondering what if responseMessage is null

PR #1215, opened 2021-08-05
Use bait and switch (shim) packaging to support multiple versions of Roslyn

sharwell:
* The main source generator assembly is a "shim assembly" that just contains an empty generator implemented through a base type. The shim assembly contains a module initializer which examines the other assemblies currently loaded in the app domain, and from those assemblies determines the currently running compiler version. Based on the compiler version, it loads the target assembly containing the base type to match the current compiler version. It works because the module initializer is evaluated before the base type is evaluated.

* The assemblies containing the base type are added to the shim assembly as embedded resources. This prevents the target assemblies from getting picked up by the compiler as separate source generator assemblies, and ensures they will be located by step (1) even if the source generator assembly is shadow copied prior to load.


InterfaceStubGenerator.Shim/ModuleInitializer.cs
```diff
@@ -0,0 +1,43 @@
...
+namespace Refit.Generator
+{
+    internal static class ModuleInitializer
+    {
+        internal static void Initialize()
+        {
+            try
+            {
+                var codeAnalysis = AppDomain.CurrentDomain.GetAssemblies().FirstOrDefault(
+                    assembly => assembly.GetName().Name == "Microsoft.CodeAnalysis");
```
sharwell, line 15: 💭 I suppose this could just use `typeof(GeneratorAttribute).Assembly`

InterfaceStubGenerator.Shim/ModuleInitializer.cs
```diff
@@ -0,0 +1,43 @@
...
+                var codeAnalysisVersion = codeAnalysis.GetCustomAttribute<AssemblyFileVersionAttribute>()?.Version;
+                if (!Version.TryParse(codeAnalysisVersion, out var version))
+                {
+                    return;
+                }
+
+                var implementationName = version switch
+                {
+                    { Major: >= 4 } => "InterfaceStubGenerator.Roslyn40.dll",
+                    _ => "InterfaceStubGenerator.Roslyn38.dll",
```
sharwell, line 25: 📝 The implementations all have different embedded resource names so they can be loaded by name, even though their assembly identities are the same.

InterfaceStubGenerator.Shim/ModuleInitializer.cs
```diff
@@ -0,0 +1,43 @@
+﻿using System;
+using System.IO;
+using System.Linq;
+using System.Reflection;
+
+namespace Refit.Generator
+{
+    internal static class ModuleInitializer
+    {
+        internal static void Initialize()
```
chsienki, line 10: You need to apply the module initialize attribute.

sharwell: ➡️ This is fixed

sharwell:
This approach worked for Visual Studio extensions, but I haven't been able to get it to work here.

PR #1216, opened 2021-08-07
Support Roslyn 3.8 and Roslyn 4.0 source generator scenarios

sharwell:
* [x] Split InterfaceStubGenerator.Core into one project for Roslyn 3.8 and a second project for Roslyn 4.0
* [x] Update packaging to include both source generators in the refit package
* [x] Add build logic to selectively include the Roslyn 4.0 source generator for design-time scenarios in Visual Studio 2022
  * Design time builds in Visual Studio 2019 use the Roslyn 3.8 implementation.
  * Design time builds in Visual Studio 2022 use the Roslyn 4.0 implementation.
  * All command line builds use the Roslyn 3.8 implementation. Prior to dotnet/roslyn#55023, there is no significant performance penalty for using the Roslyn 3.8 implementation of the source generator. We can decide how to handle command line scenarios later.
* [x] Add tests verify the V1 and V2 implementations produce the same output
* [x] Convert the V2 implementation to use incremental generation instead of SyntaxReceiver
  * [x] As a separate change, show how the generator can improve incrementality by collecting an intermediate value to its own value source

PR #1222, opened 2021-08-19
Change folder path to avoid implicit NuGet inclusions

sharwell:
@Mike-E-angelo I'm hoping we can use something like this PR to avoid needing to move these files completely outside the `analyzers` folder.

Fixes https://github.com/reactiveui/refit/pull/1216#issuecomment-902132236

### DIFF
--- a/Refit/Refit.csproj
+++ b/Refit/Refit.csproj
@@ -28,12 +28,12 @@
     <None Include="targets\refit.targets" PackagePath="build\netstandard2.0" Pack="true" />
 
     <None Include="..\InterfaceStubGenerator.Roslyn38\bin\$(Configuration)\netstandard2.0\InterfaceStubGeneratorV1.dll"
-          PackagePath="analyzers\cs\roslyn38"
+          PackagePath="build\analyzers\roslyn38"
           Pack="true"
           Visible="false" />
 
     <None Include="..\InterfaceStubGenerator.Roslyn40\bin\$(Configuration)\netstandard2.0\InterfaceStubGeneratorV2.dll"
-          PackagePath="analyzers\cs\roslyn40"
+          PackagePath="build\analyzers\roslyn40"
           Pack="true"
           Visible="false" />
   </ItemGroup>
--- a/Refit/targets/refit.props
+++ b/Refit/targets/refit.props
@@ -7,12 +7,12 @@
   <Choose>
     <When Condition="$([MSBuild]::VersionGreaterThanOrEquals($(VisualStudioVersion), '17.0')) AND '$(DesignTimeBuild)' == 'True'">
       <ItemGroup>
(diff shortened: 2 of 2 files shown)


Mike-E-angelo:
Yeah it appears that any reference to this:

```
	<ItemGroup>
        <Analyzer Include="$(MSBuildThisFileDirectory)../../analyzers/cs-roslyn40/*.dll" />
    </ItemGroup>
```

Generates exceptions at build time.  I am curious on why removing it still loads it into `ServiceHub.RoslynAnalysisService.exe`.  If NuGet is auto-loading this into `ServiceHub.RoslynAnalysisService.exe` shouldn't we remove this `ItemGroup` altogether?

sharwell:
OK this is going to be more annoying than I expected. I have to move these assemblies completely outside the analyzers folder. I'll update this PR.

https://github.com/dotnet/NuGet.BuildTasks/blob/b2f0c162994237c4a9c23af79d95f79847499ebd/src/Microsoft.NuGet.Build.Tasks/ResolveNuGetPackageAssets.cs#L468-L477

jmarolf (approved):

clairernovotny:
Should this be merged?

clairernovotny:
I'm going to go with yes :) I was confused by the close/reopen.

sharwell:
> I was confused by the close/reopen.

Build had issues the first time; close/reopen triggered a new one.

Mike-E-angelo:
Woohoo... I can confirm that this works like a champ now with preview5.  Thank you so much for doing this!  It really means a lot to me personally as there are a lot of fires burning in my world and this makes one less of them. :)  I appreciate the dedication to ensuring high quality of your work out there. 👍🙏

clairernovotny:
@sharwell should I release this as stable?

sharwell:
@clairernovotny the specific risk is as follows:

* If a consumer uses this package outside of NuGet tooling, or in a manner that doesn't add a source generator assembly from the customized location, `InterfaceStubGenerator` will not execute. If the build is likely to still succeed in this scenario (i.e. no compiler errors due to missing generated code), it's possible this user would experience a silent breaking change.
* For a while, the incremental source generator APIs were considered experimental and were only enabled if a consumer manually enabled `<LangVersion>preview</LangVersion>` in their project. If this behavior has not changed, it's possible for a user to experience a silent break inside the IDE for Visual Studio 2022.

All other potentially problematic cases should consistently use the old implementation (Roslyn 3.8 APIs) during builds, so even if there are bugs in the Roslyn 4.0 portion it should not manifest in the build outputs. Once confidence is built in the Roslyn 4.0 implementation, we can

sharwell:
https://github.com/dotnet/roslyn/blob/c083dae9fcb112d82c18b89687b9af65e7ef77b7/src/Compilers/CSharp/Portable/SourceGeneration/CSharpGeneratorDriver.cs#L27

PR #1251, opened 2021-10-07
Switch to Roslyn component versioning

### LINKED ISSUE #1244: [BUG] Roslyn 4.0 Source Code Generator Does Not Load
**Describe the bug**
This is in reference to this PR:
https://github.com/reactiveui/refit/pull/1216

After installing `6.1.11-preview` and attempting to load in Visual Studio 2022 4.1, the source code generator does not load as expected:

I have traced it to the following condition [found on this line](https://github.com/reactiveui/refit/blob/c38e7be8fdf61f9f5dbbd395fbf3d84762f0b662/Refit/targets/refit.props#L9):

`AND '$(LangVersion)' == 'preview'`

If I remove this condition, the next problem is presented.

It appears that there is an issue with `*.dll`.  This doesn't work:

```
<Analyzer Include="$(MSBuildThisFileDirectory)../../build/analyzers/cs/roslyn40/*.dll" />
```

But this does:

```
<Analyzer Include="$(MSBuildThisFileDirectory)../../build/analyzers/cs/roslyn40/InterfaceStubGeneratorV2.dll" />
```

**Steps To Reproduce**

You can use the SLN found here:
https://github.com/DragonSpark/Framework/tree/projects/Starbeam

And [upgrade this project](https://github.com/DragonSpark/Framework/blob/projects/Starbeam/DragonSpark.Application/DragonSpark.Application.csproj) to `6.1.11-preview`.

You can see on [this line](https://github.com/DragonSpark/Framework/blob/projects/Starbeam/DragonSpark.Application/DragonSpark.Application.csproj#L5) that `LangVersion` is set to `preview`, so this _should_ work, but doesn't.

**Expected behavior**
The Source Generator loads and all is happy. :)

**Screenshots**

NA/above.

**Environment**

- OS:       Windows 10
- Device:   Desktop
- Version:   `6.1.11-preview`
- Working Version: `6.1.11-preview`

**Additional context**

I would submit a PR but I am unsure why the `LangVersion` doesn't work.  Better to have @sharwell take another crack at this. :)

sharwell:
* Use Roslyn 3.8 assemblies for .NET SDK prior to 6.0 RC 2 (where [component versioning](https://github.com/dotnet/sdk/issues/20355) was added)
* Use matching Roslyn version for 6.0 RC 2 and newer
* Allow users to disable the source generator by setting `DisableRefitSourceGenerator` to true (derived from [MultiTargetRoslynComponent.targets](https://github.com/dotnet/runtime/blob/main/eng/MultiTargetRoslynComponent.targets.template))

Closes #1244

### DIFF
--- a/Refit/Refit.csproj
+++ b/Refit/Refit.csproj
@@ -28,12 +28,12 @@
     <None Include="targets\refit.targets" PackagePath="build\netstandard2.0" Pack="true" />
 
     <None Include="..\InterfaceStubGenerator.Roslyn38\bin\$(Configuration)\netstandard2.0\InterfaceStubGeneratorV1.dll"
-          PackagePath="build\analyzers\roslyn38"
+          PackagePath="analyzers\dotnet\roslyn3.8\cs"
           Pack="true"
           Visible="false" />
 
     <None Include="..\InterfaceStubGenerator.Roslyn40\bin\$(Configuration)\netstandard2.0\InterfaceStubGeneratorV2.dll"
-          PackagePath="build\analyzers\roslyn40"
+          PackagePath="analyzers\dotnet\roslyn4.0\cs"
           Pack="true"
           Visible="false" />
   </ItemGroup>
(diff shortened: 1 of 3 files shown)


sharwell:
> Is that needed for .NET 5 / earlier SDK's to not break?

Yes, otherwise earlier SDKs will be back in the place that led to #1222

Issue #1274, opened 2021-11-10
feature: Add Xml documentation for public methods


laurentksh:
**Is your feature request related to a problem? Please describe.**

I recently started using this library and I found the lack of Xml documentation to be disorienting (compared to other OSS libraries and most of the .NET libraries).
For example, I had to check the source code to know exactly what exception ApiResponse<T>.EnsureSuccessStatusCodeAsync() is throwing.

**Describe the solution you'd like**

I would like to add Xml docs. to various public methods and classes in the library.

**Describe alternatives you've considered**

I guess the alternative would be to leave the code as-is.

**Describe suggestions on how to achieve the feature**

Add Xml documentation to public methods and classes.

**Additional context**

I already started working on a Pull request to add Xml documentation on my fork of the repo.
I really like this library and I think these comments would be a good addition.

Best regards,

clairernovotny:
Documentation is always very welcome!

laurentksh:
@clairernovotny glad to hear that ! :)

PR #1275, opened 2021-11-10
Add Xml documentation to exposed methods and classes.

laurentksh:

Xml Docs on publicly exposed classes and methods.

Most of the classes and methods you would use when using Refit have no Xml docs. Related issue: https://github.com/reactiveui/refit/issues/1274

**What is the new behavior?**

Developers using the library will now see comments when typing a class or method name, hovering over them, etc.

**What might this PR break?**
I can't see how adding comments could break something so nothing afaik.

- [x] Tests for the changes have been added (for bug fixes / features). Checked but doesn't apply to this PR.

**Other information**:

Let me know if I have missed something or if you see a comment that should be worded differently/is wrong !

### DIFF
--- a/Refit/ApiException.cs
+++ b/Refit/ApiException.cs
@@ -6,19 +6,60 @@ using System.Threading.Tasks;
 
 namespace Refit
 {
-
+    /// <summary>
+    /// Represents an error that occured while sending an API request.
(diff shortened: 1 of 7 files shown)


clairernovotny:
Thanks!

Issue #1277, opened 2021-11-28
feature: Built-in support for the .NET 6 JSON source generator


martincostello:
**Is your feature request related to a problem? Please describe.**

The new [JSON source generator](https://devblogs.microsoft.com/dotnet/try-the-new-system-text-json-source-generator/) in .NET 6 allows applications to use a source generator to produce more performant (de)serialization code for JSON that the default reflection-based approach.

The default Refit `IHttpContentSerializer` implementation for System.Text.Json does not use/support this feature as it is today.

**Describe the solution you'd like**

A built-in implementation of `IHttpContentSerializer` that uses the `JsonSerializerContext` class and the relevant new `JsonSerializer` methods added in .NET 6 so applications can quickly plug the JSON source generator into Refit with minimal ceremony/effort.

**Describe alternatives you've considered**

Applications write their own implementation of `IHttpContentSerializer` to achieve this as needed.

**Describe suggestions on how to achieve the feature**

[This gist](https://gist.github.com/martincostello/53b10574fbee2d8bc7b0f75bf0855a84) shows a potential way to achieve this in code as a standalone type for illustrative purposes. It could potentially be built into the existing `SystemTextJsonContentSerializer` class instead via a new constructor overload, rather than creating a new type.

martincostello:
I'm going to close this, as it turns out I'd missed a new feature as part of the JSON source generator work, that lets you bind a context to an options so it is used for type metadata when (de)serializing if there's a registration associated with it: [`JsonSerializer.AddContext<TContext>()`](https://docs.microsoft.com/en-us/dotnet/api/system.text.json.jsonserializeroptions.addcontext?view=net-6.0)

Thus, the existing `SystemTextJsonContentSerializer` class should be able to work with it as-is.

clairernovotny:
@martincostello It might be worth a note in the readme about this -- can you please submit a PR with that?

martincostello:
Sure @clairernovotny - I'll open something later on today 👍

PR #1279, opened 2021-12-01
Add note about JSON source generator

martincostello:
Add an example of how to use the JSON source generator in .NET 6 with Refit (https://github.com/reactiveui/refit/issues/1277#issuecomment-983637859).

### DIFF
--- a/README.md
+++ b/README.md
@@ -357,6 +357,20 @@ public class Foo
 }
 ```
 
+##### JSON source generator
+
+To apply the benefits of the new [JSON source generator](https://devblogs.microsoft.com/dotnet/try-the-new-system-text-json-source-generator/) for System.Text.Json added in .NET 6, you can use `SystemTextJsonContentSerializer` with a custom instance of `RefitSettings` and `JsonSerializerOptions`:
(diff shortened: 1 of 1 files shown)


martincostello:
@clairernovotny Could you take a look at this when you get some free time please? 😃

clairernovotny:
Sorry, December just disappeared....

Is there a way to use the .NET 6 source generator by default for Refit?

martincostello:
Thanks for merging.

> Is there a way to use the .NET 6 source generator by default for Refit?

As far as I'm aware you have to opt into it on the `JsonSerializerOptions` instance, so you'd have to pass a pre-configured options class to the content serializer yourself.

> Is there any downsides to it?

The only downsides I'm aware of are various code-gen bugs in the source generator I've bumped into and a few other issues, many of which are going to be addressed in upcoming patch releases of 6.0 (many as soon as 6.0.2 next month as the fixes are already merged).

Assuming people's specific POCOs etc. don't run into any of those, they should benefit from more performant runtime performance if they opt-in to use it.

PR #1288, opened 2022-01-14
feat: allow custom formatting querystring keys in objects

### LINKED ISSUE #1258: feature: Add support for custom formating when flattening complex object to query parameters
It would be nice if one could specify how objects are flattened when used as parameters to GET method. For example I have a refit API definition as follows:
```csharp
public interface IApi
{
    [Get("/info")]
    Task<string> GetInfo(InfoDto dto);
}
```
where the `InfoDto` looks like this:
```csharp
public class InfoDto
{
    public string Name { get; set; }

    [Query(Format = "yyyy-M-d")]
    public DateTime Date { get; set; }

    public Size CarSize { get; set; }
}
```
the important thing to notice is that `Size` is not a regular `enum`, but instead its a custom class implementation of enumeration pattern (simplified version of https://docs.microsoft.com/en-us/dotnet/architecture/microservices/microservice-ddd-cqrs-patterns/enumeration-classes-over-enum-types) and it looks like this:
```csharp
public class Size
{
    public static Size Small = new Size("small");
    public static Size Medium = new Size("medium");
    public static Size Large = new Size("large");

    public string Key { get;  }

    private Size(string key)
    {
        Key = key;
    }
}
```

When calling the APi using refit code below
```csharp
var api = RestService.For<IApi>("http://localhost:54321");
            
var infoDto = new InfoDto { Name = "John Doe", Date = DateTime.Now, CarSize = Size.Medium };
var response = await api.GetInfo(infoDto);
```
the resulting URL looks like this `/info?Name=John%20Doe&Date=2021-10-23&CarSize.Key=medium`. Notice that `CarSize` property is flattened as `CarSize.Key` in query string.

It would be nice if one could specify how the individual properties are flattened, so it would be possible to tell the `CarSize` property what its name should be in URL query parameter. The aim in this case it so achieve that `CarSize` is flattened under name `CarSize` in URL, so the result looks like  `/info?Name=John%20Doe&Date=2021-10-23&CarSize=medium`

kamranayub:

Closes #1258 I believe! This adds the ability to customize the key handling when using complex objects. It encapsulates the default handling of looking at `[AliasAs]` and allows consumers to override or extend this behavior.

The current behavior will serialize property names in querystrings as their object property names by default. If `[AliasAs]` is used, it uses that name instead.

**What is the new behavior?**

The logic that performed the default behavior is now encapsulated into a `UrlParameterKeyFormatter` to stay consistent with the other `UrlParameterFormatter` class.

_Update:_ I noticed `BuildQueryMap` against a dictionary formats keys using `UrlParameterFormatter.Format`. It may be useful then to consolidate my impl with that or switch that to this. But that is a change in behavior since it might be looking at `Query` or enum attributes right now.

I think there is _slight_ responsibility differences between these:

```c#
var keyType = key.GetType();
var formattedKey = settings.UrlParameterFormatter.Format(key, keyType, keyType);
```

vs.

```c#
var key = settings.UrlParameterKeyFormatter.Format(propertyInfo.Name, propertyInfo);
```

By delegating to a _specific_ key handler, the consumer can know it's a query string key. With the dictionary approach, the consumer doesn't know if the value being passed is meant to be a key or a value.

I'd propose then either:

1. Introducing a `FormatKey` method onto `IUrlParameterFormatter`
2. Changing the dictionary approach to call the new key formatter (breaking change)
3. Add another intermediate step to call key formatter on the _return value_ (`formattedKey`) of the dictionary value formatter

Like so:

```c#
var keyType = key.GetType();
var keyValueFormatted = settings.UrlParameterFormatter.Format(key, keyType, keyType);
var formattedKey = settings.UrlParameterKeyFormatter.Format(keyFormattedValue, keyType);
```

But I am not positive. It's inconsistent at the moment which would be confusing I expect.

**What might this PR break?**

As-is, this will not have any breaking changes and is backwards-compatible. See above for some other approaches though.

**Other information**:

### DIFF
--- a/README.md
+++ b/README.md
@@ -42,9 +42,11 @@ services
 * [Where does this work?](#where-does-this-work)
   * [Breaking changes in 6.x](#breaking-changes-in-6x)
 * [API Attributes](#api-attributes)
-* [Dynamic Querystring Parameters](#dynamic-querystring-parameters)
-* [Collections as Querystring parameters](#collections-as-querystring-parameters)
-* [Unescape Querystring parameters](#unescape-querystring-parameters)
+* [Querystrings](#querystrings)
+  * [Dynamic Querystring Parameters](#dynamic-querystring-parameters)
+  * [Collections as Querystring parameters](#collections-as-querystring-parameters)
+  * [Unescape Querystring parameters](#unescape-querystring-parameters)
+  * [Customize Querystring parameter formatting](#customize-querystring-parameter-formatting)
 * [Body content](#body-content)
   * [Buffering and the Content-Length header](#buffering-and-the-content-length-header)
   * [JSON content](#json-content)
@@ -170,7 +172,9 @@ Search("admin/products");
 >>> "/search/admin/products"
 ```
 
-### Dynamic Querystring Parameters
+### Querystrings
+
+#### Dynamic Querystring Parameters
 
 If you specify an `object` as a query parameter, all public properties which are not null are used as query parameters.
 This previously only applied to GET requests, but has now been expanded to all HTTP request methods, partly thanks to Twitter's hybrid API that insists on non-GET requests with querystring parameters.
@@ -224,9 +228,9 @@ Task<Tweet> PostTweet([Query]TweetParams params);
 
 Where `TweetParams` is a POCO, and properties will also support `[AliasAs]` attributes.
 
-### Collections as Querystring parameters
+#### Collections as Querystring parameters
 
-Use the `Query` attribute to specify format in which collections should be formatted in query string
+Use the `Query` attribute to specify format in which collections should be formatted in querystring
 
 ```csharp
 [Get("/users/list")]
@@ -251,7 +255,7 @@ var gitHubApi = RestService.For<IGitHubApi>("https://api.github.com",
     });
 ```
 
-### Unescape Querystring parameters
+#### Unescape Querystring parameters
 
 Use the `QueryUriFormat` attribute to specify if the query parameters should be url escaped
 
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -42,9 +42,11 @@ services
...
   * [Breaking changes in 6.x](#breaking-changes-in-6x)
 * [API Attributes](#api-attributes)
-* [Dynamic Querystring Parameters](#dynamic-querystring-parameters)
-* [Collections as Querystring parameters](#collections-as-querystring-parameters)
-* [Unescape Querystring parameters](#unescape-querystring-parameters)
+* [Querystrings](#querystrings)
+  * [Dynamic Querystring Parameters](#dynamic-querystring-parameters)
+  * [Collections as Querystring parameters](#collections-as-querystring-parameters)
+  * [Unescape Querystring parameters](#unescape-querystring-parameters)
+  * [Customize Querystring parameter formatting](#customize-querystring-parameter-formatting)
```
kamranayub, line 49: Sorry, this felt like it was getting long and other sections are organized under topical headings.

README.md
```diff
@@ -264,6 +268,92 @@ Query("Select+Id,Name+From+Account")
 >>> "/query?q=Select+Id,Name+From+Account"
 ```
 
+#### Customize Querystring parameter formatting
+
+Using `RefitSettings.UrlParameterKeyFormatter` and `RefitSettings.UrlParameterFormatter`, you can customize the way querystrings
```
kamranayub, line 273: This PR only introduces `UrlParameterKeyFormatter` so I did not document existing functionality or an example of `UrlParameterFormatter`

Refit.Tests/RequestBuilder.cs
```diff
@@ -234,19 +234,19 @@ public interface IRestMethodInfoTests
...
         [Post("/foo")]
         Task ImpliedComplexQueryType(ComplexQueryObject queryParams, [Body] Dictionary<int, string> theData1);
 
         [Get("/api/{id}")]
-        Task MultipleQueryAttributes(int id, [Query]string text = null, [Query]int? optionalId = null, [Query(CollectionFormat = CollectionFormat.Multi)]string[] filters = null);
+        Task MultipleQueryAttributes(int id, [Query] string text = null, [Query] int? optionalId = null, [Query(CollectionFormat = CollectionFormat.Multi)] string[] filters = null);
 
         [Get("/api/{id}")]
-        Task NullableValues(int id, string text = null, int? optionalId = null, [Query(CollectionFormat = CollectionFormat.Multi)]string[] filters = null);
+        Task NullableValues(int id, string text = null, int? optionalId = null, [Query(CollectionFormat = CollectionFormat.Multi)] string[] filters = null);
```
kamranayub, line 249: Sorry, editor formatting on save. 😅

Refit/RefitSettings.cs
```diff
@@ -30,13 +31,16 @@ public RefitSettings()
         /// <param name="contentSerializer">The <see cref="IHttpContentSerializer"/> instance to use</param>
         /// <param name="urlParameterFormatter">The <see cref="IUrlParameterFormatter"/> instance to use (defaults to <see cref="DefaultUrlParameterFormatter"/>)</param>
         /// <param name="formUrlEncodedParameterFormatter">The <see cref="IFormUrlEncodedParameterFormatter"/> instance to use (defaults to <see cref="DefaultFormUrlEncodedParameterFormatter"/>)</param>
+        /// /// <param name="urlParameterKeyFormatter">The <see cref="IUrlParameterKeyFormatter"/> instance to use to format query key names (defaults to <see cref="DefaultUrlParameterKeyFormatter"/>)</param>
         public RefitSettings(
             IHttpContentSerializer contentSerializer,
             IUrlParameterFormatter? urlParameterFormatter = null,
-            IFormUrlEncodedParameterFormatter? formUrlEncodedParameterFormatter = null)
+            IFormUrlEncodedParameterFormatter? formUrlEncodedParameterFormatter = null,
+            IUrlParameterKeyFormatter? urlParameterKeyFormatter = null)
```
kamranayub, line 39: Wasn't sure about this, feels weird to have it be out of place...

Refit/RequestBuilderImplementation.cs
```diff
@@ -445,7 +440,7 @@ void AddMultipartItem(MultipartFormDataContent multiPartContent, string fileName
                 var keyType = key.GetType();
                 var formattedKey = settings.UrlParameterFormatter.Format(key, keyType, keyType);
```
kamranayub, line 441: Actually, I didn't see this before 😮 Apparently dictionary keys are formatted using `UrlParameterFormatter.Format`.

This accepts any `object` type whereas my impl normalizes (and requires) `string`.

clairernovotny:
Thanks for this.

I like the idea, but as you say, we should normalize the dictionaries with the object, but we still need to allow the parameter to be an object as not all dictionary keys are strings.

For the documentation, can you please also add docs/sample for `UrlParameterFormatter` as it'll help people see the difference.

PR #1364, opened 2022-05-23
Fix condition of method return type in RequestBuilder to only allow Task<> and IObservable<>

barchkile:
Bug fix

Generic return type of any type is allowed in the creation of REST client

**What is the new behavior?**
Generic return type in method won't be allowed unless they are `Task<>` or `IObservable<>` as stated in docs and exception.

> Note that in Refit unlike in Retrofit, there is no option for a synchronous network request - all requests must be async, either via Task or via IObservable. There is also no option to create an async method via a Callback parameter unlike Retrofit, because we live in the async/await future.

**What might this PR break?**
If anyone created an interface with a wrong generic return value their generation will fail now


dnfadmin:
[CLA assistant check](https://cla.dotnetfoundation.org/reactiveui/refit?pullRequest=1364) <br/>Thank you for your submission, we really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla.dotnetfoundation.org/reactiveui/refit?pullRequest=1364) before we can accept your contribution.<br/><br/>:x: barchkile [sign now](https://cla.dotnetfoundation.org/reactiveui/refit?pullRequest=1364)<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](https://cla.dotnetfoundation.org/check/reactiveui/refit?pullRequest=1364) it.</sub>

glennawatson (approved):

PR #1375, opened 2022-06-13
Add ability to scan for types in assemblies

DenisBalan:
As per doc
https://docs.microsoft.com/en-us/dotnet/api/system.type.gettype

And
https://stackoverflow.com/questions/48279264/type-gettype-doesnt-find-dynamically-created-type

The assembly-qualified name of the type to get. See AssemblyQualifiedName. If the type is in the *currently executing assembly or in mscorlib.dll/System.Private.CoreLib.dll*, it is sufficient to supply the type name qualified by its namespace.

Small demo
```
    public interface myapi<Instance> : myapi
    {

    }

    [Headers("Authorization: Bearer")]
    public interface myapi
    {
    }
```

Where `Instance` is a generated type during runtime
with `AssemblyBuilder` and `ModuleBuilder`.

So its required to be in a new assembly.

Doing so, Refit will not found the type, and throw an exception.

This PR is to address and fix that.

There is NO HARM at all doing this change.

If needed, i can make also PoC of this issue, and how this PR is solving that.

Bug fix

described above

**What is the new behavior?**
no bug

**What might this PR break?**
nothing!


DenisBalan:
Please review it and merge..

glennawatson:
This one seems like it'd be a performance hit. 

Something that can be addressed when more work is done on converting more of the project over to using a source generator

Issue #1387, opened 2022-07-25
[BUG] Unable to build Refit with MSBuild >= 17.x


modplug:
**Describe the bug**
Unable to build Refit with MSBuild 17.x.

**Steps To Reproduce**
1. Install MSBuild >= 17
2. run dotnet build.
3. See error

`refit/Refit/targets/refit.targets(45,11): error MSB4086: A numeric comparison was attempted on "$([System.Version]::Parse($(_RefitMSBuildMinVersion)).CompareTo($([System.Version]::Parse($(MSBuildVersion)))))" that evaluates to "−1" instead of a number, in condition " '$([System.Version]::Parse($(_RefitMSBuildMinVersion)).CompareTo($([System.Version]::Parse($(MSBuildVersion)))))' > '0' ". [/refit/Refit.Tests/Refit.Tests.csproj]`

**Expected behavior**
I expect to be able to build the project without errors.

If I remove the following condition from the targets file it builds as expected

`<Target Name="_RefitMSBuildVersionCheck"
          Condition=" '$([System.Version]::Parse($(_RefitMSBuildMinVersion)).CompareTo($([System.Version]::Parse($(MSBuildVersion)))))' &gt; '0' "
     BeforeTargets="ResolveAssemblyReferences;Build;Rebuild" >
    <Error
        Text = "Projects using Refit cannot build using MSBuild '$(MSBuildVersion)'. MSBuild '$(_RefitMSBuildMinVersion)' or later is required." />
  </Target>`

**Environment**
.NET SDK (reflecting any global.json):
 Version:   6.0.400-preview.22330.6
 Commit:    da7c9ccceb

Runtime Environment:
 OS Name:     Mac OS X
 OS Version:  12.4
 OS Platform: Darwin
 RID:         osx.12-arm64
 Base Path:   /usr/local/share/dotnet/sdk/6.0.400-preview.22330.6/

global.json file:
  Not found

Host:
  Version:      6.0.7
  Architecture: arm64
  Commit:       0ec02c8c96

.NET SDKs installed:
  6.0.302 [/usr/local/share/dotnet/sdk]
  6.0.400-preview.22330.6 [/usr/local/share/dotnet/sdk]

.NET runtimes installed:
  Microsoft.AspNetCore.App 6.0.6 [/usr/local/share/dotnet/shared/Microsoft.AspNetCore.App]
[...]

PR #1415, opened 2022-09-13
Changed the condition of _RefitMSBuildVersionCheck

dannythunder:
[[BUG] Unable to build Refit with MSBuild >= 17.x #1387](https://github.com/reactiveui/refit/issues/1387)

Can't evaluate if build versions is OK

**What is the new behavior?**
Changed the logic how they evaluate

**What might this PR break?**
Nothing i suppose

- [ ] ~~Tests for the changes have been added (for bug fixes / features)~~
- [ ] ~~Docs have been added / updated (for bug fixes / features)~~

**Other information**:
Somehow the Condition thinks 1 is bigger than 2, hence the ! (not) infront of the condition. 
Guess I lack some understanding about .target and such.

### DIFF
--- a/Refit/targets/refit.targets
+++ b/Refit/targets/refit.targets
@@ -42,8 +42,8 @@
   </Target>
 
   <Target Name="_RefitMSBuildVersionCheck"
-          Condition=" '$([System.Version]::Parse($(_RefitMSBuildMinVersion)).CompareTo($([System.Version]::Parse($(MSBuildVersion)))))' &gt; '0' "
-     BeforeTargets="ResolveAssemblyReferences;Build;Rebuild" >
+          Condition="!($(MSBuildVersion) >= $(_RefitMSBuildMinVersion))"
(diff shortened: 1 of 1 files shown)


Refit/targets/refit.targets
```diff
@@ -42,8 +42,8 @@
   </Target>
 
   <Target Name="_RefitMSBuildVersionCheck"
-          Condition=" '$([System.Version]::Parse($(_RefitMSBuildMinVersion)).CompareTo($([System.Version]::Parse($(MSBuildVersion)))))' &gt; '0' "
-     BeforeTargets="ResolveAssemblyReferences;Build;Rebuild" >
+          Condition="!($(MSBuildVersion) >= $(_RefitMSBuildMinVersion))"
```
clairernovotny, line 45: I don't think this is doing what you think it is -- we need to be using comparisons that are aware of version numbers.

PR #1448, opened 2022-11-29
Use Error property when throwing exception

### LINKED ISSUE #1376: [BUG] Unable to throw a ValidationApiException exception
I have a Web API that returns a `ProblemDetails` with the `application/problem+json` content type:

image

Then, I have the following Refit interface:

```
public interface IPeopleServiceApi
{
    [Post("/people")]
    Task<ApiResponse<Person>> GetAsync(Person person);
}
```

If I try to call this API from Refit, when there is a validation problem, in the `Error` property of the `ApiResponse` I see a `ValidationApiException` object:

image

However, if I call `EnsureSuccessStatusCodeAsync`, the exception that is raised is `ApiException`, not `ValidationApiException`. Moreover, the `HasContent` property of the exception is _false_:

image

If, instead, I change the interface to directly return a `Person `instead of an `ApiResponse<Person>`, when I call the API, a `ValidationApiException` is thrown, with the `Content `correctly set.

**Steps To Reproduce**

1. Invoke an API that return a problem details with the `application/problem+json` content type
2. Define a method in the interface for Refit that returns an _ApiReponse<T>_ for this API
3. Invoke the API from the client
4. Call the `ApiResponse.EnsureSuccessStatusCodeAsync` method
5. An `ApiException` is thrown

**Expected behavior**
A `ValidationApiException` should be thrown

**Environment**
- OS:       Windows 11
- Device:   Desktop
- Version:   6.3.2

marcominerva:
Bug fix

See https://github.com/reactiveui/refit/issues/1376

**Other information**:
Fixes #1376

### DIFF
--- a/Refit.Tests/ResponseTests.cs
+++ b/Refit.Tests/ResponseTests.cs
@@ -116,6 +116,43 @@ namespace Refit.Tests
             Assert.Equal("title", actualException.Content.Title);
             Assert.Equal("type", actualException.Content.Type);
         }
+        
+        /// <summary>
+        /// Test to verify if EnsureSuccessStatusCodeAsync throws a ValidationApiException for a Bad Request in terms of RFC 7807
+        /// </summary>
+        [Fact]
+        public async Task When_BadRequest_EnsureSuccessStatusCodeAsync_ThrowsValidationException()
+        {
+            var expectedContent = new ProblemDetails
+            {
+                Detail = "detail",
+                Errors = { { "Field1", new string[] { "Problem1" } }, { "Field2", new string[] { "Problem2" } } },
+                Instance = "instance",
+                Status = 1,
+                Title = "title",
+                Type = "type"
+            };
+
+            var expectedResponse = new HttpResponseMessage(HttpStatusCode.BadRequest)
+            {
+                Content = new StringContent(JsonConvert.SerializeObject(expectedContent))
+            };
+
+            expectedResponse.Content.Headers.ContentType = new System.Net.Http.Headers.MediaTypeHeaderValue("application/problem+json");
+            mockHandler.Expect(HttpMethod.Get, "http://api/GetApiResponseTestObject")
+                .Respond(req => expectedResponse);
+
+            try
+            {
+                using var response = await fixture.GetApiResponseTestObject();
+                await response.EnsureSuccessStatusCodeAsync();
+            }
+            catch (Exception ex)
+            {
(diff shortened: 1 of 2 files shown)


304NotModified:
> ✅️ Tests for the changes have been added (for bug fixes / features)

I dont see any tests in this PR?

marcominerva:
Is adding a new test class to Refit.Tests enough? Thank you @304NotModified

304NotModified:
That's a bit hard to decide without any test code. 

But I think only an unit tests is needed, not a integration test / service test or something like that. (See https://martinfowler.com/articles/practical-test-pyramid.html)

marcominerva:
Thank you for the clarification @304NotModified, I have added a Unit test for this change.

Refit.Tests/ResponseTests.cs
```diff
@@ -116,6 +116,43 @@ public async Task ThrowsValidationException()
...
+            var expectedResponse = new HttpResponseMessage(HttpStatusCode.BadRequest)
+            {
+                Content = new StringContent(JsonConvert.SerializeObject(expectedContent))
+            };
+
+            expectedResponse.Content.Headers.ContentType = new System.Net.Http.Headers.MediaTypeHeaderValue("application/problem+json");
+            mockHandler.Expect(HttpMethod.Get, "http://api/GetApiResponseTestObject")
+                .Respond(req => expectedResponse);
+
+            try
```
304NotModified, line 145: I think you should use `Assert.ThrowsAsync...` - see other tests

Refit.Tests/ResponseTests.cs
```diff
@@ -116,6 +116,43 @@ public async Task ThrowsValidationException()
...
+            {
+                using var response = await fixture.GetApiResponseTestObject();
+                await response.EnsureSuccessStatusCodeAsync();
+            }
+            catch (Exception ex)
+            {
+                Assert.True(ex is ValidationApiException);
+                Assert.NotNull((ex as ValidationApiException)?.Content);
```
304NotModified, line 153: I would expect that we could do better here, right now it's only checking if the content is `not null`

304NotModified:
Thanks for the tests! Added some comments to make them a bit better :)

marcominerva:
Done :) Let me know if now is OK!

304NotModified (approved):
LGTM

nathanpovo:
This will also fix #1465

PR #1468, opened 2023-01-24
Do not require starting slash in refit attributes with relative path

zetroot:
Feature/enhancement

Adding attribute like `[Get("foo/bar")]` throws `ArgumentException`

**What is the new behavior?**
You can use attributes like `[Get("foo/bar")]` without throwing exceptions.

**What might this PR break?**
I have changed validation rules for relative URLs. All tests passes ok, but i could miss something

**Other information**:
There is not essential need to start relative path, provided in method attributes, with '/'.


zetroot:
Seems to me test fail due to failure of mock server used for integration testing

glennawatson:
Seems to be some performance issues with this one, converting simple string comparison with the regex. Closing

PR #1418, opened 2022-09-20
Adds named httpclient support

redbaty:
**What kind of change does this PR introduce?**
Makes it possible to set a custom name to the HTTP Client factory

**What is the current behavior?**
You can't use the same interface to point at multiple environments for example.

**What is the new behavior?**
You can use custom names to the same interface but with different environment/handlers.

**What might this PR break?**
In the original implementation `UniqueName.ForType<T>()` was used as the client's name, so if a custom name is used it might break somewhere, tho I've looked and not found any other references for this other than tests.

**Please check if the PR fulfills these requirements**
- [ ] Tests for the changes have been added (for bug fixes / features)
- [ ] Docs have been added / updated (for bug fixes / features)

**Other information**:

PR #1367, opened 2022-05-28
(Please) allow developers to inject the MethodInfo as a Property

james-s-tayler:
**What kind of change does this PR introduce?**
A feature to make the `MethodInfo` of the method on the Refit client interface invoked available to handlers via the `HttpRequestMessage.Options` / `HttpRequestMessage.Properties`.

This behavior is **opt-in** via the `RefitSettings`
```csharp
services.AddRefitClient<ISomeAPI>(provider => new RefitSettings
            {
                InjectMethodInfoAsProperty = true
            })
        .ConfigureHttpClient(c => c.BaseAddress = new Uri("https://api.example.com"));
```

**What is the current behavior?**
People either can't do what they want or are implementing workarounds by polluting their interfaces. 

https://github.com/reactiveui/refit/pull/1180
https://github.com/reactiveui/refit/pull/1219
https://github.com/reactiveui/refit/pull/1317

https://github.com/reactiveui/refit/issues/1150
https://github.com/reactiveui/refit/issues/1156

**What is the new behavior?**
New behavior is `RefitSettings.InjectMethodInfoAsProperty` defaults to `false` so Refit as a library remains (mostly) unopinionated about the contents of `HttpRequestMessage.Options`/`HttpRequestMessage.Properties`, however if set to `true` it will inject the `MethodInfo` into  `HttpRequestMessage.Options`/`HttpRequestMessage.Properties` under the key `Refit.MethodInfo`.

**What might this PR break?**
Nothing, the behavior is opt-in, so it has to explicitly be enabled giving the developer the chance to consider if there might be any implications for their codebase if the full `MethodInfo` were to be populated into and to put any mitigations necessary should they require it if they do wish to take advantage of the functionality. 

**Please check if the PR fulfills these requirements**
- [x] Tests for the changes have been added (for bug fixes / features)
- [x] Docs have been added / updated (for bug fixes / features)

**Other information**:
I know previously it has been stated that it would be acceptable if we store the method name as a string, but I for one just don't feel comfortable relying on something as critical as an integration point between two systems containing some "stringly typed" behavior when it **could** be strongly typed. Once someone introduces an overloaded method then all bets are off and it wouldn't surprise me to see that cause a potentially nasty production incident for someone somewhere, and/or future support issues.

I believe this implementation is a good half-way point between simplicity, safety, and usability and sincerely hope we can get this merged.

Issue #1467, opened 2023-01-16
Remove assembly full name from generated HttpClient name


yohny:
When Refit is used with a logging library like serilog that logs the context of a log message (typically a class where log originated) the generated Refit HttpClient name causes log to be hard to read as its very long.
An example of the generated name: `System.Net.Http.HttpClient.Refit.Implementation.Generated+SerilogRefiReproControllersIApi, SerilogRefiRepro, Version=1.0.0.0, Culture=neutral, PublicKeyToken=null`
The implementation is in https://github.com/reactiveui/refit/blob/246ee8d9989c29092fdaba6821d5b3098a5ccf9e/Refit/UniqueName.cs#L49 where its clear that asembly full name is appended.
I asked for reasoning of such naming in discussion https://github.com/reactiveui/refit/discussions/1464, but got no answer.
Related issue in serilog that resulted in this feature request https://github.com/serilog/serilog/issues/1793

**Describe the solution you'd like**
The generated class name would be simlified - no asembly full name would be present.

soarwing52:
I'd like to know if there's a way to stop the log with this logger name
I use a custom DelegatingHandler as LogginHandler, and wont require the logger of System.Net.Http.HttpClient.Refit.Implementation.*

I tried to filer the name but it just always log
`{ "time": "2023-09-05 17:52:32.4068", "level": "INFO", "logger": "System.Net.Http.HttpClient.Refit.Implementation.Generated+GtrPlatformExtendApisIConsultantApiClient, Gtr.Platform.ExtendApis, Version=1.0.0.0, Culture=neutral, PublicKeyToken=null.LogicalHandler", "message": "End processing HTTP request after 73.0084ms - 200" }
{ "time": "2023-09-05 17:52:32.4068", "level": "INFO", "logger": "System.Net.Http.HttpClient.Refit.Implementation.Generated+GtrPlatformExtendApisIConsultantApiClient, Gtr.Platform.ExtendApis, Version=1.0.0.0, Culture=neutral, PublicKeyToken=null.ClientHandler", "message": "Received HTTP response headers after 32.7993ms - 200" }`

cdumange:
I would add that some CMS solution do not allow to change field size. I have a problem with Kentico logs. Max size = 100. My generated type full name ends at... 125 character !

Would be nice to be able to limit the size.

Redplcs:
Yea will be better if we will had option to configure logging category name

glennawatson:
[...]

PR #1496, opened 2023-04-17
Revert #1418, #1367, introduce public version of RestMethodInfo

anaisbetts:
Reverts #1418 and #1367, we will add the functionality of #1367 back as part of a fixed up version of #1317, and we'll re-evaluate #1418. This PR also removes access to the internal type RestMethodInfo, replacing it with a read-only sanitized version. While initially this public type will be limited, we can evaluate adding new fields as-needed. 

## TODO:

- [x] Revert #1418 
- [x] Revert #1367
- [x] Reimplement #1317 without using internal types
- [x] add functionality from #1367 in new public types as needed 
- [ ] Fix the merge commit that I have inevitably borked
- [ ] Docs / tests / etc


anaisbetts:
@james-s-tayler Can you make sure that the change proposed here still works for your use-case?

james-s-tayler:
@anaisbetts I think it will.

cbrevik:
I was looking forward to the changes in #1418 being released, can I ask why it was reverted? Current name schemes coming from the `UniqueName`-implementation seemingly really bloats up logs / metrics.

anaisbetts:
@cbrevik I reverted it because the implementation was too Java'y and it was extremely unclear from the PR what problem this actually solves or why it should exist - can you explain further?

cbrevik:
The problem it solves relates to how other logging and metric libraries logs the name of the HTTP-client.

See for example #1467 in this repository and the issue it links back to from the Serilog-repository: https://github.com/serilog/serilog/issues/1793#issue-1521198225

Serilog pulls the `FullName` from the HttpClient which normally includes namespace + type, as per the comment: https://github.com/serilog/serilog/issues/1793#issuecomment-1372632571

Logging of "normal" HTTP-clients will be in this format:
`2023-01-05 18:45:53.898 +01:00 [INF] System.Net.Http.HttpClient.MyController.LogicalHandler Start processing HTTP request POST http://example.com/stuff`

Whilst logging of Refit-clients will be like this:
`2023-01-05 18:45:53.861 +01:00 [INF] System.Net.Http.HttpClient.Refit.Implementation.Generated+MyProjectIApi, MyProject, Version=1.0.0.0, Culture=neutral, PublicKeyToken=null.LogicalHandler Start processing HTTP request POST http://example.com/stuff` 

Which including the prefixed stuff includes Version, PublicKeyToken etc.

Mostly then because of the implementation of https://github.com/reactiveui/refit/blob/main/Refit/UniqueName.cs

I've seen the same issue with for example using Prometheus, a sample metric from there:
`httpclient_response_duration_seconds_sum{method="GET",host="api.example.com",client="Refit.Implementation.Generated+MyFancyApiInfrastructureInsightClientIInsightHttpClient, My.Fancy.Api, Version=1.0.0.0, Culture=neutral, PublicKeyToken=null",code="200"} 0.2517555`

So the issue is that it really bloats the logs and metrics up, and makes them a bit unreadable. Especially if you have (like we have) several different HTTP-clients with Refit. So I'd like for the generated name to be either shorter by default (and I see from your explanation why it shouldn't), or at least make it overridable if not (like in #1418 or a less Java-y solution)

anaisbetts:
@cbrevik Got it, thanks for the explanation. I think we can figure something out then

anaisbetts:
@cbrevik Can you confirm that https://github.com/reactiveui/refit/pull/1496/commits/3248979b46a3d800d40cdb7c1bfdfefac32431f8 works for you?

Refit/HttpRequestMessageProperties.cs
```diff
@@ -12,24 +10,9 @@ public static class HttpRequestMessageOptions
...
-        /// <summary>
-        /// A typed key to access the <see cref="System.Type"/> of the top-level interface where the method was called from
-        /// on the <see cref="System.Net.Http.HttpRequestMessage.Options"/>.
-        /// </summary>
-        public static System.Net.Http.HttpRequestOptionsKey<System.Type> InterfaceTypeKey { get; } = new(InterfaceType);
-#endif
-
         /// <summary>
-        /// Returns the <see cref="Refit.RestMethodInfo"/> of the method that was called
+        /// Returns the <see cref="Refit.RestMethodInfoInternal"/> of the top-level interface
```
anaisbetts, line 14: Typo

cbrevik:
> @cbrevik Can you confirm that [3248979](https://github.com/reactiveui/refit/commit/3248979b46a3d800d40cdb7c1bfdfefac32431f8) works for you?

It looks like it'll be able to do what I want yeah! I can test it too if a preview nuget-package is released too? Unless there is some other easy way for me to test out that commit specifically?

Issue #1598, opened 2023-11-14
Generated code should use `Array.Empty` or constant field where possible


TimothyMakkison:
## `Array.Empty`
Each method call with no parameters allocates an empty params and type array. To prevent unneeded allocations `Array.Empty` should be used instead.

```C#
public async global::System.Threading.Tasks.Task<global::System.Net.Http.HttpResponseMessage> GetIndex()
{
    var ______arguments = new object[] {  }; // unneeded allocation
    var ______func = requestBuilder.BuildRestResultFuncForMethod("GetIndex", new global::System.Type[] {  } );
    ...
```

### With Array.Empty
```C#
public async global::System.Threading.Tasks.Task<global::System.Net.Http.HttpResponseMessage> GetIndex()
{
    var ______arguments = global::System.Array.Empty<object>();
    var ______func = requestBuilder.BuildRestResultFuncForMethod(""GetIndex"", global::System.Array.Empty<global::System.Type>() );
```

## `typeParameters`
The `parameterTypes` array remains the same between calls, this could be moved out into a `private static` or `private readonly` field. The generator will have to falback to the current method if one of the parameters is a generic type scoped to the method.

### Before
```C#
public async global::System.Threading.Tasks.Task<global::Refit.Tests.UserSearchResult> FindUsers(string @q)
{
    var ______arguments = new object[] { @q };
    var ______func = requestBuilder.BuildRestResultFuncForMethod(""FindUsers"", new global::System.Type[] { typeof(string) } );
    ...
```

### After
```C#
private static global::System.Type[] typeArray1 = new global::System.Type[] { typeof(string) };

public async global::System.Threading.Tasks.Task<global::Refit.Tests.UserSearchResult> FindUsers(string @q)
{
    var ______arguments = new object[] { @q };
    var ______func = requestBuilder.BuildRestResultFuncForMethod(""FindUsers"", typeArray1 );
    ...
```
## Generic `typeParameters`

```C#
private static global::System.Type[]? typeArray1 = null;

public async global::System.Threading.Tasks.Task<global::Refit.Tests.UserSearchResult> FindUsers<T>(string @q, [Body]T body)
{
    if (typeArray1 == null)
    {
        typeArray1 = new global::System.Type[] { typeof(string), typeof(T) };
    }
    var ______arguments = new object[] { @q };
    var ______func = requestBuilder.BuildRestResultFuncForMethod(""FindUsers"", typeArray1 );
    ...
```

PR #1599, opened 2023-11-14
feat: generate code that uses `Array.Empty` where possible

TimothyMakkison:
Use `Array.Empty` over allocating an empty array where possible. Done for `______arguments` and `parameterTypes`..

See #1598

### DIFF
--- a/InterfaceStubGenerator.Shared/InterfaceStubGenerator.cs
+++ b/InterfaceStubGenerator.Shared/InterfaceStubGenerator.cs
@@ -413,11 +413,19 @@ namespace Refit.Implementation
                 genericList.Add($"typeof({typeParam.ToDisplayString(SymbolDisplayFormat.FullyQualifiedFormat)})");
             }
 
-            var genericString = genericList.Count > 0 ? $", new global::System.Type[] {{ {string.Join(", ", genericList)} }}" : string.Empty;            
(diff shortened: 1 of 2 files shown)


InterfaceStubGenerator.Shared/InterfaceStubGenerator.cs
```diff
@@ -95,7 +95,7 @@ public void Execute(GeneratorExecutionContext context)
             var methodSymbols = new List<IMethodSymbol>();
             foreach (var group in candidateMethods.GroupBy(m => m.SyntaxTree))
             {
-                var model = compilation.GetSemanticModel(group.Key);
```
anaisbetts, line 98: It's not a big deal with this PR but try not to make a bunch of whitespace-only changes, I know it can be a pain in the ass

anaisbetts (approved):
This is great, thanks for the solid PR. @clairernovotny, can you think of anything this might break?

clairernovotny (approved):

clairernovotny:
Can't think of any issues this would cause. Thanks!

TimothyMakkison:
Sorry about that, should be fixed now 🤞 

Any chance you'd be open to running a formatter on this repository? A lot of lines appear to have random 4-8 sections of whitespace.

PR #1606, opened 2023-11-16
feat: use `private static` fields to store constant `typeParameters` where possible

TimothyMakkison:
Modify source generator to use `private static readonly` fields to store the `typeParameter` arrays, saving the allocations for each request.

- Only generates when there are actual arguments, otherwise `Array.Empty` is used - #1599
- `private static` fields are generated beside its respective method, this could be moved to the start of the class if requested.
- I've used a `HashSet` of each member in scope to accidental name collisions. Fields will have names like `_typeParameters3` incrementing for each field generated. 
- ~Uses nullable coalescing assignment if one of the method parameters uses type arguments. This is done as we may not be able to access it from outside the method scope making the array initialization invalid.~
- ~It should be safe to initialize and set the static arrays in multithreaded contexts as setting references are thread safe in C#.~

Edit: `typeof(T)` will be different for each usage of T so it isn't possible to have a reusable `_genericType` array. This can also cause issues for `_typeParameter`. `GenerateTypeParameterExpression` ensures that each parameter doesn't contain a generic type parameter, falling back to the current solution if it detects one.

See #1598

### DIFF
--- a/InterfaceStubGenerator.Shared/InterfaceStubGenerator.cs
+++ b/InterfaceStubGenerator.Shared/InterfaceStubGenerator.cs
@@ -23,6 +23,8 @@ namespace Refit.Generator
     public class InterfaceStubGenerator : ISourceGenerator
 #endif
     {
+        private const string TypeParameterVariableName = "_typeParameters";
+
 #pragma warning disable RS2008 // Enable analyzer release tracking
         static readonly DiagnosticDescriptor InvalidRefitMember =
             new(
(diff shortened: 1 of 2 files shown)


anaisbetts:
Seems legit, @clairernovotny?

anaisbetts:
@ChrisPulman Any objections to this PR?

ChrisPulman:
> @ChrisPulman Any objections to this PR?

I am currently in India, but I will check it this evening and merge it if it's fitting to our requirements, thank you.

ChrisPulman:
All looks good, thank you Tim

ChrisPulman (approved):

Issue #1602, opened 2023-11-15
Suggestion: run code cleanup/formatter


TimothyMakkison:
The use of a formatter/code cleanup would go a long way in improving the readability of the code, there's a lot of unused usings and strange blocks of whitespace.

Perhaps adding something like csharpier or similar to the CI pipeline would help.

PR #1596, opened 2023-11-13
Minor optimizations and cleanup code. 

TimothyMakkison:
- Remove unneeded usings
- Remove some null checks (it looks like nullable annotations were hastily added so I kept most of the redundant checks in place)
- Fixed typos
- Inverted `ifs` and used early returns/continutes to reduce nesting
- Removed unneeded `parameterTypes.ToArray()` call
- Replaced `dict.ContainsKey` -> `dict[key]` with `TryGetValue` where appropiate

PR #1617, opened 2023-11-20
chore: run csharpier

TimothyMakkison:
Ran csharpier `0.25.0`, haven't made any other changes.

See #1596 & #1602


anaisbetts:
I think we probably also want to set this up as a build tool? Running it Just Once doesn't really make much sense

TimothyMakkison:
Good idea, I'm not the best at CI but I know how to run csharpier via husky if that's okay

anaisbetts:
Paging @ChrisPulman @glennawatson who I bet have a great idea on how to do this

glennawatson:
We can enforce the stylecop rules which probably be more useful long term. What we do on other rxui projects. Basically means it won't get out of control again. 

Stylecop has the advantage of allowed exception etc.

TimothyMakkison:
#1622 should automatically run csharpier on all commits.

glennawatson:
I'd prefer we don't run the csharpier ever commit and we use stylecop analysers to maintain the styling. It has more flexibility then overriding deliberate style choices.

TimothyMakkison:
Fair enough, as long as the solution removes the whitespace I'm happy 😊

glennawatson:
Let's commit this one as a one time commit, I just don't want this running EVERY commit.

Sorry for the confusion.

glennawatson:
Apologies, can you rerun the application on that failing file? Thanks.

glennawatson (approved):

Issue #1671, opened 2024-02-22
[NiceToHave]: Interface `IApiResponse<out T>` misses `[MemberNotNullWhen(true, nameof(Content))]` attribute


sguryev:
### Describe the bug 🐞

Analyzer shows the warming for the `IApiResponse<out T>.Content` property after checking `IsSuccessStatusCode`.

### Step to reproduce

1. Return `IApiResponse<T>` from the method.
2. Call the method
3. Check the response `IsSuccessStatusCode == true` 
4. See the error\warning

### Reproduction repository

https://github.com/reactiveui/refit

### Expected behavior

Analyzer should consider `Content` as not null after checking `IsSuccessStatusCode` for `true`

### Screenshots 🖼️

![image](https://github.com/reactiveui/refit/assets/6978458/fc075887-55fa-443a-a9fd-5f7b6fd8c88e)

![image](https://github.com/reactiveui/refit/assets/6978458/77ba2a33-e5ac-40da-8f58-1f8a0248d446)
![image](https://github.com/reactiveui/refit/assets/6978458/f8427f4f-4047-4430-81a4-85a2323a1ba6)
![image](https://github.com/reactiveui/refit/assets/6978458/cd127e86-0c6b-4e4c-aa28-5099c5f4ae37)

### Operating system

Windows

### Version

11

### Refit Version

7.0.0

PR #1672, opened 2024-02-22
Issue 1671: Add MemberNotNullWhen attribute for Content property in IApiResponse<T>

sguryev:
Potential fix for https://github.com/reactiveui/refit/issues/1671

An alternative way:
`new bool IsSuccessStatusCode { get; }` property in the `IApiResponse<T>` interface with all 3 attributes above

### DIFF
--- a/Refit/ApiResponse.cs
+++ b/Refit/ApiResponse.cs
@@ -29,7 +29,7 @@ namespace Refit
     /// Implementation of <see cref="IApiResponse{T}"/> that provides additional functionalities.
     /// </summary>
     /// <typeparam name="T"></typeparam>
-    public sealed class ApiResponse<T> : IApiResponse<T>
+    public sealed class ApiResponse<T> : IApiResponse<T>, IApiResponse
     {
         readonly HttpResponseMessage response;
         bool disposed;
--- a/Refit/ApiResponse.cs
+++ b/Refit/ApiResponse.cs
@@ -60,6 +60,8 @@ namespace Refit
         /// </summary>
         public T? Content { get; }
 
+        object? IApiResponse.Content => Content;
+
         /// <summary>
         /// Refit settings used to send the request.
         /// </summary>
@@ -158,13 +160,19 @@ namespace Refit
     }
 
     /// <inheritdoc/>
-    public interface IApiResponse<out T> : IApiResponseBase
+    public interface IApiResponse<out T> : IApiResponse
     {
         /// <summary>
         /// Deserialized request content as <typeparamref name="T"/>.
         /// </summary>
-        T? Content { get; }
+        new T? Content { get; }
+    }
 
+    /// <summary>
+    /// Base interface used to represent an API response.
+    /// </summary>
+    public interface IApiResponse : IDisposable
+    {
         /// <summary>
         /// Indicates whether the request was successful.
         /// </summary>
@@ -174,25 +182,12 @@ namespace Refit
         [MemberNotNullWhen(true, nameof(Content))]
 #endif
         bool IsSuccessStatusCode { get; }
-    }
 
-    public interface IApiResponse : IApiResponseBase
-    {
         /// <summary>
-        /// Indicates whether the request was successful.
+        /// Deserialized request content as an object.
         /// </summary>
-#if NET6_0_OR_GREATER
-        [MemberNotNullWhen(true, nameof(ContentHeaders))]
-        [MemberNotNullWhen(false, nameof(Error))]
-#endif
-        bool IsSuccessStatusCode { get; }
-    }
+        object? Content { get; }
 
-    /// <summary>
-    /// Base interface used to represent an API response.
-    /// </summary>
-    public interface IApiResponseBase : IDisposable
-    {
         /// <summary>
         /// HTTP response headers.
         /// </summary>
--- a/Refit/ApiResponse.cs
+++ b/Refit/ApiResponse.cs
@@ -131,18 +131,40 @@ namespace Refit
     }
 
     /// <inheritdoc/>
-    public interface IApiResponse<out T> : IApiResponse
+    public interface IApiResponse<out T> : IApiResponseBase
     {
         /// <summary>
         /// Deserialized request content as <typeparamref name="T"/>.
         /// </summary>
         T? Content { get; }
+
+        /// <summary>
+        /// Indicates whether the request was successful.
+        /// </summary>
+#if NET6_0_OR_GREATER
+        [MemberNotNullWhen(true, nameof(ContentHeaders))]
+        [MemberNotNullWhen(false, nameof(Error))]
+        [MemberNotNullWhen(true, nameof(Content))]
+#endif
+        bool IsSuccessStatusCode { get; }
+    }
+
+    public interface IApiResponse : IApiResponseBase
+    {
+        /// <summary>
+        /// Indicates whether the request was successful.
+        /// </summary>
+#if NET6_0_OR_GREATER
+        [MemberNotNullWhen(true, nameof(ContentHeaders))]
+        [MemberNotNullWhen(false, nameof(Error))]
+#endif
+        bool IsSuccessStatusCode { get; }
     }
 
     /// <summary>
     /// Base interface used to represent an API response.
     /// </summary>
-    public interface IApiResponse : IDisposable
+    public interface IApiResponseBase : IDisposable
     {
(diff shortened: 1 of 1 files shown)


HavenDV:
This is similar to Breaking Change because `IApiResponse<out T>` is no longer `IApiResponse` and this may affect those using such a cast.

I suggest doing this via virtual, like here - https://stackoverflow.com/questions/72232151/membernotnullwhen-or-other-code-analysis-attributes-for-inherited-members

sguryev:
@ChrisPulman oopsie, I'm sorry. I have forgotten to add IApiResponse interface to the ApiResponse<T> class.

ChrisPulman:
I have a PR that needs to be merged to get this to build successfully, once this is in it will build and should be acceptable

ChrisPulman (approved):

sguryev:
@ChrisPulman did you have time to check how analyzer react on `new T? Content` property? 
I have tried to set target `net6.0,net7.0` for Refit itself in order to check how does it work for `IApiResponse<T>`
It works for `IApiResponse` but it doesn't work for `IApiResponse<T>`
My original idea was to modify the Refit the way I don't need to suppress analyzer for instances I have checked `IsSuccessStatusCode`. Now I have to check the flag and content together (and keep in mind that Content can't be null for true) ` if (!saleSessionResponse.IsSuccessStatusCode || saleSessionResponse.Content == null)`
image

ChrisPulman:
I have a PR that needs to go in to revert the targets back to Net6.0, Net8.0, the current code still has the NET6.0_Or_Greater flags

ChrisPulman:
I have now merged my PR, I will add the tests you have shared pictorially, and we can go from there. 

Note for next time, adding tests to show how your change will work helps us to understand your use case, it also means that once we accept that we can ensure nothing else breaks this functionality without us knowing.

ChrisPulman:
I ran the tests and the attributes don't seem to do anything, something is wrong here.

I understand the Attributes are supposed to return a non-null value when the condition is met.
Therefore, the test below should have been success, however it fails.

```
class MemberNotNullWhenTest(bool isSuccessStatusCode)
{
    public object? Content { get; set; }
    public object? ContentHeaders { get; set; }
    public object? Error { get; set; }


    [MemberNotNullWhen(true, nameof(ContentHeaders))]
    [MemberNotNullWhen(false, nameof(Error))]
    public bool IsSuccessStatusCode { get; } = isSuccessStatusCode;
}

[Fact]
public void MemberNotNullWhenTest_IsSuccessStatusCodeReturnsNonNullValues()
{
var memberNotNullWhenTest = new MemberNotNullWhenTest(true);
    if (memberNotNullWhenTest.IsSuccessStatusCode)
    {
        Assert.NotNull(memberNotNullWhenTest.Content);
        Assert.NotNull(memberNotNullWhenTest.ContentHeaders);
        Assert.Null(memberNotNullWhenTest.Error);
    }
    else
    {
        Assert.False(true);
    }
}
```

What are your thoughts on this?
I should have added the tests before merging this, but it made sense at the time.
In my opinion, the Attributes are failing to do their job.

sguryev:
@ChrisPulman let me look closer later today

ChrisPulman:
@sguryev thank you, the attributes really seem to be doing nothing, I have tested them in a plain library ensured that it was Nullable but nothing happens. In theory they are great but only if they do the task required

sguryev:
@ChrisPulman Unfortunately it's not something you can easily test with unit testing. It's about null-state static analysis. 

You can find explanation here: https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/attributes/nullable-analysis#helper-methods-membernotnull-and-membernotnullwhen

I have also prepared the sample you can try. Please notice the `WarningsAsErrors` attribute in the csproj file
csproj:
```
<Project Sdk="Microsoft.NET.Sdk">

    <PropertyGroup>
        <OutputType>Exe</OutputType>
        <TargetFramework>net8.0</TargetFramework>
        <ImplicitUsings>enable</ImplicitUsings>
        <Nullable>enable</Nullable>

        <WarningsAsErrors>CS8600</WarningsAsErrors>
        <!--<WarningsAsErrors>CS8600;CS8601;CS8602;CS8603;CS8604;CS8613;CS8614;CS8619;CS8620;CS8622;CS8625;CS8629;CS8633,CS8767,CS8524,CS8509</WarningsAsErrors>-->
    </PropertyGroup>

</Project>
```

Program:
```
// See https://aka.ms/new-console-template for more information

using System.Diagnostics.CodeAnalysis;

string str = "sample";

var nonAttrModel = new NonAttrModel();
if (nonAttrModel.HasContent)
    str = nonAttrModel.Content;

var attrModel = new AttrModel();
if (attrModel.HasContent)
    str = attrModel.Content;

Console.WriteLine($"Hello, World! {str}");

public class NonAttrModel
{
    public bool HasContent { get; set; }
    public string? Content { get; set; }
}

public class AttrModel
{
    public bool HasContent { get; set; }
    public string? Content { get; set; }
}
```
You should see something like this:
image

There is a bunch of warnings which can be threated as errors in order to prevent assign nullable ref types to non-null ref types to avoid null-ref exceptions in runtime.

According to the current code in Refit I think than `new` modifier breaks the analysis. So attribute works for the `IApiResponse.Content` and not for `IApiResponse<T>.Content` because of the `new` modified.
image

Please let me know if you have any questions

ChrisPulman:
I also saw the difference with the new
image

In addition to this I have the following interfaces

The IApiResponse<T> interface
    ```
/// <inheritdoc/>
    public interface IApiResponse<out T> : IApiResponseBase
    {
        /// <summary>
        /// Deserialized request content as <typeparamref name="T"/>.
        /// </summary>
        T? Content { get; }

        /// <summary>
        /// Indicates whether the request was successful.
        /// </summary>
#if NET6_0_OR_GREATER
        [MemberNotNullWhen(true, nameof(Content), nameof(ContentHeaders))]
        [MemberNotNullWhen(false, nameof(Error))]
#endif
        bool IsSuccessStatusCode { get; }

        /// <summary>
        /// HTTP response content headers as defined in RFC 2616.
        /// </summary>
        HttpContentHeaders? ContentHeaders { get; }

        /// <summary>
        /// The <see cref="ApiException"/> object in case of unsuccessful response.
        /// </summary>
        ApiException? Error { get; }
    }
```
The IApiResponse interface
```
    /// <summary>
    /// IApiResponse.
    /// </summary>
    /// <seealso cref="Refit.IApiResponseBase" />
    public interface IApiResponse : IApiResponseBase
    {
        /// <summary>
        /// Indicates whether the request was successful.
        /// </summary>
#if NET6_0_OR_GREATER
        [MemberNotNullWhen(true, nameof(ContentHeaders))]
        [MemberNotNullWhen(false, nameof(Error))]
#endif
        bool IsSuccessStatusCode { get; }

        /// <summary>
        /// HTTP response content headers as defined in RFC 2616.
        /// </summary>
        HttpContentHeaders? ContentHeaders { get; }

        /// <summary>
        /// The <see cref="ApiException"/> object in case of unsuccessful response.
        /// </summary>
        ApiException? Error { get; }
    }
```
The Base Interface
```
/// <summary>
/// Base interface used to represent an API response.
/// </summary>
public interface IApiResponseBase : IDisposable
{
    /// <summary>
    /// HTTP response headers.
    /// </summary>
    HttpResponseHeaders Headers { get; }

    /// <summary>
    /// HTTP response status code.
    /// </summary>

    /// <summary>
    /// The reason phrase which typically is sent by the server together with the status code.
    /// </summary>
    string? ReasonPhrase { get; }

    /// <summary>
    /// The HTTP Request message which led to this response.
    /// </summary>
    HttpRequestMessage? RequestMessage { get; }

    /// <summary>
    /// HTTP Message version.
    /// </summary>
    Version Version { get; }
}
```

I will be travelling for the majority of Tuesday and Wednesday but will pick this up again when I can.

sguryev:
My idea was to add `IApiResponseBase` and extract `IsSuccessStatusCode` property from the base interface in order to be able to add attributes over `IsSuccessStatusCode` in each implementation.

AFAIU these attributes don't like base or hidden properties inside. 
We don't need interface-specific `IsSuccessStatusCode` implementation since it has identical signature.
image

ChrisPulman:
Really sorry, would it be possible for you to create a new PR with the updated version. I will try to come up with a test that we can prove this works and ensure that future changes maintain the functionality

Issue #1384, opened 2022-07-19
[BUG] When response deserialization fails, Content is null in ApiException


KarolStosik:
**Describe the bug**

When Refit can't deserialize response from api, ApiException is thrown. Unfortunately, Content property is always null, but it should be populated with the response content, that can't be deserialized.

**Steps To Reproduce**

Reproduction with simple XUnit test (api can be arbitrary, here I've used first public api I've found):

```
using System.Threading.Tasks;
using Refit;
using Xunit;

namespace RefitReproduction
{
    public class RefitNullContentTest
    {
        [Fact]
        public async Task ShouldHaveContentInException()
        {
            var api = RestService.For<IApi>("https://api.spacexdata.com");
            var exception = await Assert.ThrowsAsync<ApiException>(() => api.GetLaunches());
            Assert.NotNull(exception.Content); // fails!
        }
    }

    public interface IApi
    {
        [Get("/v3/launches")]
        public Task<double> GetLaunches(); // any type not matching api response
    }
}

```

**Expected behavior**

When deserialization fails, Content property should be populated with response content.

**Environment**

- OS:       Windows
- Device:   PC
- .NET version: 6.0
- Refit version:   6.3.2

KarolStosik:
I've found the source of bug:

ApiException.cs, line 132:

```
            try
            {
                exception.ContentHeaders = response.Content.Headers;
                var content = await response.Content.ReadAsStringAsync().ConfigureAwait(false);
                exception.Content = content;

                if (response.Content.Headers?.ContentType?.MediaType?.Equals("application/problem+json") ?? false)
                {
                    exception = ValidationApiException.Create(exception);                    
                }

                response.Content.Dispose();
            }
            catch
            {
                // NB: We're already handling an exception at this point, 
                // so we want to make sure we don't throw another one 
                // that hides the real error.
            }
```

`response.Content.ReadAsStringAsync()` throws `System.InvalidOperationException: 'The stream was already consumed. It cannot be read again.'`

Anyone knows how to fix it?

XVCoder:
So, any solutions now?

thomasb-gr:
Hi, were you able to figure out a solution to this problem?

KSwift87:
[...]

PR #1705, opened 2024-05-31
Load content to buffer before attempting deserialization 

LichP:

Fixes issue #1384 by ensuring response content is loaded to a buffer before attempting serialization.

When an exception is thrown during deserialization, `ApiException` attempts to capture the request content to the `Content` property to make it available when the exception is being handled (e.g. so apps can log the content to aid debugging). However, the underlying stream in the `HttpContent` object has already been consumed by the deserialization and cannot be re-read.

**What is the new behavior?**

By loading the content to a buffer before attempting to deserialize we ensure the stream can be re-read by `ApiException`.

**What might this PR break?**

I've moved the `LoadIntoBufferAsync` call to just before the call to the (de)serializer to ensure we only do this on this exact code path, rather than as soon as the `content` variable is set as I originally suggested in [#1384](https://github.com/reactiveui/refit/issues/1384#issuecomment-2140509439). This should avoid any problems that could arise from API calls that expect to handle the `HttpContent` or `Stream` directly (calling `LoadIntoBufferAsync` is probably not desirable for these scenarios where we might expect endpoints to return long-lived connections or very large payloads).

**Other information**:


ChrisPulman:
Unfortunately the tests are falling with this addition, please could you see how this may be resolved.

LichP:
The failing test was for the non-seekable stream case: to address this I'm now obtaining the underlying stream with `ReadAsStreamAsync` and only attempting the call to `LoadIntoBufferAsync` if the stream is seekable. I've also adjusted the `ThrowOnGetLengthMemoryStream` mock to appropriately implement the `CanSeek` property.

ChrisPulman (approved):

PR #1719, opened 2024-06-14
feat: use `ValueStringBuilder` adding the query parameters

TimothyMakkison:
Uses [`ValueStringBuilder`](https://github.dev/dotnet/runtime/blob/main/src/libraries/Common/src/System/Text/ValueStringBuilder.cs) to generate query parameters for the uri and prepend a question mark.

Update: turns out `UriBuilder` has been changed between .NET Framework and Core. Core checks for leading `?` symbols whereas Framework always adds one. I've added a preprocessor directive to resovle this issue.

### Original
| Method                     | Mean     | Error    | StdDev   | Gen0   | Gen1   | Allocated |
|--------------------------- |---------:|---------:|---------:|-------:|-------:|----------:|
| TaskTQueryLong_Async       | 12.97 us | 0.258 us | 0.665 us | 1.5717 | 0.0458 |  14.45 KB |
| TaskTCollectionQuery_Async | 18.64 us | 0.372 us | 0.848 us | 1.8921 | 0.0610 |   17.4 KB |

### Changes

| Method                     | Mean     | Error    | StdDev   | Median   | Gen0   | Gen1   | Allocated |
|--------------------------- |---------:|---------:|---------:|---------:|-------:|-------:|----------:|
| TaskTQueryLong_Async       | 12.81 us | 0.278 us | 0.712 us | 12.60 us | 1.5259 |      - |  14.15 KB |
| TaskTCollectionQuery_Async | 18.21 us | 0.325 us | 0.348 us | 18.24 us | 1.8311 | 0.0305 |  16.85 KB |

PR #1730, opened 2024-06-22
feat: WIP optimize url creation

TimothyMakkison:
WIP; Uses the first call to determine the substitutions for the URL.

- This might be a "breaking change" as the url conversion is done later on and in bulk, in addition to calling `UrlParameterFormatter.Format` on the same value if it appears several times in the URL.
    - If this is an issue I can use a `ValueListBuilder` to preserve the order of operations and to cache values,
- `ParameterFragment` is my messy attempt to create a tagged union.
  - Instead of having a `string Value` property it could be an index and count representing the reevant area in the original url.
- Uses `ValueStringBuilder` I'll need to rebase this after #1719 is merged.
- This would ideally be done in the source generator.
- Could do with some benchmarks, especially for startup times.

### DIFF
--- a/Refit.Tests/API/ApiApprovalTests.Refit.DotNet6_0.verified.txt
+++ b/Refit.Tests/API/ApiApprovalTests.Refit.DotNet6_0.verified.txt
@@ -393,6 +393,11 @@ namespace Refit
         public System.Net.Http.HttpContent ToHttpContent<T>(T item) { }
         public static System.Text.Json.JsonSerializerOptions GetDefaultJsonSerializerOptions() { }
     }
+    public class UnreachableException : System.Exception
+    {
+        public UnreachableException() { }
+        public UnreachableException(string message) { }
+    }
     [System.Serializable]
     public class ValidationApiException : Refit.ApiException
     {
--- a/Refit.Tests/API/ApiApprovalTests.Refit.DotNet8_0.verified.txt
+++ b/Refit.Tests/API/ApiApprovalTests.Refit.DotNet8_0.verified.txt
@@ -393,6 +393,11 @@ namespace Refit
         public System.Net.Http.HttpContent ToHttpContent<T>(T item) { }
         public static System.Text.Json.JsonSerializerOptions GetDefaultJsonSerializerOptions() { }
     }
+    public class UnreachableException : System.Exception
+    {
+        public UnreachableException() { }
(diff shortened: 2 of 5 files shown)


TimothyMakkison:
Rebased and done some modifications. Added `Refit.UnreachableException` for cases that should not be able to happen.

Not sure what I'm missing but bizarrely the benchmarks don't show a major change in memory usage. I'll review this a different day but I might abandon this and reuse the logic for the source generator.

### Before
| Method                   | Mean      | Error     | StdDev    | Gen0   | Gen1   | Allocated |
|------------------------- |----------:|----------:|----------:|-------:|-------:|----------:|
| ConstantRouteAsync       |  2.045 us | 0.0395 us | 0.0638 us | 0.6828 | 0.0114 |   6.28 KB |
| DynamicRouteAsync        |  2.601 us | 0.0518 us | 0.0532 us | 0.7172 | 0.0038 |    6.6 KB |
| ComplexDynamicRouteAsync |  3.799 us | 0.0651 us | 0.0891 us | 0.7858 |      - |   7.24 KB |
| ObjectRequestAsync       |  4.560 us | 0.0886 us | 0.1021 us | 0.8698 |      - |   8.05 KB |
| ComplexRequestAsync      | 12.951 us | 0.2575 us | 0.4643 us | 1.5259 |      - |  14.19 KB |

### Changes
| Method                   | Mean      | Error     | StdDev    | Median    | Gen0   | Gen1   | Allocated |
|------------------------- |----------:|----------:|----------:|----------:|-------:|-------:|----------:|
| ConstantRouteAsync       |  3.037 us | 0.0602 us | 0.1174 us |  3.028 us | 0.6828 |      - |    6.3 KB |
| DynamicRouteAsync        |  3.195 us | 0.2061 us | 0.6076 us |  3.410 us | 0.7172 | 0.0076 |   6.63 KB |
| ComplexDynamicRouteAsync |  3.054 us | 0.0496 us | 0.0464 us |  3.033 us | 0.7858 | 0.0076 |   7.23 KB |
| ObjectRequestAsync       |  4.194 us | 0.0613 us | 0.0512 us |  4.180 us | 0.8698 |      - |   8.04 KB |
| ComplexRequestAsync      | 11.623 us | 0.1973 us | 0.1749 us | 11.590 us | 1.5259 | 0.0153 |  14.16 KB |

PR #1735, opened 2024-06-24
Fix for Common Parameter Name used in Generated code

### LINKED ISSUE #1724: [Bug]: Parameter named ex causes compile error
### Describe the bug 🐞

When using a parameter named `ex` in a method of an interface used by refit, there's a name clash in the generated code that prevents compilation.

### Step to reproduce

- Create an interface with a method like so:
``` csharp
interface IMyClient 
{
    [Get("/")]
    Task MyMethodAsync(string ex);
}
```
- Try to compile
- You will get the following error:
```
error CS0136: A local or parameter named 'ex' cannot be declared in this scope because that name is used in an enclosing local scope to define a local or parameter
```

### Reproduction repository

_No response_

### Expected behavior

It should compile

### Screenshots 🖼️

_No response_

### IDE

Visual Studio 2022

### Operating system

Windows

### Version

10

### Device

_No response_

### Refit Version

7.1.0

### Additional information ℹ️

_No response_

ChrisPulman:

closes #1724

ex is used within the generated code, this is a common name

**What is the new behavior?**

Updated ex to be refit_ex to minimise the likelihood that there will be conflict.

**What might this PR break?**

None expected due to new name

**Other information**:

### DIFF
--- a/InterfaceStubGenerator.Shared/InterfaceStubGenerator.cs
+++ b/InterfaceStubGenerator.Shared/InterfaceStubGenerator.cs
@@ -535,9 +535,9 @@ namespace Refit.Implementation
             {{
                 {@return}({returnType})______func(this.Client, ______arguments){configureAwait};
             }}
-            catch (global::System.Exception ex)
+            catch (global::System.Exception refit_ex)
             {{
-                throw ex;
+                throw refit_ex;
             }}
 "
             );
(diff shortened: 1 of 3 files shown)


TimothyMakkison:
Would `______refit_ex` or `______ex` be better here? Similar to how the variables and properties are handled.

ChrisPulman:
> Would `______refit_ex` or `______ex` be better here? Similar to how the variables and properties are handled.

Yes, that's a fair point, I am happy with that, I will update to `______ex` later.
I was thinking along the lines of having refit in the name of the generated code would be more of an identification of source. But will stick with the previous trend until now.

PR #1737, opened 2024-06-25
feat: fix existing query values bug

### LINKED ISSUE #1732: [Bug]: Query parameters in path are reversed
### Describe the bug 🐞

Query parameters defined in the path are reversed in the resulting request.

### Step to reproduce

See `RequestBuilderTests.ParametersShouldBePutAsExplicitQueryString`.

```C#
 [Get("/query?q1={param1}&q2={param2}")]
 Task QueryWithExplicitParameters(string param1, string param2);
```
Becomes `/query?q2=value2&q1=value1`

### Reproduction repository

### Expected behavior

Expected `/query?q1=value1&q2=value2`

### Cause
The code that causes this is found in `RequestBuilderImplementation`, path queries are readded to `queryParamsToAdd` to the start using insert. We can prevent this by either iterating `query.AllKeys` in reverse or by keeping count of how many items have been prepended and using this number as an insert index.
```C#
var query = HttpUtility.ParseQueryString(uri.Query ?? "");
foreach (var key in query.AllKeys)
{
    if (!string.IsNullOrWhiteSpace(key))
    {
        queryParamsToAdd.Insert(
            0,
            new KeyValuePair<string, string?>(key, query[key])
        );
    }
                }
```

TimothyMakkison:
Closes #1732 
Fixes a bug where existing query parameters would have their positions reversed #1732

- Extracted logic into a method
- If the existing `Query` is empty we early return
  - This early return prevents the allocation of an empty 300 byte `NameValueCollection`
- Otherwise the query is parsed with each key pair added
  - There might be room for improvement here by using a linq `Select` method. `Select` would know the length of the collection and `InsertRange` would be able to bulk copy the values. This could only be done if we know for certain that the key can't be empty and `Where` is not required.

### Before
| Method                   | Mean      | Error     | StdDev    | Gen0   | Gen1   | Allocated |
|------------------------- |----------:|----------:|----------:|-------:|-------:|----------:|
| ConstantRouteAsync       |  2.606 us | 0.0251 us | 0.0235 us | 0.7668 | 0.0038 |   7.06 KB |
| DynamicRouteAsync        |  3.306 us | 0.0589 us | 0.0550 us | 0.8087 |      - |   7.45 KB |
| ComplexDynamicRouteAsync |  4.635 us | 0.0730 us | 0.0683 us | 0.8926 | 0.0076 |   8.27 KB |
| ObjectRequestAsync       |  5.169 us | 0.0992 us | 0.0974 us | 0.9766 | 0.0076 |   8.99 KB |
| ComplexRequestAsync      | 13.306 us | 0.1515 us | 0.1343 us | 1.6479 |      - |  15.24 KB |

### Query changes
| Method                   | Mean      | Error     | StdDev    | Gen0   | Gen1   | Allocated |
|------------------------- |----------:|----------:|----------:|----|-------:|----------:|
| ConstantRouteAsync       |  2.696 us | 0.0302 us | 0.0236 us | 0.7362 | 0.0038 |   6.77 KB |
| DynamicRouteAsync        |  3.369 us | 0.0666 us | 0.0684 us | 0.7782 |      - |   7.16 KB |
| ComplexDynamicRouteAsync |  4.529 us | 0.0412 us | 0.0344 us | 0.8621 |      - |   7.97 KB |
| ObjectRequestAsync       |  5.061 us | 0.0476 us | 0.0445 us | 0.9384 | 0.0076 |   8.69 KB |
| ComplexRequestAsync      | 13.274 us | 0.1464 us | 0.1222 us | 1.6174 | 0.0153 |  14.94 KB |

### DIFF
--- a/Refit.Tests/RequestBuilder.cs
+++ b/Refit.Tests/RequestBuilder.cs
@@ -2531,7 +2531,7 @@ namespace Refit.Tests
 
             var uri = new Uri(new Uri("http://api"), output.RequestUri);
 
-            Assert.Equal("/query?q2=value2&q1=value1", uri.PathAndQuery);
+            Assert.Equal("/query?q1=value1&q2=value2", uri.PathAndQuery);
         }
 
         [Fact]
--- a/Refit/RequestBuilderImplementation.cs
+++ b/Refit/RequestBuilderImplementation.cs
@@ -980,17 +980,7 @@ namespace Refit
                 // UriBuilder business so that we preserve any hardcoded query
                 // parameters as well as add the parameterized ones.
                 var uri = new UriBuilder(new Uri(new Uri("http://api"), urlTarget));
-                var query = HttpUtility.ParseQueryString(uri.Query ?? "");
-                foreach (var key in query.AllKeys)
-                {
-                    if (!string.IsNullOrWhiteSpace(key))
-                    {
-                        queryParamsToAdd.Insert(
-                            0,
-                            new KeyValuePair<string, string?>(key, query[key])
-                        );
-                    }
-                }
+                ParseExistingQueryString(uri, queryParamsToAdd);
 
                 if (queryParamsToAdd.Count != 0)
                 {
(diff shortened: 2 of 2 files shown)


ChrisPulman:
Nice, I really want to make some kind of live test, for example two console apps that can execute to test the functionality more fully. There's many parts of the code which are either untestable or have no existing tests. 
If you have any ideas how we can achieve this, please shout.

ChrisPulman (approved):

TimothyMakkison:
> I really want to make some kind of live test, for example two console apps that can execute to test the functionality more fully.

Like a sandbox project? It could be a single file with an interface and dummy http request handler.

>There's many parts of the code which are either untestable or have no existing tests.

Which parts did you have in mind? When experimenting with the source generator I have added snapshot testing. The current system is okay because the generator is basic, but any future improvements will need a better test setup. See mapperly as a good example (not biased)😄

PR #1739, opened 2024-06-25
feat: refactors, cache attributes, use helper methods

TimothyMakkison:
- Replace `BodylessMethods HashSet<HttpMethod>` with `IsBodyless` function
- Prevent `headerCollection` from allocating a new dictionary if the value isn't a dictionary.
  - This shouldn't be possible, I'm not sure if this should be an error.
- Use a default static `QueryAttribute` if `queryAttribute` is null.
- Cache `QueryUriFormatAttribute` in `RestMethodInfo` instead of every http request.
- Use a switch expression for `collectionFormat`

Sorry about the hundreds of prs, I did a code review a while back and am slowly getting through my list of suggestions. I don't want to create one giant hard to follow pr.


ChrisPulman:
@TimothyMakkison  This has a merge conflict, could you solve please, thank you

ChrisPulman (approved):

TimothyMakkison:
Mb, thought I'd solved them. Give me 5 kins

PR #1740, opened 2024-06-26
chore: extract methods

TimothyMakkison:
Extracted methods into:
- `AddBodyToRequest`
- `AddQueryParameters`
- `AddMultiPart`
- `AddHeadersToRequest`
- `AddPropertiesToRequest`

I haven't changed any logic, this commit is to make it easier to understand the scope of logic and to cut down on the massive function. I plan on removing the following collections in a subsequent pr 🤞 

```c#
var queryParamsToAdd = new List<KeyValuePair<string, string?>>(); // either use `ValueStringBuilder` or `ValueListBuilder`
var headersToAdd = new Dictionary<string, string?>(restMethod.Headers); `ValueListBuilder` or add directly with a check
var propertiesToAdd = new Dictionary<string, object?>(); // iterate params later on and directly add to `HttpRequest` both property/option are internally dictionaries.
```

### DIFF
--- a/Refit/RequestBuilderImplementation.cs
+++ b/Refit/RequestBuilderImplementation.cs
@@ -726,70 +726,7 @@ namespace Refit
                         && restMethod.BodyParameterInfo.Item3 == i
                     )
                     {
-                        if (param is HttpContent httpContentParam)
-                        {
-                            ret.Content = httpContentParam;
-                        }
-                        else if (param is Stream streamParam)
-                        {
-                            ret.Content = new StreamContent(streamParam);
-                        }
-                        // Default sends raw strings
-                        else if (
-                            restMethod.BodyParameterInfo.Item1 == BodySerializationMethod.Default
-                            && param is string stringParam
-                        )
-                        {
-                            ret.Content = new StringContent(stringParam);
-                        }
-                        else
-                        {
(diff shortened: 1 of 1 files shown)


ChrisPulman:
@TimothyMakkison next conflict 😊

TimothyMakkison:
> @TimothyMakkison next conflict 😊

Should be done 🤞 

I've been looking at removing the header dicitonary and realised that I have to support .Net framework 😄 I'm currently planning on creating an inlinearray and converting it into a `ValueListBuilder<KeyValuePair<string, string?>>` for headers. The only way this is possible in .Net framework wou;d be the use of unsafe code. 

Would you be okay with this? Otherwise I can use some preprocessors and add use a list for .Net framework.

ChrisPulman:
Let's keep the Netframework as a special case, there's a lot of companies that won't move forward until the last second and adopt new technologies. So long as the functionality of all TFM's are equal with regards to the output then we can even keep a separated section for Netframework, but it's a no from me for unsafe calls.
I am happy to provide a highly performant Netstandard2.0 / net6.0+ Version and if possible include Net462 otherwise create patches to keep the API and output standard.

ChrisPulman (approved):

Issue #1389, opened 2022-08-01
feature: Refit should be linker-friendly and support trimming


Sergio0694:
**Is your feature request related to a problem? Please describe.**

Yes. The issue is that despite Refit uses source generators, the generated code is just a thin proxy back to the reflection-powered functionality in the main library. This means that using Refit with trimming enabled **completely breaks it**. As it is, there's not much value in having source generators. They should be updated so that all the reflection-powered logic is executed at compile time, and the generated stubs will just have all the logic that is today done at runtime by the method builder APIs. This library should effectively require no runtime reflection at all (or, virtually none, but especially none that can break trimming).

**Describe the solution you'd like**

As mentioned, the source generator should be updated so that instead of just calling back into the method builder and invoking that reflection-based callback to perform the request, it'd instead emit all the same logic that's today done at runtime. That is, all the various marshalling of parameters, request building etc. should just be done in a statically-analyzeable way and generated at compile time. The library should just expose helper APIs for the generator to use to construct the individual bits of a request/response.

All the information that the method builder is using at runtime is there at compile time too anyway. That logic should just essentially be inverted, and the generator APIs should be used to achieve static reflection, instead of runtime reflection like today.

If there's any reflection left, additionally, all those APIs should be properly annotated.

Refit as a whole should have the trimming analyzers enabled when on the .NET 6+ target.

**Describe alternatives you've considered**

Runtime directives can be used to patch the problem. This is extremely cumbersome, hard to do especially for beginners, and causes an unnecessary binary size increased, given you'd be forced to preserve way more stuff than strictly needed, to be safe.

**Describe suggestions on how to achieve the feature**

- Investigate all the individual bits that method builder uses to construct the request.
- Expose the necessary helper APIs for them from that type.
- Update the generator to do all that introspection at compile time.
- Update the generator to perform the needed calls to those helpers to construct and execute the request right there.

Sergio0694:
Hey @clairernovotny! 👋
Can we coordinate with you to discuss, plan and eventually implement this feature, or would you be able to connect us with the right folks to talk about this? @Kant2002 has kindly offered to help with this (he has experience doing very similar work, and has contributed to several source generator and trimming related work on other projects in the past), and he was wondering how to go about this, given that the scope of the work involved for this really needs some discussion first 🙂

[...]

PR #1746, opened 2024-06-29
First step to target fully source gen implementation

kant2002:
I extract simple model for Refit client representation withing source gen That to allow facilitate process
- Parse Syntax (done by Roslyn itslef)
- Build internal model
- Write based on model only, so logic would be as simple as possible

Also that allows write more robust test if we test just internal model extraction from symbols. Otherwise if we will capture generated output, if we decide improve generation it will produce a lot of churn. I do not agains churn like that, but not sure that's would be appropriate.

Related to https://github.com/reactiveui/refit/issues/1389


glennawatson:
Thanks @kant2002, and the direction here is sound (a model-based generation step for simpler, more testable codegen). It's been conflicting and stalled for ~18 months and the source-gen/AOT work has moved on since, so closing to keep things tidy rather than leave it lingering. The idea is still welcome — please reopen or rebase a fresh PR against current `main` if you'd like to pick it up.

PR #1753, opened 2024-07-01
feat: added snapshot testing for generated code

TimothyMakkison:
Created a separate file for snapshot testing the generated code.
- Verify doesn't support .NET 462 so I'm using NET 481, let me know if this is a deal breaker.
- I want to move the generated code tests from `Refit.Tests` to this project.
- This is intended to support a possible source generator rewrite.


ChrisPulman:
Net 4.7.2 is supported by Verify I would rather keep a lower end test than upper, would this be acceptable, the other option is to move the API Tests to a separate test project. 
I wanted to get the API tests added and for the moment to avoid breaking changes, I have excluded net 4.6.2 from the API tests and am only checking net 6 and net 8

TimothyMakkison:
> Net 4.7.2 is supported by Verify I would rather keep a lower end test than upper, would this be acceptable,

Absolutely, no idea why I didn't use this version. Its a shame as I'd prefer to test the generated code for .NET 461 as its the most different version (Iirc `Unsafe.As` isn't available.)

> the other option is to move the API Tests to a separate test project.
I wanted to get the API tests added and for the moment to avoid breaking changes, I have excluded net 4.6.2 from the API tests and am only checking net 6 and net 8

Which API tests are you talking about, are you talking about all the tests in `Refit.Tests`? I wasn't aware you'd recently added tests 🤔 


I'm also not sure if my generator tests are even using .NET Framework, NET 6.0, NET 8.0 when running. AFAIK `refit.Tests\InterfaceStubGenerator` uses `VerifyCSV2.Test`and `ReferenceAssemblies` to build the given file in the correct version (I think). I don't know how `CSharpCompilation.Create` chooses which version of .NET to use.

I'm tempted to close this PR and use it for testing a source generator rewrite (if I ever get round to it). Moving the tests from  `InterfaceStubGenerator` isn't the main goal. I was planning on using this project for tests like the following:

```C#
public async Task TestRouting()
{
  // fixture adds this method into a blank refit interface file and verifies the output.
  await Fixture.VerifyMethod(
  """
  [Get("/group/{id}/users")]
  Task<List<User>> GroupList([AliasAs("id")] int groupId, [AliasAs("sort")] string sortOrder);
  """);
}

Issue #1791, opened 2024-08-27
Improve incremental generator caching


TimothyMakkison:
While working on a more source generated approach, I noticed that Refit does not correctly use incremental generator caching. While  a previous PR uses some of the more modern techniques such as `CreateSyntaxProvider`, the lines [829-846](https://github.com/reactiveui/refit/blob/7cfdeded6363adf5c7eb0d3902db528e92a17f17/InterfaceStubGenerator.Shared/InterfaceStubGenerator.cs#L829) are functionally useless and will never successfully cache anything. This is due to how both `ImmutableArray` handles equality and the passing of the reference type `Compilation` (this will always change between re-runs).

- Methods should be represented by a structurally comparable object
- A benchmark could be created to confirm the benefits. Its possible that generating the code in a single pass is faster than caching.
- Tests should be added to verify that the caching works as intended
- Why is `PreserveAttribute.cs` added and then its [namespace queried](https://github.com/reactiveui/refit/blob/7cfdeded6363adf5c7eb0d3902db528e92a17f17/InterfaceStubGenerator.Shared/InterfaceStubGenerator.cs#L231)? Would the compiler change this? I assume it would always be `global::RefitInternalGenerated.PreserveAttribute`
- `CancellationToken` [should be respected](https://github.com/dotnet/roslyn/blob/main/docs/features/incremental-generators.md#handling-cancellation) to reduce IDE lag

I'll start work on a PR for this but I don't promise anything.

Related https://github.com/reactiveui/refit/issues/1389

ChrisPulman:
Thanks Tim, much appreciated, I was going to head down the same rabbit hole and try to make things better, I see that AoT is important to many people too :)

TimothyMakkison:
Hey, added some preliminary benchmarks in #1798.

I was planning on adding some tests to verify steps in the incremental generator like I've done for [Mediator](https://github.com/martinothamar/Mediator/blob/main/test/Mediator.SourceGenerator.Tests/Incremental/IncrementalGeneratorTest.cs). I used `WithTrackingName` to check specific steps. Unfortunately this is only available in Roslyn 4.1+. I believe I could either:
- Update `InterfaceStubGenerator.Roslyn40` to `InterfaceStubGenerator.Roslyn41` (probably a breaking change)
- Add `InterfaceStubGenerator.Roslyn41` and use it for the tests.

PR #1816, opened 2024-09-16
chore: rename tests add derived interface test

TimothyMakkison:
- Renamed `DerivedRefitInterfaceTest` and `DerivedNonRefitInterfaceTest`
- Added `InterfaceDerivedFromRefitBaseTest`


TimothyMakkison:
@ChrisPulman did you have any thoughts regarding my [comment?](https://github.com/reactiveui/refit/issues/1791#issuecomment-2322902743) The lack of incremental generator tests are the main blocker preventing me from adding incremental generation.

ChrisPulman:
> @ChrisPulman did you have any thoughts regarding my [comment?](https://github.com/reactiveui/refit/issues/1791#issuecomment-2322902743) The lack of incremental generator tests are the main blocker preventing me from adding incremental generation.

Could we go with Add a new **InterfaceStubGenerator.Roslyn41** option for the time being, this way we can maintain older versions as they currently exist and open a path to pure incremental generators, we can drop the relevant support if required once we get a stable version of 41.

ChrisPulman (approved):

ChrisPulman:
Roslyn 38, 40 and 41 are all supported by Rider, VS2019 (Fully updated) and VS2022, so really using Roslyn 41 shouldn't be an issue.

ChrisPulman:
I am happy with updating the version of Rosyln to a single newer version rather than trying to support 2 or 3 versions, this would involve changing the project setup to have a single NetStandard project instead of a shared project. 
Would you be okay with going down that route? 
While understanding that we will drop support for some versions of Visual Studio we can advance Refit to be what people are calling for. So, while keeping a version as low as possible i.e. Roslyn 41, I am happy with anything up to Roslyn 46 if that makes anything easier.

TimothyMakkison:
> Would you be okay with going down that route?

I'm happy with either solution. My current solution was to update Roslyn 4.0 to Roslyn 4.1, this way we keep support for 3.9.

> we will drop support for some versions of Visual Studio we can advance Refit to be what people are calling for.
 
Personally, none of my planned changes require dropping support for anything. I only need 4.1 to use [`WithTrackingName`](https://learn.microsoft.com/en-us/dotnet/api/microsoft.codeanalysis.incrementalvalueproviderextensions.withtrackingname?view=roslyn-dotnet-4.7.0) to determine if the generator is caching correctly. This would be an extremely minimal change regardless of which approach is taken.

Issue #1817, opened 2024-09-17
[Bug]: IExampleInterface doesn't look like a Refit interface


aywee26:
### Describe the bug 🐞

The project is ASP.NET Core Web API that uses Refit packages (Refit 7.2.0, Refit.HttpClientFactory 7.2.0, Refit.Newtonsoft.Json 7.2.0). It contacts 3rd party resource JSONPlaceholder via HTTP.

The interface IBlogApi lists several methods, all of them are annotated by appropriate Refit attributes.

```
public interface IBlogApi
{
    [Get("/posts/{id}")]
    Task<Post> GetPostAsync(int id);

    [Get("/posts")]
    Task<List<Post>> GetPostsAsync();

    [Post("/posts")]
    Task<Post> CreatePostAsync([Body] Post post);

    [Put("/posts/{id}")]
    Task<Post> UpdatePostAsync(int id, [Body] Post post);

    [Delete("/posts/{id}")]
    Task DeletePostAsync(int id);

    [Get("/posts")]
    Task<List<Post>> GetPostsAsync([Query] PostQueryParameters parameters);
}
```

The project uses source-generated implementations of these methods to provide various routes. When "GET /posts/1" query is performed, the following exception is thrown:

```
System.InvalidOperationException: IBlogApi doesn't look like a Refit interface. Make sure it has at least one method with a Refit HTTP method attribute and Refit is installed in the project.
   at Refit.RestService.GetGeneratedType(Type refitInterfaceType) in c:\temp\releaser\refit\Refit\RestService.cs:line 188
   at System.Collections.Concurrent.ConcurrentDictionary`2.GetOrAdd(TKey key, Func`2 valueFactory)
   at Refit.RestService.For(Type refitInterfaceType, HttpClient client, IRequestBuilder builder) in c:\temp\releaser\refit\Refit\RestService.cs:line 79
   at Refit.RestService.For[T](HttpClient client, IRequestBuilder`1 builder) in c:\temp\releaser\refit\Refit\RestService.cs:line 19
   at Refit.HttpClientFactoryExtensions.<>c__2`1.<AddRefitClient>b__2_3(HttpClient client, IServiceProvider serviceProvider) in c:\temp\releaser\refit\Refit.HttpClientFactory\HttpClientFactoryExtensions.cs:line 87
   at Microsoft.Extensions.DependencyInjection.ServiceLookup.CallSiteVisitor`2.VisitCallSiteMain(ServiceCallSite callSite, TArgument argument)
[...]

PR #1819, opened 2024-09-18
feat: revert 29e0e1c

TimothyMakkison:
Fix for #1817 by reverting 29e0e1c. Not sure if `Microsoft.CodeAnalysis.CSharp` was intentionally updated to the latest versions, guessing it broke the source generator for anyone on an older ide.


TimothyMakkison:
Update, appears to fix it for me on Rider.

I tested this by creating a Sandbox application see here #1820.

ChrisPulman (approved):

ChrisPulman:
@TimothyMakkison I will find a way to lock these versions, thank you.
BTW, are you on Slack?

PR #1856, opened 2024-10-02
[Doc] Adding a friend projects section

JeremyBP:
Adding a friend projects final section into the doc

Doc doesn't talk about projects built ont top of Refit or using it to add some value

**What is the new behavior?**
Doc talks about project built ont top of Refit or using it to add some value

**What might this PR break?**
Nothing, just a doc update

**Other information**:
The PR about discussion #1832 
Just thinking it could be usefull to let Refit's users know that some projects around there may help them building things with Refit and more :)


glennawatson:
Thanks for putting this together @JeremyBP. We've decided not to maintain a "friend / related projects" section in the docs — curating third-party links is hard to keep current and even-handed, so we'd rather not take it on. Closing with thanks; this is no reflection on the projects themselves.

PR #1859, opened 2024-10-03
feat: added source gen tests for generic constraints

TimothyMakkison:
Rel #1791

Will have to be updated after #1861


TimothyMakkison:
Thanks for merging #1861, I've updated the tests to reflect the corrected version.

[Regarding](https://github.com/reactiveui/refit/pull/1819#issuecomment-2359335657), never used it before but I've created an account with the same email address as the one I use for github. Not sure if I set it up correctly 😅 

Otherwise I'm pretty responsive on github

ChrisPulman (approved):

PR #1864, opened 2024-10-06
feat: update `InterfaceStubGenerator` to be an incremental generator

### LINKED ISSUE #1791: Improve incremental generator caching

TimothyMakkison:
Closes #1791

- `Parser` converts each `INamedTypeSymbol` into `InterfaceModel` and extracts additional information.
- All `Model` types should be structurally equatable to enable incremental generation.
  - `ImmutableEquatableArray` is a structurally equatable collection.
- `Emitter` takes each model and creates an interface stub.
- I’ve tried to keep the logic and structure as close to the original as I could to prevent errors.

### TODO
- Add more tests
- Delete unused code
- Need to do a **thorough** comparison between the old logic and new to ensure that I haven't made a translation error.

### Future changes
- Remove some of the unneeded `ToList` calls
- Optimize the model memory usage. ie `ImmutableEquatableArray` could be made into a struct.
- I'd like to copy `System.Text.Json` and rename files to `InterfaceStubGenerator.Parser` etc

#### Original
| Method      | Mean       | Error     | StdDev    | Gen0     | Gen1     | Allocated  |
|------------ |-----------:|----------:|----------:|---------:|---------:|-----------:|
| Compile     |   249.6 us |   4.97 us |   6.46 us |  12.6953 |   2.9297 |  117.52 KB |
| Cached      |   216.1 us |   4.10 us |   4.02 us |  12.2070 |   2.9297 |  114.86 KB |
| CompileMany | 8,898.8 us | 171.75 us | 197.78 us | 687.5000 | 343.7500 | 6354.62 KB |
| CachedMany  | 7,897.2 us | 139.81 us | 137.32 us | 671.8750 | 328.1250 | 6289.45 KB |

#### Incremental

| Method      | Mean       | Error     | StdDev    | Gen0     | Gen1     | Gen2    | Allocated  |
|------------ |-----------:|----------:|----------:|---------:|---------:|--------:|-----------:|
| Compile     |   215.4 us |   3.76 us |   3.69 us |  12.2070 |   1.4648 |       - |   115.5 KB |
| Cached      |   158.1 us |   2.97 us |   2.91 us |   5.8594 |   0.2441 |       - |   54.44 KB |
| CompileMany | 8,142.1 us | 138.00 us | 129.08 us | 765.6250 | 296.8750 | 31.2500 | 6793.62 KB |
| CachedMany  | 5,173.9 us |  97.88 us | 100.52 us | 156.2500 |  54.6875 |       - |  1455.1 KB |


ChrisPulman:
Hi @TimothyMakkison , do you think you could try to update dependency verify.sourcegenerators to 2.5.0 along with this PR or as a separate PR.
I think I managed to conflict with one of your updates with some of my PR's, hence the conflict, sorry.
So far this is looking good, thank you, I have updated the Minor, however this looks like a Major version change; I think we can wait until this section of work is complete before releasing again, please let me know when you think the code is ready to release should be a nice update by the looks of things. If you think it can stay with a minor update then do let me know.

TimothyMakkison:
> Hi @TimothyMakkison , do you think you could try to update dependency verify.sourcegenerators to 2.5.0 along with this PR or as a separate PR. 

Sure, I don't think changing this should affect me.
> I think I managed to conflict with one of your updates with some of my PR's, hence the conflict, sorry. 

nw, it looks like most of the changes can be copied over or ignored, shouldn't be a problem.

> So far this is looking good, thank you, I have updated the Minor, however this looks like a Major version change; I think we can wait until this section of work is complete before releasing again, please let me know when you think the code is ready to release should be a nice update by the looks of things. If you think it can stay with a minor update then do let me know.

Thanks 😄, in theory this should be an invisible change and will only provide a smoother ide experience for users. Assuming that I haven't broken anything🤞 . I'm not too familiar with versioning but this seems like a "minor" change, albeit with the chance of breaking. Does refit have a pre release version?

I was hoping to finish this in a couple of days, I need to do some cleanup, re-read `Parser`/`Emitter` a couple of times and probably write more tests.

TimothyMakkison:
🤞 should be sound. As mentioned before, I tried to keep the logic to be a straight translation from the original. Note that bools passed to `ParseMethod` may appear to be inverted as the its logic is a mix of `ProcessRefitMethod` and `WriteMethodOpening`.

ChrisPulman (approved):

ChrisPulman:
Shall we get 1801 out the way first as it's producing errors in the code, then tackle 1389

Issue #1915, opened 2024-11-06
[Bug]: "CancellationToken?" mishandling in refit requests


AFract:
### Describe the bug 🐞

Hello, 

We are happily using Refit in a project with a lot of calls like : 

```
    [Post("SomeRoute)]
    public Task<ApiResponse<SomeDto>> DoSomething(CancellationToken? cancellationToken = null);
```
(this example is for POST, but PUT or GET does the same)

We have noticed that when a CancellationToken is provided, it is handled as a query string parameter : 
![image](https://github.com/user-attachments/assets/d6b569ec-f11c-4acd-96a6-c0414010b983)
(note the IsCancellationRequested, CanBeCancelled... Appearing in the query string).

If the signature is updated as below, the CancellationToken is handled as expected:
```
    [Post("SomeRoute)]
    public Task<ApiResponse<SomeDto>> DoSomething(CancellationToken cancellationToken);
```

But we were willing to make the CancellationToken optional, hence the "?" and "= null". We didn't figure it won't work.

One workaround would be to change the signature for 
```
    [Post("SomeRoute)]
    public Task<ApiResponse<SomeDto>> DoSomething(CancellationToken cancellationToken = default);
```

to make the parameter both non nullable and optional, "default" does exactly what a CancellationToken.None does:
![image](https://github.com/user-attachments/assets/f0f66c03-1a4f-4b94-a87e-06364db8adea)

Do you have a suggestion regarding my workaround above, is it good for you ? Do you plan to improve this to handle "nullable CancellationToken" ? 

Thank you very much !

### Step to reproduce

Write a refit method like     
[Post("SomeRoute)]
public Task<ApiResponse<SomeDto>> DoSomething(CancellationToken? cancellationToken = null);
and see query string.
 
### Expected behavior

It would be nice and safer if nullable "CancellationToken?" methods signatures could be handled as non nullable "CancellationToken" instead of to be processed as query string.

### IDE

Visual Studio 2022

### Operating system

Windows

### Refit Version

Refit.Newtonsoft.Json (and refit) in version 7.2.1 with .Net 8.  Same behavior with Refit.Newtonsoft.Json in version 8.

AFract:
Thank you for your quick answer ! If it's not possible to use the nullable cancellation token, at least throwing an exception rather than serializing it in query would avoid to end up with our situation :). 
[...]

PR #1917, opened 2024-11-06
feat: prevent serialization of `CancellationToken?`

TimothyMakkison:
- Prevent `CancellationToken?` being serialized.
- Add cancellation tests to `Tests/RestService`, including `NullableCancellationTokenShouldBeIgnored` 

I'm not sure how Refit should handle nullable Cancellation token, it's a struct so it isn't nullable by default. My two ideas:

- Refit functions containing a `CancellationToken` parameter should throw an exception or emit a diagnostic.
- Accept `CancellationToken?`, check these tokens for nullability. If they are null then substitute it with a default token.

See #1915

### DIFF
--- a/Refit.Tests/RestService.cs
+++ b/Refit.Tests/RestService.cs
@@ -324,6 +324,18 @@ public interface IFragmentApi
     Task QueryAfterFragment();
 }
 
+public interface ICancellableApi
+{
+    [Get("/foo")]
+    Task GetWithCancellation(CancellationToken token = default);
+
+    [Get("/foo")]
+    Task<string> GetWithCancellationAndReturn(CancellationToken token = default);
+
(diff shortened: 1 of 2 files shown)


ChrisPulman (approved):

ChrisPulman:
I am thinking that raising a diagnostic would be the best, accepting a null token may cause more harm further down the line.

ChrisPulman:
So it seems that CancellationToken = default gets turned into CancellationToken.None by the internal workings. However a completely unset value would result in a null reference exception. So I am thinking that a nullable unset token should raise a Error level diagnostic and the use of nullable Tokens should have a warning diagnostic. I have not run any tests on these scenarios yet, and there may be some checks within HttpClient to replace the null value with None.

PR #1943, opened 2024-12-27
Proposal: Add a `TreatAsString` field to `QueryAttribute`

mark-pro:
The proposal arises from the need to pass custom types or structs which might represent types with rules and conversions. Most of these types can override `ToString()` which can then be called when constructing query parameters.

For example:

```csharp
record readonly struct ZipCode(ushort ZipCode) {
    public override string ToString() =>
        ZipCode.ToString();
}
```

To solve the problem the proposal seeks to add a field to the QueryAttribute called `TreatAsString`.

The new field is used when the query string is created and simply calls the `ToString()` function on a type.

There is a workaround that is not documented. The current workaround is to implement `IFormattable` on the type, which feels redundant if `ToString()` is already implemented on a custom type.


mark-pro:
I am open to suggestions on how to implement this functionality. In many of my project, both professional and personal, I find myself adding `IFormattable` to those types which feels a little clunky when I really need that complex type to be a string.

The solution could also work on types not owned by the implementer such as passing a GUID as a query parameter.

`GetUser([Query(TreatAsString = true)] Guid userId)`.

mark-pro:
@ChrisPulman I will work to get have the project build successfully and will message again when it is.

mark-pro:
@ChrisPulman I have updated the unit tests. They are now passing in my fork of the project. The latest build run is awaiting approval.

- https://github.com/reactiveui/refit/actions/runs/15882821586

ChrisPulman (approved):

PR #1964, opened 2025-04-02
test: add `IObservable<IApiResponse<T>>` test

TimothyMakkison:
Add test for `IObservable<IApiResponse<T>>`

### DIFF
--- a/Refit.Tests/GitHubApi.cs
+++ b/Refit.Tests/GitHubApi.cs
@@ -86,6 +86,9 @@ public interface IGitHubApi
     [Get("/users/{username}")]
     IObservable<ApiResponse<User>> GetUserObservableWithMetadata(string userName);
 
+    [Get("/users/{username}")]
+    IObservable<IApiResponse<User>> GetUserIApiResponseObservableWithMetadata(string userName);
+
     [Post("/users")]
     Task<User> CreateUser(User user);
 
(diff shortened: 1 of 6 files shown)


ChrisPulman:
Hi @TimothyMakkison were you expecting this to be building and passing the tests at this point, or is there further work to be done?

TimothyMakkison:
I thought it was a simple test addition 😔 

I have no idea who it keeps failing. The failure message doesn't even seem related to my change no?

TimothyMakkison:
Nvm githubapi is used by source generator test hence why it's breaking

ChrisPulman:
> Nvm githubapi is used by source generator test hence why it's breaking
> 
> Should I move this test to a different file or fix the generator test?

Perhaps fixing the generator test would be best. I am currently working in Australia so haven't had much availability. Back to the UK on Tuesday

TimothyMakkison:
> Perhaps fixing the generator test would be best

👍 

Is it okay if I add verify to `Refit.Tests`, updating this file was a massive pain for #1967

I can't remember why I didn't do this before

ChrisPulman:
> > Perhaps fixing the generator test would be best
> 
> 👍
> 
> Is it okay if I add verify to `Refit.Tests`, updating this file was a massive pain for #1967
> 
> I can't remember why I didn't do this before

Yes that's fine, we use it widely across the ReactiveUI projects

TimothyMakkison:
Build keeps failing because the initialisation of verify is somehow running twice despite being inside a static constructor. I have no idea how this happens, I thought type initialisation code only ran once.

Normally this would be fixed by using `ModuleInitializer` to run the code once, unfortunately net framework doesn't support this.
Instead I've added a static bool to check if initialisation code has been run yet.

ChrisPulman (approved):

Issue #1885, opened 2024-10-18
Refit should emit diagnostics for invalid interfaces


TimothyMakkison:
Refit should emit diagnostic warnings and errors for invalid methods and interfaces. Many exceptions in `RequestBuilderImplementation` could be statically determined via a source generator.

- Invalid url backslash 
- Multiple `CancellationToken` params
- `HeadersCollection` must be `IDictionary`

Severity to be determined layer.

More advanced checks may to read the attribute values. It isn't required but I'd prefer to reference `../Refit.csproj` and construct each attribute in the source generator. I don't know if this could cause issues 🤔 

Would this be considered to be a breaking change? It may prevent users from building code that previously ran, although it was likely invalid.

glennawatson:
This is addressed in the Refit 14 beta, now available on NuGet: https://www.nuget.org/packages/Refit/14.0.0-beta.1

Please give it a try - feedback ahead of the stable 14.0.0 release is very welcome.

<blockquote><img src="https://api.nuget.org/v3-flatcontainer/refit/14.0.0-beta.1/icon" width="48" align="right"><div><strong><a href="https://nuget.org/packages/Refit/">Refit 14.0.0-beta.1</a></strong></div><div>Type-safe REST API clients for .NET, generated from annotated interfaces with build-time source generation, direct request construction, System.Text.Json defaults, and support for trimming and AOT.</div></blockquote>

Issue #1882, opened 2024-10-15
[Bug]: derived types erroneously add additional query parameters.


TimothyMakkison:
### Describe the bug 🐞

Derived types will cause properties to be used in both the path and query.

### Step to reproduce

```c#
var client = RestService.For<IApi>("http://bar");

// http://bar/foo/road?MyQuery=quiz
var _ = await client.Send(new MyObject { MyPath="road", MyQuery = "quiz"});

// http://bar/foo/road?MyPath=road&MyQuery=quiz&MyQuery2=quibble
// adds MyPath as a query parameter
// expected:
// http://bar/foo/road?MyQuery=quiz&MyQuery2=quibble
var _ = await client.Send(new MyDerived { MyPath="road", MyQuery = "quiz", MyQuery2 = "quibble"});

public interface IApi
{
    [get("/foo/{request.MyPath}"]
    Task<string> Send(MyObject request);
}

public class MyObject
{
    public string MyPath { get; set; }
    public string MyQuery { get; set; }
}

public class MyDerived : MyObject { public string MyQuery2 {get; set; } }
```

### Reproduction repository

https://github.com/reactiveui/refit

### Expected behavior

Refit should not add a property as a query parameter if it is already mapped to the path.

Cause:
- Refit constructs `RestMethodInfoInternal` using `MethodInfo`
- `BuildParameterMap` identifies that the url is parameterised and that the corresponding property belongs to a parameters property. 
- `request.MyPath`, is added to the `ParameterProperties` list, adding the `ParameterInfo` for `MyBase.MyPath`. 
- `RequestBuilderImplementation` uses this list in `BuildQueryMap` to [ensure that it doesn't add properties already used in the path to the query.](https://github.com/reactiveui/refit/blob/5df30d9df9ca45e185f9d5ec5eb3545c57bf7dc2/Refit/RequestBuilderImplementation.cs#L502)
  - Note that `BuildQueryMap` uses reflection for each object passed to it. In our case it is building a query map for the type `MyDerived`.
- This comparison compares the current property `MyDerived.MyPath` to any values in `ParameterProperties`.
  - The relevant value has a `ParameterInfo` for `MyBase.MyPath`, the comparison fails because they are different properties.
  - Refit adds `MyDerived.MyPath` to the query.
[...]

PR #1965, opened 2025-04-03
Aot gen

TimothyMakkison:
Incomplete source generated refit logic. Needless to say this PR is still in progress, I tried to throw it together.

Future Plans
- Fix the todos
- I need to do a lot of refactoring
- Check logic
- Add `Query` and `MultiPart` support
- Hopefully figure out how to reference `Refit`types in the source generator without it breaking
  - This is why I had to add the Configuration types
- If my understanding of AOT is correct I can safely get the interface `MethodInfo` and pass it to functions.
- Support #1885
- Rewrite half the tests to work with source generated logic :(

If we go ahead with this a couple of things may break.
- I don't think derived attributes will work (Is this a backwards compatibility thing)
- #1882 will have to be fixed. Anything that does reflection on a runtime type will have to be changed (I think, I've never used AOT)

image

### DIFF
--- a/InterfaceStubGenerator.Shared/AttributeDataAccessor.cs
+++ b/InterfaceStubGenerator.Shared/AttributeDataAccessor.cs
@@ -1,0 +1,220 @@
+﻿// using System.Reflection;
+//
+// using Microsoft.CodeAnalysis;
+// using Microsoft.CodeAnalysis.CSharp.Syntax;
+// using Microsoft.CodeAnalysis.Operations;
+//
+// namespace Refit.Generator
+// {
+//     public class AttributeDataAccessor(WellKnownTypes symbolAccessor)
+//     {
+//         private const string NameOfOperatorName = "nameof";
+//         private const char FullNameOfPrefix = '@';
+//
+//         public TAttribute AccessSingle<TAttribute>(ISymbol symbol)
+//             where TAttribute : Attribute => AccessSingle<TAttribute, TAttribute>(symbol);
+//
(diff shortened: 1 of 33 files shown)


TimothyMakkison:
Okay, after reading my notes and doing a couple of experiments:
- We might be able to support attributes inheriting from attributes, it will be very awkward and doable, although I'm not sure it would be worth it.
- Query parameters won't be able to support runtime serialisation. (See below)
  - It will be possible to create a compile time serializer, but it won't handle inheritance.
  - I could make this backwards compatible by:
    - Making users add an attribute somewhere to enable AotRefit. This would tell the generator to emit a compile time query serialiser.
    - Emit some code that tries to use runtime serialisation, if it fails we switch to a compile time version.
      - This would generate a load of 
    - Determine if the project has `PublishAot` enabled (I don't know if this is possible)
 and emit compile time serialisation code 


```C#
void Descend([DynamicallyAccessedMembers(DynamicallyAccessedMemberTypes.PublicMethods)] Type t)
{
    foreach (var p in t.GetProperties())
    {
        Console.WriteLine(p.Name);
        Descend(p.PropertyType); //  't' argument does not satisfy 'DynamicallyAccessedMemberTypes.PublicMethods'
    }
}
```

glennawatson:
Thanks for taking a swing at this @TimothyMakkison. This is an explicitly-incomplete draft (conflicting, ~14 months stale) and the AOT / source-gen direction is being carried forward elsewhere (see #1389), so closing to keep the PR list clean. No problem at all picking it back up — reopen or open a fresh PR against current `main` if you return to the AOT gen work.

PR #1981, opened 2025-04-28
Add the ability to register Refit clients as keyed services

### LINKED ISSUE #1876: [Feature] Add support for keyed services
It would be useful to register the same API client for multiple accounts. Imagine the following situation:
* Account A
* Account B

Based on some data from the request, I would like to be able to select an appropriate account and set a correct `Authorization` header.

Setting the `Authorization` header requires obtaining an access token. This logic is already done in the delegating HTTP handler. Unfortunately, the only way to pass some data to that handler (and indicate which account should be used) is to add the `Property` option, for example:

```csharp
public interface IClient
{
    [Post("/api/test")]
    Task<Response> DoSomethingAsync([FromBody] Request request, [Property("Account")] string account);
}
```

If the client has multiple endpoints, it is not a convenient way to do that. We can use [keyed services](https://learn.microsoft.com/en-us/aspnet/core/fundamentals/dependency-injection?view=aspnetcore-8.0#keyed-services) instead.

[Keyed services](https://learn.microsoft.com/en-us/aspnet/core/fundamentals/dependency-injection?view=aspnetcore-8.0#keyed-services) were introduced in .NET 8. Instead of using the approach mentioned above, we could do:
```csharp
services
    .AddRefitClient<IClient>("Account A")
    .ConfigureHttpClient(client =>
        client.DefaultRequestHeaders.Authorization = new AuthenticationHeaderValue("Bearer", "***")
    );
services
    .AddRefitClient<IClient>("Account B")
    .ConfigureHttpClient(client =>
        client.DefaultRequestHeaders.Authorization = new AuthenticationHeaderValue("Bearer", "***")
    );

// or
services
    .AddRefitClient<IClient>("Development")
    .ConfigureHttpClient(client => client.BaseAddress = new Uri("https://dev.api.com"));
services
    .AddRefitClient<IClient>("Production")
    .ConfigureHttpClient(client => client.BaseAddress = new Uri("https://prod.api.com"));

// or
services
    .AddRefitClient<IClient>("v1")
    .ConfigureHttpClient(client => client.BaseAddress = new Uri("https://api.com/v1"));
services
    .AddRefitClient<IClient>("v2")
    .ConfigureHttpClient(client => client.BaseAddress = new Uri("https://api.com/v2"));
```

And later use an `IKeyedServiceProvider` or `IServiceProvider` to get the appropriate client:
```csharp
var client = keyedServiceProvider.GetRequiredKeyedService<IClient>("Account A");
```

As mentioned above, we can alternatively use the `Property` option or register Refit clients for different interfaces. However, neither of these approaches is convenient.

Maybe adding an API for .NET 8 and greater to register keyed Refit clients could solve that issue.

**Additional context**
Feel free to suggest other ideas 🙂

fubar-coder:
This PR fixes #1876.

1. Added new constructor to `AuthenticatedHttpClientHandler` which doesn't automatically set the `InnerHandler` property
2. Adds extension methods to register the Refit client as keyed service
3. Added unit tests

# Why did I add the new constructor to `AuthenticatedHttpClientHandler`?

- To keep the PR as small as possible by avoiding the duplication of code
- Enable the a behavior compatible with `IHttpClientBuilder` by ...
  - ... avoiding the usage of obsolete functions
  - ... avoid setting `InnerHandler`, because this will be done by the builder itself

# Future changes (not scope of this PR)

- `AuthenticatedHttpClientHandler`: Remove the old constructor 
- `AuthenticatedHttpClientHandler`: Remove the `innerHandler` parameter for the new constructor
- `AuthenticatedHttpClientHandler`: Fix the call sites
- Don't call the obsoleted `ConfigureHttpMessageHandlerBuilder`
- Fix the unit tests

# Behavioral changes

When no `httpClientName` was specified, the `UniqueName.ForType<T>` is used and the string representation of the `serviceKey` (when not null or empty string) gets appended to the unique type name, e.g. `Refit.Implementation.Generated+IFooWithOtherAttribute, Refit.Tests, Version=8.0.0.0, Culture=neutral, PublicKeyToken=null, ServiceKey=keyed`

### DIFF
--- a/Refit.HttpClientFactory/HttpClientFactoryExtensions.cs
+++ b/Refit.HttpClientFactory/HttpClientFactoryExtensions.cs
@@ -154,8 +154,13 @@ namespace Refit
                     )
             );
 
+            httpClientName ??=
+                UniqueName.ForType<T>()
+                + (serviceKey is null or ""
+                    ? string.Empty
+                    : $";{serviceKey}");
             return services
-                .AddHttpClient(httpClientName ?? UniqueName.ForType<T>())
+                .AddHttpClient(httpClientName)
                 .ConfigurePrimaryHttpMessageHandler(serviceProvider =>
                 {
                     var settings = serviceProvider.GetRequiredKeyedService<SettingsFor<T>>(serviceKey).Settings;
--- a/Refit.HttpClientFactory/HttpClientFactoryExtensions.cs
+++ b/Refit.HttpClientFactory/HttpClientFactoryExtensions.cs
@@ -154,13 +154,8 @@ namespace Refit
                     )
             );
 
-            httpClientName ??=
-                UniqueName.ForType<T>()
-                + (serviceKey is null or ""
-                    ? string.Empty
-                    : $";{serviceKey}");
             return services
-                .AddHttpClient(httpClientName)
+                .AddHttpClient(httpClientName ?? UniqueName.ForType<T>(serviceKey))
                 .ConfigurePrimaryHttpMessageHandler(serviceProvider =>
                 {
                     var settings = serviceProvider.GetRequiredKeyedService<SettingsFor<T>>(serviceKey).Settings;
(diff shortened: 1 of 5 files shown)


ChrisPulman:
Hi Mark, thank you for this, please could you make your next contribution to provide suitable tests to ensure this change doesn't become affected by future additions and alterations within the code base.
I look forward to seeing your list of future changes completed.

ChrisPulman (approved):

Issue #2016, opened 2025-10-06
[Bug]: Multipart requests is not working correctly with Guid and DateOnly


arthurlomakin11:
### Describe the bug 🐞

Using Refit for integration testing for our internal API. When working with it, discovered a problem with Guid processing in request. Guid is added with `"..."` in it that causes incorrect processing and gives BadRequest in test:

<img width="973" height="527" alt="Image" src="https://github.com/user-attachments/assets/1f9bb404-41fb-471f-9ae4-d14db5241328" />

### Step to reproduce

1. Go to '[Reproduction repository](https://github.com/arthurlomakin11/RefitInvalidMultipartGuidTest)'
2. Run test
3. I've also added `RequestMiddleware` and `LoggingHandler` for easier investigation in debug mode

### Reproduction repository

https://github.com/arthurlomakin11/RefitInvalidMultipartGuidTest

### Expected behavior

Here's the correct request(used Swagger from '[Reproduction repository](https://github.com/arthurlomakin11/RefitInvalidMultipartGuidTest)'):

<img width="1055" height="873" alt="Image" src="https://github.com/user-attachments/assets/2114190c-86ad-478a-ae96-638dada68768" />

<img width="1286" height="915" alt="Image" src="https://github.com/user-attachments/assets/be8b0cb9-b5d5-4d5e-8878-0da923550b3d" />

As you can see, Guid is shown in plain form, without `"..."`

### IDE

Rider Windows

### Refit Version

8.0.0

arturlomakinsparky:
Found the core problem itself and the way to temporary fix it.
When multipart item is processed, it's run through [`AddMultipartItem` method](https://github.com/reactiveui/refit/blob/a6a1304172ad02ff8745fef0b391c25a1902f810/Refit/RequestBuilderImplementation.cs#L265)
There it matches types to it's HttpContent and if not processed by regular conditions, fallbacks to serializer.

<img width="887" height="822" alt="Image" src="https://github.com/user-attachments/assets/57320154-acac-4f5e-b64e-a08b674343e4" />

So, I tried to debug library code of this method in [Reproduction repository](https://github.com/arthurlomakin11/RefitInvalidMultipartGuidTest) and got the expected result - Guid is processed through JSON serializer and it serializes `754c7d0c-41bf-406b-bed6-1f35e0bfd895` to `"754c7d0c-41bf-406b-bed6-1f35e0bfd895"` with quotes.

[...]

PR #2026, opened 2025-11-19
Quote types in README.md

TimothyMakkison:
See also misleading description for `MultiPart` parameters #2016


TimothyMakkison:
@ChrisPulman any thoughts on #2016?

Issue #1949, opened 2025-01-20
[Bug]:


snake-scaly:
### Describe the bug 🐞

The PR #1879 breaks our old code. Our API returns a generic version of `IApiResponse<T>`, but there are some implementation methods which accept the base `IApiResponse`. This works fine in 7.2.22. But in 8.0.0 `IApiResponse<T>.Error` and `IApiResponse.Error` are different fields, so in this case `IApiResponse.Error` is always `null` no matter what.

### Step to reproduce

In pseudocode:

```csharp
interface Api
{
  [Get("/resource")]
  Task<IApiResponse<string>> Foo();
}

async Task Bar()
{
  var client = RestService.For<Api>("https://example.com");
  var response = await client.Foo();
  Verify(response);
}

void Verify(IApiResponse r)
{
  if (r.Error != null) throw r.Error; // never happens
}
```

### Expected behavior

Applications should be able to use the base class to access common response fields.

### Refit Version

8.0.0

sguryev:
Duplicate of https://github.com/reactiveui/refit/issues/1933

snake-scaly:
So my bug report is incorrect. The production code will still work, it's just the mocking that's the problem. Good to know. Still, introducing a breaking change (even if just in tests) for slightly nicer IDE reporting seems excessive.

sguryev:
Yep. It's a "kind of" breaking change. We need to find a tradeoff between mocking correct interfaces and using NRT annotations

snake-scaly:
Closing as a duplicate, because #1933 is a much better description of the problem.

PR #2155, opened 2026-06-22
break: request-building fixes, JSON Lines, and response ergonomics

### LINKED ISSUE #1108: Is it possible to pull EnsureSuccessStatusCodeAsync into the IApiResponse<T> interface?
Before I create a PR I'd like to know if there's any objection to doing this that I might overlook. I just switched to 6.x and ran into this issue.

For unit testing scenarios it would be much easier if my methods exposed `IApiResponse<T>` instead of `ApiResponse<T>`. It would make mocking the response a lot easier.

However, at quite a few locations I call `EnsureSuccessStatusCodeAsync` on the return value of API calls. This only exists on `ApiResponse<T>`, meaning it's impossible to switch all my methods to `IApiResponse<T>`.

Pull `EnsureSuccessStatusCodeAsync` into the `IApiResponse<T>` interface.

Leave as is, making my unit tests a little more convoluted because creating an `ApiResponse<R>` there now also requires a `RefitSettings`.

PR to pull method into interface plus maybe a couple tests (haven't looked into that).

### LINKED ISSUE #1281: [BUG] query string parameter is not using parameter name from method signature
I have an API definition as follows:
```csharp
public interface IApi
{
    [Get("/info")]
    Task<string> GetInfo(string name, [Query(Format = "yyyy-M-d")] DateTime date, [Query(Format = "")] Size carSize);
}
```
where `Size` is a simplified implementation of enumeration pattern like below:
```csharp
public class Size
{
    public static Size Small = new Size("small");
    public static Size Medium = new Size("medium");
    public static Size Large = new Size("large");

    public string Key { get;  }

    private Size(string key)
    {
        Key = key;
    }

    public override string ToString()
    {
        return Key;
    }
}
```

When I call the api
```csharp
var response = await api.GetInfo("John Doe", DateTime.Now, Size.Medium);
```
the resulting URL is `/info?name=John%20Doe&date=2021-12-7&Key=medium` - note that `carSize` parameter is in query string under name `Key`

**Expected behavior**
the resuting URL shold be `/info?name=John%20Doe&date=2021-12-7&carSize=medium` as the query parameter names should respect parameter names from method signature

**Environment**
- OS:       Ubuntu 20.04 and Windows 10
- Device:   laptop
- Version:   6.1.15
- Working Version:  unknown

**Additional context**
The `[Query(Format = "")]` in front of `carSize` parameter was inspired by #1258 in attempt to get the expected URL, but even if not present the resulting URL is the same. The only thing that influenced the URL was making the `Key` property of `Size` class private - in that case the parameter was not present in URL at all (which suggests that serialization of parameter into URL does not use its `ToString()` method). I believe that sufficient fix would be that if format is specified as empty that the complex object should be serialized by just calling its `ToString()`method. That would make it aligned with behavior in #1258 and also kind of what user expects when explicitly specifying empty format. In such case also the name of serialized parameter should be just a name of parameter in method signature.

### LINKED ISSUE #1334: [BUG] Query delimiter is not working in property
**Describe the bug**

If you set Query delimiter on property it will not work

```c#
public class Params
{
    [Query("")]
    public string q { get; set; }
}
```

After entering the source code, I found that it only looks for the outermost delimiter(That is to say, `AttributeTargets.Parameter` level)

https://github.com/reactiveui/refit/blob/246ee8d9989c29092fdaba6821d5b3098a5ccf9e/Refit/RequestBuilderImplementation.cs#L654

https://github.com/reactiveui/refit/blob/246ee8d9989c29092fdaba6821d5b3098a5ccf9e/Refit/RequestBuilderImplementation.cs#L349

**Steps To Reproduce**

**Expected behavior**

**Screenshots**

**Environment**

- OS:       
- Device:   
- Version:   
- Working Version: 

**Additional context**

### LINKED ISSUE #1576: JSON Lines support

Some APIs provide batching via sending multiple records in the [JSON Lines](https://jsonlines.org/) format. It is a set of JSON documents with a line break as a separator.

An example of such API is the [Typesense API for importing multiple documents](https://typesense.org/docs/0.25.1/api/documents.html#index-multiple-documents):

```sh
curl "http://localhost:8108/collections/companies/documents/import?action=create" \
        -H "X-TYPESENSE-API-KEY: ${TYPESENSE_API_KEY}" \
        -H "Content-Type: text/plain" \
        -X POST \
        -d '{"id": "124","company_name": "Stark Industries","num_employees": 5215,"country": "USA"}
            {"id": "125","company_name": "Acme Corp","num_employees": 2133,"country": "CA"}'
```

Having JSON Lines support for the HTTP request body in Refit would be nice.

Marking a collection parameter in the method will do:

```c#
[Post("/collections/companies/documents/import")]
public Task InsertDocuments(string collectionName, [JsonLines] IEnumerable<Company> documents);
```

Completely replacing the JSON formatter. However, it is quite tedious.

**Additional context**

### LINKED ISSUE #1933: [Bug]: new keyword for properties on generic IApiResponse
### Describe the bug 🐞

Version 8 introduced a change in  #1879 where the `IApiResponse<out T>` interface applies the `new` keyword on certain properties like `StatusCode` and `ContentHeaders`.

If you have a method that uses the non-generic `IApiResponse` as parameter, the base interface properties will be used which will then not be set. This is especially difficult when mocking these interfaces where we now have to setup both the properties of the generic and non-generic interface.

This is a breaking change which is not mentioned in the release notes of release 8.0.0. This works perfectly in version 7.2.2

### Step to reproduce

Have a method that uses the non-generic `IApiResponse`:

```csharp
public static bool IsProblemDetails(this IApiResponse apiResponse)
{
    Guard.Against.Null(apiResponse);

    var contentType = apiResponse.ContentHeaders?.ContentType?.MediaType;

    return
        apiResponse.StatusCode == HttpStatusCode.UnprocessableEntity &&
        contentType?.Equals(MediaTypeNames.Application.ProblemJson) == true;
}
```

Create a unit test that mocks the generic `IApiResponse<out T>` interface and calls the method:

```csharp
[Fact]
public void IsProblemDetails_should_return_true_for_unprocessable_entity_with_content_type_problem_details()
{
    // Arrange
    var headers = new StringContent(string.Empty).Headers;
    headers.ContentType = new MediaTypeHeaderValue(MediaTypeNames.Application.ProblemJson);
    var mock = new Mock<IApiResponse<string>>();
    mock.SetupGet(x => x.IsSuccessStatusCode).Returns(false);
    mock.SetupGet(x => x.StatusCode).Returns(HttpStatusCode.UnprocessableEntity);
    mock.SetupGet(x => x.ContentHeaders).Returns(headers);
    var apiResponse = mock.Object;

    // Act
    var result = apiResponse.IsProblemDetails();

    // Assert
    Assert.True(result);
}
```
This results in `null` values for the non-generic interface, making the test fail:
image

### Reproduction repository

_No response_

### Expected behavior

Setting properties on the generic interface should result in being able to use it as the non-generic interface (not breaking inheritance)

### Screenshots 🖼️

_No response_

### IDE

_No response_

### Operating system

_No response_

### Version

.NET >6

### Device

_No response_

### Refit Version

8.0.0

### Additional information ℹ️

_No response_

### LINKED ISSUE #2016: [Bug]: Multipart requests is not working correctly with Guid and DateOnly

### LINKED ISSUE #2147: Feature Request: Auto-Disposed ApiResponeSnapshot<T> Implementation

Recently discovered that instances of `ApiResponse<T>` are, in fact, disposable, and should be treated as such. This would require a scouring of our application to use the proper disposal patterns. (In our case, we *never* do anything with the HTTP response once it's complete; we just need the content / status / result of the request. So, we'd immediately dispose every single time.)

e.g. transform every case of 

```csharp
ApiRespone<MyObject> foo = myApi.GetFooAsync(ct);
```

into something like

```csharp
using ApiRespone<MyObject> foo = myApi.GetFooAsync(ct);
```

This isn't causing major issues or fires. But it does mean that we're never disposing of HTTP Responses, which could lead to the usual issues surrounding that (leaving sockets open, connection pool exhaustion, etc.)

Rather than relying on consumers of the Refit APIs in code to have to dispose everything, being able to instruct Refit to do it for us (with the understanding that the underlying HTTP Response is already disposed) so that we don't have to do it ourselves would be ideal:

```csharp
public interface IMyRefitApi
{
  // ApiResponseSnapshot<T> takes a snapshot of all the HTTP response data/content/status/error/etc. and immediately disposes of resources.
  [Get("/path/to/some/api")]
  Task<ApiResponseSnapshot<MyObject>> GetFooAsync(CancellationToken ct);
}
```

A workaround would be to make my own `ApiResponseSnapshot<T>` that takes in `ApiResponse<T>`, grabs what I want, and dispose of the response immediately. But this would still require modifying every use of Refit (and we'd have to remember to do this going forward -- which inevitably we'll forget, short of having some static analysis / enforcement in place):

```csharp
// ToSnapshot() would be an extension method to do the transformation / disposal automatically.
ApiResponeSnapshot<MyObject> foo = myApi.GetFooAsync(ct).ToSnapshot();
```

We can't just put it on the Refit API interface directly, because anything that isn't `IApiResponse` or `ApiResponse<T>` is assumed to be a custom object to deserialize the Response Content. Even implementing `IApiResponse` is not treated as "hydrate this with the HTTP Response" and instead is treated like any other custom object.

Whatever's done in Refit to see `ApiRespone<T>` and hydrate it with the HTTP Response content, do the same with a hypothetical `ApiResponseSnapshot<T>` and immediately dispose of the underlying HTTP Response. It could even be an `ApiResponse<T>` (although that class is sealed), and just `try/finally` in constructor to dispose the underlying response immediately, potentially.

The cleaner way would be to implement `IApiResponse` and earmark `ApiResponseSnapshot<T>` as another special case for "no this isn't a custom object: Give me the HTTP Response."

I'm not super familiar with the underlying source though.

**Alternate Possibility (Requires Users of this NuGet to Implement)**

Some means of making it so we can provide our own custom `ApiResponse<T>`'s, whether this means unsealing `ApiResponse<T>` or making implementations of `IApiResponse` behave like `ApiResponse<T>` does, would let us implement our own solutions. Right now this doesn't seem possible, due to any not-`(I)ApiResponse` object being assumed to be content to deserialize to.

**Additional context**

I don't know how prevalent our use case is. But across a couple applications now, my usage of Refit APIs is generally "we don't care about the response once we have the data, and it would be nice if we didn't have to manage the disposable nature of it -- just immediately dispose for us please and thank you."

Thanks!

glennawatson:

- Bug fixes, features, docs, and a generator refactor. One small, documented breaking change to `IApiResponse<T>`.

## What is the new behavior?

- **fix(#2016):** Multipart `Guid`/`DateTime`/`DateTimeOffset`/`TimeSpan` (and `DateOnly`/`TimeOnly` on net6+) are sent as plain text via the form parameter formatter instead of JSON-quoted. Numbers, booleans and enums intentionally still serialize as before.
- **fix(#1334):** A property-level `[Query(delimiter, prefix)]` is now honored when flattening a complex object into the query string.
- **fix(#1281):** A query parameter marked `[Query(Format = "")]` serializes a complex value via `ToString()` under the parameter name instead of flattening its public properties (opt-in via the explicit empty format).
- **fix(#1933):** `IApiResponse<T>` no longer shadows any base member with `new`. The generic interface now adds only `Content` plus a new covariance-safe `HasContent` member for `Content` flow-narrowing. A single setup on the generic interface is now fully observed through the non-generic `IApiResponse` (mocking works).
- **feat(#1108):** `EnsureSuccessStatusCodeAsync` / `EnsureSuccessfulAsync` are now reachable from `IApiResponse<T>` via extension members - no need for the concrete `ApiResponse<T>` type.
- **feat(#1576):** JSON Lines (newline-delimited JSON) request bodies via `[Body(BodySerializationMethod.JsonLines)]` and a streaming `JsonLinesContent`, wired through the reflection and source-gen request paths (`application/x-ndjson`).
- **docs(#2147):** README now documents response disposal semantics - non-streaming responses are already disposed by Refit (no leak; `using` is a harmless no-op), and only `HttpResponseMessage`/`HttpContent`/`Stream` (and their `ApiResponse<>` wrappers) require caller disposal. Also documents JSON Lines, the naming-convention presets, the new query-key prefix/delimiter and `ToString()` formatting, multipart scalar values, `HasContent`, and the `EnsureSuccess*` helpers.
- **refactor:** Split `Emitter.cs` into `Emitter.Inline.cs` and `Emitter.Constraints.cs`.

- Multipart Guid/date values are JSON-quoted and rejected by form processors.
- Property-level query delimiters/prefixes are ignored; complex query params are always flattened to their inner property names.
- A mocked `IApiResponse<T>` cast to `IApiResponse` returns null for the `new`-shadowed members.
- The success guards exist only on the concrete `ApiResponse<T>`.
- No first-class JSON Lines body support; disposal semantics were undocumented.

Closes #2016
Closes #1334
Closes #1281
Closes #1933
Closes #1108
Closes #1576
Closes #2147

## Breaking changes

**`IApiResponse<T>` drops its `new`-shadowed members** (`Error`, `ContentHeaders`, `IsSuccessStatusCode`, `IsSuccessful`). This is what actually fixes #1933.

- **Source:** unchanged - `response.Error`, `.ContentHeaders`, `.IsSuccessStatusCode`, `.IsSuccessful` on an `IApiResponse<T>` still compile and work (they bind to the inherited base members).
- **Binary:** code *compiled* against v8 to v11 that bound to the generic-interface slots (e.g. `IApiResponse<T>.get_ContentHeaders`) needs a recompile.
- **Nullable analysis:** on an `IApiResponse<T>`-typed value, `if (response.IsSuccessful)` no longer narrows `Content` to non-null. Use the new **`HasContent`** member instead: `if (response.HasContent) { /* response.Content is non-null */ }`. (On the concrete `ApiResponse<T>`, `IsSuccessful` still narrows `Content`.)

## Helpers for the v11.0.0 breaking workflows

- **`Error` is now `ApiExceptionBase?`** (no `Content`): the response-side `ApiException` is reachable via the existing `response.HasResponseError(out var apiEx)` then `apiEx.Content` (and `HasRequestError(out var reqEx)` for transport failures) - now documented.
- **Success guards** previously required the concrete type - now on the interface via `EnsureSuccessStatusCodeAsync()` / `EnsureSuccessfulAsync()` (#1108).
- **`Content` null-checks with flow analysis:** the new `HasContent` member replaces the removed `IsSuccessful`-based narrowing on the interface.

## What might this PR break?

- Only the `IApiResponse<T>` shadow removal above (source-compatible, recompile for binary, `HasContent` for narrowing). Everything else is additive or opt-in.

## Checklist

## Additional information

- 100% patch coverage (runtime + generator); generator branches covered via direct `...ForTesting` unit tests.
- Public API moved from unshipped to shipped across all TFMs.
- Full solution builds clean on all TFMs (net462 to net11); Refit.Tests (697) and GeneratorTests (136) pass.


ChrisPulman (approved):

sguryev:
@glennawatson am I right from now there is no way to make success+content check with one condition and we should use something like `if (!response.IsSuccessful || !response.HasContent)` to calm analyzer down?

So it's similar to what I had before (except "a kind of useless" property):
image

Basically nobody cares about null in the `Content` property except explicit check `Content is null`.

Also I have noticed that lists of analyzer suppressions are suspicious identical for these two different properties
image

glennawatson:
Thanks @sguryev - good feedback, and you're right to call it a compromise.

On your code: nothing here breaks it. Your `if (!response.IsSuccessStatusCode || response.Content == null)` pattern (the screenshot) compiles and behaves exactly as before - `HasContent` is purely opt-in and you can ignore it if you always null-check `Content` explicitly.

On the missing single-condition success+content narrowing: that is the part that genuinely can't come back. `IsSuccessful` only narrowed `Content` before because it was redeclared with `new` on the generic interface to carry `[MemberNotNullWhen(true, nameof(Content))]`, and that `new` shadow is exactly what broke a single mock setup on `IApiResponse<T>` from being seen through the non-generic `IApiResponse` (#1933). Narrowing requires the shadow; the shadow re-introduces the bug - so you can have the mocking fix or the one-condition narrowing, not both. `HasContent` is the only way to keep compiler narrowing on `Content` from the interface without the shadow: `if (response.HasContent) { /* Content is non-null */ }` is a single condition if you ever want it. For your explicit-null-check style it adds nothing, which is fair.

On the identical `[MemberNotNullWhen]` lists for `IsSuccessStatusCode` and `IsSuccessful`: that's pre-existing on the base interface (this PR only touched the generic one), and it's actually correct - both states imply a response was received, so they narrow the same metadata. But you're right that copy-pasted annotations deserve a correctness pass (e.g. 204 No Content), so I've opened #2156 to audit them.

tl;dr: compromise was deliberate - a full fix that restores the one-condition narrowing would re-break mocking. Your code is unaffected.

sguryev:
@glennawatson thank you for the response.

Screen shows my OLD code. I had to make double checking every single time to narrow analyzer warnings. I have copy-pasted if from the bug mentioned above. I have refactored all places to use `IsSuccessful` prop because it has (had) ` [MemberNotNullWhen(true, nameof(Content))]` attr.

Description clearly says that it means content was deserialized with no errors:
image
And identical list is an evidence about ambiguous properties here now. 
I'm curious why it requires to check Content for null after it. 

I understand all the tech details (because I'm the guy who introduced `new` properties). I just can't believe wrong interface usages has beaten property meaning.

glennawatson:
Thanks @sguryev, and genuinely - I appreciate you pushing on this, especially as the author of the original `new` properties. Fair questions, so let me give you a proper answer on two fronts.

First, on mocking: I'd gently push back on framing it as wrong usage. Mocking `IApiResponse<T>` in tests is completely normal, first-class usage - test projects are part of everyday development, and a public interface is a contract that should behave the same however it is reached. With the `new` shadow, a single setup on `IApiResponse<T>` silently was not observed through the non-generic `IApiResponse` (#1933). That is a footgun for ordinary test code, not someone holding it wrong - so supporting it is part of the property's meaning, not a defeat of it.

Second, and I think this is what actually answers your "why check Content for null after IsSuccessful" question: the old annotation was not quite correct either. The doc says "no error during deserialization" - but "no error" is not the same as "non-null content". A successful response can deserialize cleanly to null: a 200 or 204 with an empty or literal `null` JSON body returns `Content == null` with no error, so `IsSuccessful == true` while `Content` is null. The old `[MemberNotNullWhen(true, nameof(Content))]` told the compiler Content was non-null in exactly those cases, which is an over-promise that can NRE on a perfectly valid empty response. So the null-check after `IsSuccessful` is not ceremony - it is correct, because success genuinely does not imply content.

That is really what `HasContent` is: an honest "Content is non-null" signal you can use in a single condition - `if (response.HasContent) { /* Content non-null */ }` - without claiming something success cannot guarantee. So the new shape is both more mockable and arguably more correct; it did not trade correctness for pragmatism, the old narrowing was just optimistic.

Your identical-attributes observation feeds straight into this, so I have folded the "IsSuccessful does not imply non-null Content" point into #2156 alongside the 204 / no-body audit - the docs and annotations can be clearer about it. Really do appreciate the back-and-forth.

sguryev:
@glennawatson 

> That is a footgun for ordinary test code, not someone holding it wrong - so supporting it is part of the property's meaning, not a defeat of it.

It's a tradeoff between the mocking approach with Moq library and expectable properties NRT null states. #1933 has Sharplab sample which proves that `IApiResponse<T>` properties are observed through the non-generic `IApiResponse`.

image

I would disagree with you on this
> A successful response can deserialize cleanly to null: a 200 or 204 with an empty or literal null JSON body returns Content == null with no error, so IsSuccessful == true while Content is null.

200 with empty body is a bad habit and should be handled by `IApiResponse` not the generic one. I can't believe somebody can write the code which handles all: 
- empty body, 
- `{}` 
- filled body. 
But it can be debatable at least.

204 with no body by design should not be handled by `IApiResponse<T>` by design.
I would say we should teach users to use interfaces correctly rather than trying to handle all their mistakes. 

>  So the null-check after IsSuccessful is not ceremony - it is correct, because success genuinely does not imply content.

In this case it looks like `IsSuccessful / IsSuccessStatusCode` duplicates each other.

> That is really what HasContent is: an honest "Content is non-null" signal you can use in a single condition - if (response.HasContent) { /* Content non-null */ } - without claiming something success cannot guarantee. So the new shape is both more mockable and arguably more correct; it did not trade correctness for pragmatism, the old narrowing was just optimistic.

Can't imagine the case where I need to use `HasContent` instead of checking `Content` myself.

I'm really keen to have some property allowing a single-shot analyzer warning narrowing.

glennawatson:
Your SharpLab shows the one case that was never broken. `public int MyProperty => 45;` is a single implicit implementation, so one method satisfies both interface slots - that is exactly the concrete `ApiResponse<T>`.

`new` creates two distinct slots in metadata, and a mock/proxy fills them independently. Here is the IL for `interface IBase { int P; } interface IDerived : IBase { new int P; }` plus a class that implements them separately (which is what Castle/Moq generate):

```
.method ... newslot abstract virtual ... int32 IBase::get_P()       // slot #1
.method ... newslot abstract virtual ... int32 IDerived::get_P()    // slot #2  (the new)

.method ... int32 ProxyLike::IBase.get_P()    { ldc.i4.0    ret }   // 0
.method ... int32 ProxyLike::IDerived.get_P() { ldc.i4.s 45 ret }   // 45
```

```csharp
IDerived d = new ProxyLike();
d.P;            // 45
((IBase)d).P;   // 0   <-- same object, two slots
```

And the literal Moq form, which prints 45 then 0 (ran against Moq 4.20.72):

```csharp
var m = new Mock<IDerived>();
m.SetupGet(x => x.P).Returns(45);  // binds to IDerived.P only
m.Object.P;                         // 45
((IBase)m.Object).P;               // 0   (IBase.P slot never configured)  -> #1933
```

So it is a property of `new`, not a Moq quirk: it splits the member into two slots, and the observed value depends on how the type is implemented rather than on its state. Your sample hid that because it has one implicit implementation.

On the narrowing: a 200 with a literal `null` JSON body deserializes to `Content == null` with no error (verified - an empty body throws, but `null` does not), so `IsSuccessful == true` while `Content == null`. `[MemberNotNullWhen(true, nameof(Content))]` is an unconditional promise, so it was unsound there, not just inconvenient.

That said, the underlying ask is reasonable and we can give it to you without the shadow. A new, mock-safe member:

```csharp
bool IsSuccessfulWithContent { get; }   // => IsSuccessful && Content is not null
```

`if (response.IsSuccessfulWithContent) { /* Content is non-null */ }` is one check, narrows `Content`, keeps success semantics, and is sound. Because it is a new member (no `new` shadow) it has no base twin to diverge from - verified it compiles on a covariant `IApiResponse<out T>`, narrows under `Nullable=enable`, and mocks correctly. Would that resolve it for you? `HasContent` stays for pure content-presence; the annotation/duplication cleanup is tracked in #2156.

glennawatson:
I did some digging - the "a successful response must always have a body" assumption isn't true, and it doesn't hold up against RFC 9110. Hopefully the below clears up why, and why the compromise member is the better way to land this.

RFC 9110, Section 6.4.1 (Content Semantics):
> All 1xx (Informational), 204 (No Content), and 304 (Not Modified) responses do not include content.
> All other responses do include content, although that content might be of zero length.

And a 204 is a success - Section 15.3.5 (204 No Content):
> The 204 (No Content) status code indicates that the server has successfully fulfilled the request and that there is no additional content to send in the response content.

So a 2xx can legitimately come back with no body (a 204, a HEAD response, or any status with zero-length content). A successful status can't guarantee a non-null deserialized body, so an unconditional `[MemberNotNullWhen(true, nameof(Content))]` ends up asserting something HTTP doesn't - and it's already tripped a couple of users (#1933, #1949).

That's why I'd lean to keeping the body-presence signal explicit and separate from success - `HasContent` today, plus `IsSuccessfulWithContent` for the single-check narrowing you're after. You get the one-condition narrowing soundly, without promising content the spec says might not be there.

Ref: RFC 9110, https://www.rfc-editor.org/rfc/rfc9110.html

<blockquote><div><strong><a href="https://www.rfc-editor.org/rfc/rfc9110.html">RFC 9110: HTTP Semantics</a></strong></div><div>
       
   The Hypertext Transfer Protocol (HTTP) is a stateless application-level
   protocol for distributed, collaborative, hypertext information systems.
   This document describes the overall architecture of HTTP, establishes common
   terminology, and defines aspects of the protocol that are shared by all
   versions. In this definition are core protocol elements, extensibility
   mechanisms, and the "http" and "https" Uniform Resource Identifier (URI)
   schemes.
       
       
   This document updates RFC 3864 and
   obsoletes RFCs 2818, 7231, 7232, 7233,
   7235, 7538, 7615, 7694, and portions of 7230.
 
    </div></blockquote>

sguryev:
> I did some digging - the "a successful response must always have a body" assumption isn't true, and it doesn't hold up against RFC 9110. Hopefully the below clears up why, and why the compromise member is the better way to land this.

I have never said that "a successful response must always have a body". Sorry. I said that can't imagine API scheme returning 200 with body and without it at the same time (especially null instead of the zero length body). And I would prefer to have 204 as a separate response code for such cases. That's why I would use `IApiResponse` for cases with no content\zero length content\null body\ and `IApiResponse<T>` for `{}` and filled body cases. But it's debatable I agree.

Issue #1358, opened 2022-05-11
Best practice for base address when HttpClient is provided


antoinebj:
I have a project where my HTTP clients are configured through the mechanisms provided by HttpClientFactory. They get policies and handlers assigned to them and I want to use that configuration everywhere.
For most endpoints, I use Refit which gets its `HttpClient` supplied by the factory, but for some complex scenarios, I have to use the HttpClient directly without Refit.

My base address is in the form of `http://example.com/api/`. The trailing slash is required for using `HttpClient` directly, otherwise it will ignore the `api` segment when forming the complete URL.

Refit seems to not play nice with that. Instead of working in harmony with that behavior, it tries to work around it. I can see in `BuildVoidTaskFuncForMethod` that it takes the base path from the `HttpClient`'s base address in order to prepend it itself to the "relative path" in `RequestBuilderImplementation.BuildRequestFactoryForMethod`. In that prepending, it expects that the base path not end with a slash, and has logic to remove the slash in the case that the base path is the root:
```
var urlTarget = (basePath == "/" ? string.Empty : basePath) + restMethod.RelativePath;
```

That "relative path" in `RestMethodInfo.RelativePath` is enforced to start with a slash, despite the fact that by most if not all familiar conventions, a path that starts with a slash is supposed to be absolute.

Now if Refit creates the `HttpClient` itself, then it will remove the trailing slash from the base address that it supplies to the `HttpClient`, so that when it is used, it will forego the base path which has been "manually" prepended to the relative path by Refit.
However if the `HttpClient` was supplied to Refit, then it has no control over the base address, and if that base address has a trailing slash, then there will be a double slash in the URL: `http://example.com/api//my-resource`

1. What is the best way of reconciling the direct use of `HttpClient` with using Refit on top, when the base address has a path after the domain?

2. Why wasn't Refit designed to take actual relative paths (not starting with a slash) for the endpoint paths? Those slashes seem to be the source of the conflict in behavior with plain `HttpClient`. Would it be possible to add a configuration option to switch that behavior?

nazarkryp:
I'm also facing this problem. I have different clients where BaseAddresses would be like which target different microservices: 
https://example.com/payments/
https://example.com/orders/
https://example.com/shopping-carts/

And I'd like to specify relative path without leading / , eg:
[Post("v2/payments")]
Task<CreatePaymentResponse> CreatePaymentAsync(CreatePaymentRequest request);

[...]

PR #2157, opened 2026-06-22
break: add IAsyncEnumerable streaming and opt-in URL, route and JSON serialization modes

### LINKED ISSUE #1299: feature: Support IAsyncEnumerable as a return type
It would be nice if Refit would support streaming deserialization of an API response. This could be useful for endpoints that return a large amount of data, since it could avoid buffering the whole API response in memory. It can also allow the consumer to start enumerating the items before the whole API response has been received.

Refit is now using `System.Text.Json`, and with .NET 6, [System.Text.Json got support for streaming deserialization](https://devblogs.microsoft.com/dotnet/announcing-net-6-preview-4/#system-text-json-support-for-iasyncenumerable). Using it could be as easy as having an `IAsyncEnumerable` as the return type in an interface:
```csharp
public interface IApi
{
    [Get("/items")]
    IAsyncEnumerable<Item> GetItems();
}
```

Using `Task<IEnumerable<Item>>` as the return type, but this will buffer the whole API response in memory.

### LINKED ISSUE #1803: [Bug]: BaseAddress / relative URL formats in documentation
### Describe the bug 🐞

The `HttpClient`'s `BaseAddress and relative URLs in the README.md are not written according to the way how HttpClient is meant to be used:
- the `BaseAddress` is expected to end with a `/`
- relative URLs are expected to start without a `/`

Sources:
- https://developercommunity.visualstudio.com/t/httpclient-not-smart-about-combining-baseaddress-w/1592519
- https://stackoverflow.com/questions/23438416/why-is-httpclient-baseaddress-not-working

However, the samples show things like `c.BaseAddress = new Uri("https://api.github.com")` (missing '/') and `Get("/users/{user}")` (starting `/`):
https://github.com/reactiveui/refit/blob/f77f0f39a89323a1566182e2135a0d30382a22a0/README.md?plain=1#L15-L35

### Reproduction repository

### Expected behavior

See https://github.com/reactiveui/refit/blob/main/README.md

### Screenshots 🖼️

_No response_

### IDE

_No response_

### Operating system

_No response_

### Version

_No response_

### Device

_No response_

### Refit Version

_No response_

### Additional information ℹ️

_No response_

### LINKED ISSUE #1968: Add option to not throw exception when route parameter does not have a matching argument.
There might be another way to solve my problem, so if there is please advise. Currently I have a Blazor application with a WASM frontend and a server hosting with my API endpoints on the backend. Seeing as these are in the same VS solution, I have created a Endpoints class to maintain all of my base routes.  
For example:
```csharp
public static class Endpoints 
{
    public const string Route1 = "/api/v1/MyRoute";
    public const string Route2 = "/api/v1/MyOtherRoute";
}
```

These base endpoints are then used as the base to defining my routes within Refit and the endpoints on the server. However, I would like to use Api Versioning which as far as I can tell, the route needs to have `v1` replaced with `v{version:apiVersion}`. While this works for the serve routes, Refit complains about not having a matching parameter in its argument list. 

If my approach is incorrect let me know. If this approach is valid, then it would be nice to be able to specify a setting in Refit that prevents the runtime from throwing when a parameter does not match in the argument list. This way I can replace the unused parameter myself inside a delegating handler. 

Alternatively, it would also work if I could specify a parameter in the route, and have it replaced by an attribute, as opposed to an argument. 

I think the easiest solution would be to add a setting to the RefitSettings, that prevents an exception from being thrown when not all parameters are matched.

### LINKED ISSUE #2022: Default Json options: NumberHandling
Most users would expect the default JSON serialization to match ASP.NET default serialization.

I suggest that `GetDefaultJsonSerializerOptions()` here: https://github.com/reactiveui/refit/blob/main/Refit/SystemTextJsonContentSerializer.cs#L64-L78 should be based on `JsonSerializerOptions.Web` in .NET 9+, or equivalent options in lower runtimes.

The key difference (besides the converters added by Refit) is `NumberHandling`.
Default web serializer sets `NumberHandling = JsonNumberHandling.AllowReadingFromString`.

The consequence of this change is that deserialization into numbers would be possible from strings, which should not be a breaking change.
It's not uncommon to serialize numbers into strings when using numbers larger than doubles, or `decimal` types.

glennawatson:

Feature, several additive, opt-in capabilities, plus one breaking default-behaviour change to the System.Text.Json serializer.

## What is the new behavior?

- **`IAsyncEnumerable<T>` response streaming.** Declare a method returning `IAsyncEnumerable<T>` and Refit reads the response with `ResponseHeadersRead` and yields items as they arrive. The frame format is auto-detected from the content type: `application/jsonl` / `application/x-ndjson` / `application/x-jsonlines` is read as JSON Lines, anything else as a single streamed JSON array. Backed by the new `IStreamingContentSerializer` capability (implemented by `SystemTextJsonContentSerializer`).
- **Opt-in RFC 3986 / `HttpClient`-style URL resolution.** `RefitSettings.UrlResolution = UrlResolutionMode.Rfc3986` combines the base address and relative path with `System.Uri` rules (a leading slash replaces the base path; no leading slash appends) and relaxes the leading-slash requirement. Defaults to `RefitLegacy`, so existing behaviour is unchanged.
- **Opt-in `RefitSettings.AllowUnmatchedRouteParameters`.** When set, a route placeholder with no matching argument is left in the URL verbatim (e.g. `{version:apiVersion}`) for a `DelegatingHandler` to rewrite, instead of throwing.
- **Fast-path JSON serialization for request bodies.** `RefitSettings.RequestBodySerialization` (`Buffered` or `Streamed`) serializes JSON bodies synchronously so the System.Text.Json source-generated fast-path engages, `Buffered` sends a `ByteArrayContent` with `Content-Length`, `Streamed` writes through a `Utf8JsonWriter` for large uploads. `SystemTextJsonContentSerializer.GetFastPathJsonSerializerOptions()` returns a fast-path-eligible options set (no converters, no `NumberHandling`).
- **Numbers can be read from JSON strings by default** (`NumberHandling = AllowReadingFromString`), so responses like `{"id":"123"}` bind to numeric members.
- Tests, BenchmarkDotNet benchmarks (including EventPipe CPU/alloc profiles) and README sections were added for all of the above.

Previously Refit had no streaming response support, always required leading-slash relative paths with Refit-style base-address joining, always threw on an unmatched route placeholder, always serialized request bodies asynchronously (which can never use the source-gen fast-path), and used strict number handling on the default serializer.

Closes #1299
Closes #1968
Closes #1803 (and consolidates #1358)
Closes #2022

## What might this PR break?

The default `System.Text.Json` serializer now reads numbers from JSON strings (`NumberHandling = AllowReadingFromString`). Writing is unchanged. Opt back out with:

```csharp
var options = SystemTextJsonContentSerializer.GetDefaultJsonSerializerOptions();
options.NumberHandling = JsonNumberHandling.Strict;
var settings = new RefitSettings(new SystemTextJsonContentSerializer(options));
```

Everything else is additive and opt-in (defaults preserve current behaviour).

## Checklist

## Additional information

Why these approaches:

- **`IAsyncEnumerable<T>` is generated inline, with no reflection on the hot path.** The generator emits a direct `GeneratedRequestRunner.StreamAsync<T>` call with the request constructed in source (new `ReturnTypeInfo.AsyncEnumerable`), the same approach used for `Task<T>`. The reflection request-builder remains only as the fallback for shapes that cannot be generated inline (dynamic routes, body/query objects, generic methods). This keeps allocations and dispatch cost minimal for streaming, which is the whole point of the feature.
- **Streaming format is auto-detected** rather than configured, so it "just works" for both standard JSON arrays and JSON Lines APIs. Per-TFM, .NET 9+ uses `DeserializeAsyncEnumerable(topLevelValues: true)` for JSON Lines; older TFMs use a pooled UTF-8 line reader.
- **The fast-path is synchronous-only.** We verified empirically (including on .NET 9) that `SerializeAsync(Stream)` and `SerializeAsync(PipeWriter)` both use the metadata path; only the synchronous primitives (`SerializeToUtf8Bytes`, `Serialize(Utf8JsonWriter)`) hit the generated fast-path. Refit's default `JsonContent` uses `SerializeAsync(Stream)`, so it can never use the fast-path, hence the opt-in `Buffered`/`Streamed` modes and the converter-free `GetFastPathJsonSerializerOptions()` (Refit's default converters and `NumberHandling` also disable the fast-path). A benchmark shows the fast-path is roughly 1.8x to 2x faster than the default path.
- **A single `RequestBodySerializationMode` enum** rather than stacked booleans, because the three states (async default, sync buffered, sync streamed) are mutually exclusive.
- **RFC resolution and unmatched route parameters are opt-in enums/flags** so existing consumers are unaffected; the URL joining was centralized into one helper used by both the reflection and generated request paths.
- New public API is added to `PublicAPI.Shipped.txt` (the project ships continuously).

### DIFF
--- a/README.md
+++ b/README.md
@@ -235,6 +235,7 @@ construction for request shapes the generator can safely model. It currently cov
 * `[Property]` interface properties
 * cancellation tokens
 * `Task`, `Task<T>`, `Task<ApiResponse<T>>`, and related response wrappers
+* `IAsyncEnumerable<T>` for streamed responses
 
 If a method uses a shape the generator cannot safely emit yet, Refit falls back to the existing runtime request-builder
 path for that method.
@@ -343,6 +344,39 @@ Task<List<Page>> Search(string page);
 
 Search("admin/products");
 >>> "/search/admin/products"
+```
+
+By default Refit throws if a route template contains a placeholder with no matching method argument. If you want to
+resolve a placeholder later yourself (for example an API-versioning token rewritten inside a `DelegatingHandler`), set
+`AllowUnmatchedRouteParameters` on `RefitSettings`. The unmatched `{token}` is then left in the URL verbatim instead of
+throwing.
+
+```csharp
+var settings = new RefitSettings { AllowUnmatchedRouteParameters = true };
+var api = RestService.For<IVersionedApi>("https://api.example.com", settings);
+
+// [Get("/api/{version:apiVersion}/values")]
+// the {version:apiVersion} token is left in the path for a DelegatingHandler to replace.
+```
+
+#### Base address and URL resolution
+
+By default Refit requires relative paths to start with `/`, prepends the base address path itself, and trims a trailing
+slash from the base address. If you would rather have the base address and relative URL combined the same way
+`HttpClient` and `System.Uri` do (RFC 3986), set `UrlResolution` on `RefitSettings`:
+
+```csharp
+var settings = new RefitSettings { UrlResolution = UrlResolutionMode.Rfc3986 };
+var api = RestService.For<IMyApi>("https://api.example.com/api/v1/", settings);
+```
+
+Under `UrlResolutionMode.Rfc3986` the leading-slash requirement is relaxed and the trailing slash on the base address is
+significant, exactly as with `HttpClient`:
+
+```csharp
+// base address "https://api.example.com/api/v1/"
+[Get("values")]   // -> https://api.example.com/api/v1/values  (appended)
+[Get("/values")]  // -> https://api.example.com/values         (leading slash replaces the base path)
 ```
 
 ### Querystrings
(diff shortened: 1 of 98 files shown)


ChrisPulman (approved):

Trimatix:
Wow, huge PR! fantastic changes, thank you so much :)

PR #2159, opened 2026-06-23
feat: add IsSuccessfulWithContent and correct IApiResponse nullable annotations

### LINKED ISSUE #2156: Audit IApiResponse MemberNotNullWhen success-state annotations (e.g. 204 No Content)
## Overview

Tracking issue for the `IApiResponse` / `IApiResponse<T>` content-state ergonomics and annotation work that came out of the #2155 review discussion. Covers two related threads: a small additive member to restore single-check content narrowing, and a correctness pass over the existing nullable annotations.

## Background / verified facts

From #2155, fixing #1933 removed the `new`-shadowed members on `IApiResponse<T>`, including the `[MemberNotNullWhen(true, nameof(Content))]` that used to let `if (response.IsSuccessful)` narrow `Content`. Two things were established with runnable repros:

- The `new` shadow is genuinely unsound for mocks/proxies. `new` creates two distinct interface slots; a mock (Moq/Castle DynamicProxy) or any explicit implementation fills them independently, so setting up the generic property is not observed through the non-generic interface. Verified: real Moq 4.20.72 prints `45` via `IDerived` and `0` via `IBase` for the same object (the #1933 bug). A single implicit implementation (the concrete `ApiResponse<T>`) collapses both slots, which is why the concrete type always worked.
- The old `IsSuccessful -> Content` narrowing was itself an over-promise. `IsSuccessful` means "2xx and no deserialization error", but "no error" is not "non-null content": a 200 with a literal `null` JSON body deserializes to `Content == null` with no error (verified; an empty body throws, but `null` does not), and per HTTP a success can carry no body at all (see RFC grounding below).

## RFC 9110 grounding (success does not imply content)

A successful HTTP response is not required to carry content, so `IsSuccessStatusCode` / `IsSuccessful` being true cannot guarantee `Content != null`. An unconditional `[MemberNotNullWhen(true, nameof(Content))]` therefore asserts something the protocol does not.

RFC 9110, Section 6.4.1 (Content Semantics):
> All 1xx (Informational), 204 (No Content), and 304 (Not Modified) responses do not include content.
> All other responses do include content, although that content might be of zero length.

RFC 9110, Section 15.3.5 (204 No Content) - a success with no content by definition:
> The 204 (No Content) status code indicates that the server has successfully fulfilled the request and that there is no additional content to send in the response content.

So a 2xx can legitimately have no body (204, a response to HEAD which "never include[s] content", or any status with zero-length content). This already surfaced as user-reported breakages (#1933, #1949). Ref: https://www.rfc-editor.org/rfc/rfc9110.html

## Action items

1. Add a new, mock-safe member that restores single-check content narrowing without the shadow:
   ```csharp
   bool IsSuccessfulWithContent { get; }   // ApiResponse<T>: => IsSuccessful && Content is not null
   ```
   - One condition, narrows `Content`, keeps success semantics, and is sound (it actually checks content).
   - It is a brand-new member (not a `new` shadow) and references the generic-only `Content`, so it has no base-interface twin to diverge from - mocks behave correctly. Verified it compiles on a covariant `IApiResponse<out T>` and narrows under `Nullable=enable`.
   - `HasContent` stays for the pure content-presence case.
   - Public API: additive interface + concrete member; add per-TFM PublicAPI entries; add tests (narrowing, mock setup, empty/null-content).

2. Audit the existing `[MemberNotNullWhen]` success-state annotations on `IApiResponse`. `IsSuccessStatusCode` and `IsSuccessful` carry identical lists:
   ```
   [MemberNotNullWhen(true, nameof(Headers))]
   [MemberNotNullWhen(true, nameof(ContentHeaders))]
   [MemberNotNullWhen(true, nameof(StatusCode))]
   [MemberNotNullWhen(true, nameof(Version))]
   [MemberNotNullWhen(false, nameof(Error))]
   ```
   - Verify each holds when the property is true, including 204 No Content / no body (is `ContentHeaders` guaranteed non-null?) and a 2xx with a deserialization error.
   - Trim any annotation that is not actually guaranteed; otherwise leave as-is (both states imply a response was received, so the duplication is correct, not copy-paste).

3. Docs: clarify that `IsSuccessful` / `IsSuccessStatusCode` being true does not imply `Content` is non-null, and that `HasContent` / `IsSuccessfulWithContent` are the correct signals for content presence.

Priority: low. Area: exceptions (ApiResponse) / serialization. Source: #2155 review feedback.

glennawatson:

Feature + bug fix (nullable annotation correctness) + docs + tests.

## What is the new behavior?

At a high level: checking that a response "succeeded" no longer falsely promises that it has a body, and there is now a single, sound way to check for "succeeded **and** has content".

- Adds `IsSuccessfulWithContent` to `IApiResponse<T>` / `ApiResponse<T>`:
  - One condition (`IsSuccessful && Content is not null`) that narrows `Content` to non-null via `[MemberNotNullWhen]`, so `if (response.IsSuccessfulWithContent) { /* Content is non-null */ }` works in a single check.
  - It is a brand-new generic-only member (not a `new` shadow of a base member), so it behaves correctly through mocks, explicit interface implementations, and the covariant `IApiResponse<out T>`.
- Corrects the `[MemberNotNullWhen]` annotations on `IsSuccessStatusCode`, `IsSuccessful`, and `IsReceived`:
  - Removes the over-promising `Content` (true), `ContentHeaders` (true) and `Error` (false) narrowings, which the success/received state does not actually guarantee.
  - Keeps the sound `Headers`, `StatusCode`, and `Version` (true) narrowings.
  - The non-generic interface and the concrete implementation now carry identical annotation sets.
- `HasContent` remains the pure "is there a deserialized body?" signal; `IsSuccessfulWithContent` combines it with success.
- Documents on `Error` that a failed state does not imply a non-null error (null-check it, or use `HasRequestError` / `HasResponseError`).
- Updates the README to explain that success does not imply content, with `HasContent` / `IsSuccessfulWithContent` examples and an RFC 9110 note.

- There is no single-condition member that both confirms success and narrows `Content`; callers must write a two-part guard (e.g. `if (!response.IsSuccessful || !response.HasContent)`).
- `IsSuccessStatusCode` / `IsSuccessful` / `IsReceived` assert (via `[MemberNotNullWhen]`) that `Content`, `ContentHeaders`, and a non-null `Error` follow from the success/received state. They do not:
  - Per RFC 9110, a successful response can carry no body (204 No Content, a response to HEAD, a 2xx with zero-length or literal `null` content), so `Content`/`ContentHeaders` can be null while the status is successful.
  - `Error` is supplied independently of the status, so an unsuccessful response can have a null `Error` (Refit's own `EnsureSuccessfulAsync` already falls back when it does).

Closes #2156. Also adds regression coverage for #1933 and #1949 (generic-interface setups must be observed through the non-generic `IApiResponse`).

## What might this PR break?

- Source-level only, and in the direction of correctness: code that relied on the removed narrowings (e.g. dereferencing `response.Content` / `response.ContentHeaders` after `if (response.IsSuccessful)`, or `response.Error` after `if (!response.IsSuccessful)`) may now produce a nullable warning where the value can legitimately be null. The fix is to null-check, use `HasContent` / `IsSuccessfulWithContent` for content, or `HasRequestError` / `HasResponseError` for errors.
- No runtime behavior changes and no signature removals; the only public surface addition is `IsSuccessfulWithContent`.

## Checklist

## Additional information

Scenarios covered by the tests:

- Content-state matrix across the wrapper: success with content, success with a null body, success with an empty (non-null) body, a 204 No Content success, a 304 (not a success status), a non-success with content, a non-success without a captured error, a 2xx with a captured deserialization error, and a request error with no response received.
- Value-type bodies (`int`, where content is always present) and nullable value-type bodies (`int?`, which tracks null).
- Compile-time narrowing (no null-forgiving operator) for `IsSuccessful` / `IsSuccessStatusCode` / `IsReceived` (Headers/StatusCode/Version), `HasContent` and `IsSuccessfulWithContent` (Content), including through the covariant `IApiResponse<out T>`.
- That `IsSuccessful` and `IsSuccessStatusCode` are genuinely distinct (a 2xx with a deserialization error).
- That an unsuccessful response is not guaranteed to carry an `Error`.
- Mock/proxy interop (isolated in a new `Refit.MockInterop.Tests` project): a single setup on `IApiResponse<T>` is observed through the non-generic `IApiResponse` (including the problem-details and error-cast scenarios from the original reports), `IsSuccessfulWithContent` / `HasContent` narrowing through a mock, and that a success with null content headers is a representable interface state.

### DIFF
--- a/src/Refit/ApiResponse{T}.cs
+++ b/src/Refit/ApiResponse{T}.cs
@@ -70,6 +70,10 @@ public sealed class ApiResponse<T>(
     [MemberNotNullWhen(true, nameof(Content))]
     public bool HasContent => Content is not null;
 
+    /// <summary>Gets a value indicating whether the request was successful and deserialized <see cref="Content"/> is available.</summary>
+    [MemberNotNullWhen(true, nameof(Content))]
+    public bool IsSuccessfulWithContent => IsSuccessful && Content is not null;
+
     /// <summary>Gets the Refit settings used to send the request.</summary>
     public RefitSettings Settings { get; } = settings;
 
@@ -79,32 +83,22 @@ public sealed class ApiResponse<T>(
     /// <summary>Gets the HTTP response content headers as defined in RFC 2616.</summary>
     public HttpContentHeaders? ContentHeaders => response?.Content?.Headers;
 
-    /// <summary>Gets a value indicating whether the request was successful.</summary>
+    /// <inheritdoc />
     [MemberNotNullWhen(true, nameof(Headers))]
-    [MemberNotNullWhen(true, nameof(ContentHeaders))]
     [MemberNotNullWhen(true, nameof(StatusCode))]
     [MemberNotNullWhen(true, nameof(Version))]
-    [MemberNotNullWhen(false, nameof(Error))]
     public bool IsSuccessStatusCode => response?.IsSuccessStatusCode ?? false;
 
-    /// <summary>
-    /// Gets a value indicating whether the request was successful and there wasn't any other error (for example, during content deserialization).
-    /// </summary>
+    /// <inheritdoc />
     [MemberNotNullWhen(true, nameof(Headers))]
-    [MemberNotNullWhen(true, nameof(Content))]
-    [MemberNotNullWhen(true, nameof(ContentHeaders))]
     [MemberNotNullWhen(true, nameof(StatusCode))]
     [MemberNotNullWhen(true, nameof(Version))]
-    [MemberNotNullWhen(false, nameof(Error))]
     public bool IsSuccessful => IsSuccessStatusCode && Error is null;
 
     /// <inheritdoc />
     [MemberNotNullWhen(true, nameof(Headers))]
-    [MemberNotNullWhen(true, nameof(Content))]
-    [MemberNotNullWhen(true, nameof(ContentHeaders))]
     [MemberNotNullWhen(true, nameof(StatusCode))]
     [MemberNotNullWhen(true, nameof(Version))]
-    [MemberNotNullWhen(false, nameof(Error))]
     public bool IsReceived => response is not null;
 
     /// <summary>Gets the reason phrase which typically is sent by the server together with the status code.</summary>
--- a/src/Refit/ApiResponse{T}.cs
+++ b/src/Refit/ApiResponse{T}.cs
@@ -72,7 +72,7 @@ public sealed class ApiResponse<T>(
 
     /// <summary>Gets a value indicating whether the request was successful and deserialized <see cref="Content"/> is available.</summary>
     [MemberNotNullWhen(true, nameof(Content))]
-    public bool IsSuccessfulWithContent => IsSuccessful && Content is not null;
+    public bool IsSuccessfulWithContent => IsSuccessful && HasContent;
 
     /// <summary>Gets the Refit settings used to send the request.</summary>
     public RefitSettings Settings { get; } = settings;
--- a/README.md
+++ b/README.md
@@ -1562,11 +1562,26 @@ if (response.IsReceived)
 }
 ```
 
-`HasContent` is the covariance-safe way to access `Content` with null-state flow analysis — when it is `true` the
-compiler knows `Content` is non-null:
+A successful response does **not** imply a response body. Per [RFC 9110](https://www.rfc-editor.org/rfc/rfc9110.html) a
+204 (No Content), a 304 (Not Modified), a response to `HEAD`, or any 2xx with a zero-length or literal `null` body
+deserializes to `Content == null` *without* an error — so `IsSuccessful` / `IsSuccessStatusCode` being `true` tells you
+the request succeeded, not that `Content` is present. Use one of the content-presence members below when you need the
+deserialized body:
+
+* `HasContent` — the covariance-safe way to null-check `Content` on its own; when `true` the compiler knows `Content` is
+  non-null.
+* `IsSuccessfulWithContent` — a single check that combines success **and** content presence (`IsSuccessful && Content is
+  not null`); when `true` the compiler knows `Content` is non-null.
 
 ```csharp
 if (response.HasContent)
+{
+    // response.Content is non-null here
+    Process(response.Content);
+}
+
+// Or, when you want "succeeded and has a body" in one narrowing check:
+if (response.IsSuccessfulWithContent)
 {
     // response.Content is non-null here
     Process(response.Content);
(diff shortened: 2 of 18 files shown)


src/Refit/ApiResponse{T}.cs
```diff
@@ -70,6 +70,10 @@ public ApiResponse(
     [MemberNotNullWhen(true, nameof(Content))]
     public bool HasContent => Content is not null;
 
+    /// <summary>Gets a value indicating whether the request was successful and deserialized <see cref="Content"/> is available.</summary>
+    [MemberNotNullWhen(true, nameof(Content))]
+    public bool IsSuccessfulWithContent => IsSuccessful && Content is not null;
```
sguryev: Why not `HasContent` here?

glennawatson: Good call - done, thanks! Verified the nullable narrowing is identical, so reusing `HasContent` keeps it nicely DRY.

ChrisPulman (approved):

src/Refit/ApiResponse{T}.cs
```diff
@@ -79,32 +83,22 @@ public ApiResponse(
...
-    /// <summary>
-    /// Gets a value indicating whether the request was successful and there wasn't any other error (for example, during content deserialization).
-    /// </summary>
+    /// <inheritdoc />
     [MemberNotNullWhen(true, nameof(Headers))]
-    [MemberNotNullWhen(true, nameof(Content))]
-    [MemberNotNullWhen(true, nameof(ContentHeaders))]
     [MemberNotNullWhen(true, nameof(StatusCode))]
     [MemberNotNullWhen(true, nameof(Version))]
-    [MemberNotNullWhen(false, nameof(Error))]
```
sguryev, line 98: What the reason for removing this attribute?

glennawatson: `IsReceived` only tells you a response came back - it can't promise a body. A received response may be a 204/HEAD/empty/`null`-body case (RFC 9110 6.4.1 / 15.3.5), so `Content` and `ContentHeaders` aren't guaranteed, and a not-received/failed state can be constructed without an `Error`. The metadata that is guaranteed when a response exists - `Headers`/`StatusCode`/`Version` - is kept. Same annotation audit as #2156.

sguryev: This comment is about `IsSuccessful` property. But I think we can't clearly say the reason for false in this condition

image

glennawatson: Ah, you're right - this one's `IsSuccessful`, not `IsReceived`. And you've actually nailed why it was removed: `IsSuccessStatusCode && Error is null` is false whenever the status isn't a success, regardless of `Error` - so a non-2xx response with no captured error gives `IsSuccessful == false` and `Error == null`, which `[MemberNotNullWhen(false, nameof(Error))]` would have promised was impossible. Since "false" doesn't tell you the cause (status vs deserialization), it can't guarantee `Error`. For null-safe typed access, `HasRequestError` / `HasResponseError` are the way.

glennawatson:
Thanks for the thorough review here @sguryev, and your earlier contributions in this area are genuinely appreciated. On this one we're settled on the direction - the reasoning and RFC grounding are documented in #2156 - so we'll let it land and see where it lies in practice. Thanks again for the back-and-forth.

PR #2160, opened 2026-06-23
fix: disable errors by enabling preview features

TimothyMakkison:
When trying to build the Refit repository I get the following error, it appears to be caused by the use of `await`

image

CA2252: Opt in to preview features before using them

### Details
On JetBrains Rider 2025.3.1

10.0.109 [C:\Program Files\dotnet\sdk]
11.0.100-preview.1.26104.118 [C:\Program Files\dotnet\sdk]

### DIFF
--- a/src/Directory.Build.props
+++ b/src/Directory.Build.props
@@ -31,6 +31,7 @@
     <TreatWarningsAsErrors>true</TreatWarningsAsErrors>
     <ReportAnalyzer>true</ReportAnalyzer>
     <Nullable>enable</Nullable>
+    <EnablePreviewFeatures>true</EnablePreviewFeatures>
 
     <WarningsAsErrors>nullable;CS4014</WarningsAsErrors>
 


TimothyMakkison:
- Enabled previews in `Directory.Build.Props` instead of adding it to each `csproj`

glennawatson:
Closed for the moment. I think you just need to update to preview 5 of .NET 11 to fix those issues. https://dotnet.microsoft.com/en-us/download/dotnet/11.0

Looks like you're on net 11 preview 1

<blockquote><div>Microsoft</div><div><strong><a href="https://dotnet.microsoft.com/en-us/download/dotnet/11.0">Download .NET 11.0 (Linux, macOS, and Windows) | .NET</a></strong></div><div>.NET 11.0 downloads for Linux, macOS, and Windows. .NET is a free, cross-platform, open-source developer platform for building many different types of applications.</div></blockquote>

glennawatson:
If that doesn't solve the issue let me know btw.

TimothyMakkison:
Just after making this PR I downloaded preview 5, restarted and had the exact same issue

Tried it again this evening and was able to build and run the solution with no issue 😭 no idea what happened 😅

glennawatson:
Rider tends to keep a cache I noticed sometimes. So if you opened/closed Rider inbetween it'll change which .NET version it uses, especially if you are using Resharper Build.

PR #2174, opened 2026-06-26
feat: add generated request building support for path parameters in the URL.

### LINKED ISSUE #123: How to make Refit pass through proxy? (Charlie)

calebkiage:

Adds generated request building support for path parameters provided in the URL.

## What is the new behavior?

If a URL has path parameters defined in the URL and the method signature has parameters matching the path parameter, then the source generator will create a method that uses source generation.

The method signature below will now use a generated method body.

```csharp
interface IProfileClient {
    [Get("/users/{userId}"]
    Task<User> GetUser(string userId);
}
```

The old behavior used reflection for any methods that have parameters in the path.

## What might this PR break?

None

## Checklist
- [ ] Docs have been added or updated (for bug fixes / features)


glennawatson (commented):
Thanks for the draft. Know it's WIP - a few notes that should help when you pick it back up.

**Runtime bug in `GeneratedRequestRunner.BuildRequestPath`:**

```csharp
var isPathParam = index > 0 && toReplace[index - 1] == '/';
```

`toReplace[index - 1]` indexes the literal `"{key}"` instead of `path`; should be `path[index - 1]`. As written, `/users/{user}` throws `IndexOutOfRangeException` and `/{id}` leaves the placeholder unreplaced. One-char fix.

**Test gap:** the new tests only check generated text + `CompilesWithoutErrors`; nothing executes a path-param request. The one emit-load-invoke test uses `[Get("/users")]` (no param). An emit-load-invoke case asserting the final `RequestUri` (single, multiple, non-string, null/empty) would catch the bug above - most valuable addition here.

**Smaller, for later:**
- Empty value drops the whole segment (`/users/{user}` -> `/users`); reflection substitutes `?? string.Empty` -> `/users/`. Worth aligning so inline and reflective URIs match.
- `replacementBlockRegex` adds a per-match alloc into the parse hot path stage 1 de-LINQ'd, and the camelCase static name likely trips the analyzers; a manual scan avoids both and drops the duplicate brace logic.
- Confirm `params ReadOnlySpan<(string, string?)>` (C# 13) compiles on the net4x TFMs - no CI ran on the branch.

calebkiage:
@glennawatson, good catch on the bug. Was working on this late.

> Test gap: the new tests only check generated text + CompilesWithoutErrors; nothing executes a path-param request.

It looks like there are already tests that check for parameter substitution. They were failing and I've fixed most of them.

> Empty value drops the whole segment (/users/{user} -> /users); reflection substitutes ?? string.Empty -> /users/. Worth aligning so inline and reflective URIs match.

So if a path like `/users/{id}/orders` gets a null value then the new URL will be `/users//orders`?

---

How should generated code behave when `IUrlParameterFormatter` has been set? Currently, the generated code doesn't apply any formatting to the values.

glennawatson:
> So if a path like `/users/{id}/orders` gets a null value then the new URL will be `/users//orders`?

Yep, agreed -- and I think that's the behavior we actually want. It matches what reflection already does (`Format(...) ?? string.Empty`) and it's RFC 6570 simple expansion (an empty var expands to `""`), so inline and reflective URIs stay identical. I'd rather keep the `//` than have the generator special-case it and drift from reflection. If we ever want a null path arg to fail fast, that's a separate change we'd make in both paths.

> How should generated code behave when `IUrlParameterFormatter` has been set? Currently, the generated code doesn't apply any formatting to the values.

This is the part I care more about -- it's a real parity gap vs reflection. Let's get the generated code calling into it: route each path value through `settings.UrlParameterFormatter.Format(value, attributeProvider, type)` instead of `.ToString()`. The only snag is the `ICustomAttributeProvider` arg -- we don't want to reflect `ParameterInfo` at runtime under AOT. Could we pass the statically-known `typeof(TParam)` plus a generator-emitted (or empty) `ICustomAttributeProvider`, and emit a real per-parameter provider later if a formatter needs the attributes? That honors custom/type formatters in the common case and keeps us off reflection.

calebkiage:
I have the first iteration of `UrlParameterFormatter` support. Can you take a look?

src/Refit/GeneratedParameterAttributeProvider.cs
```diff
@@ -0,0 +1,33 @@
...
+        }
+
+        return allAttributes.ToArray();
+    });
+
+    /// <inheritdoc/>
+    public object[] GetCustomAttributes(bool inherit) => _allAttributesCache.Value;
+
+    /// <inheritdoc/>
+    public object[] GetCustomAttributes(Type attributeType, bool inherit) => attributes[attributeType];
```
glennawatson: `GetCustomAttributes(Type, bool)` is part of the `ICustomAttributeProvider` contract, and that contract says return an empty array when the attribute is not present, never throw. Right now the dictionary indexer throws `KeyNotFoundException` for any type not in the map. I reproduced it against the built `Refit.dll`: a path param carrying any non-Query attribute blows up inside `DefaultUrlParameterFormatter.GetFirstQueryAttribute` at request time.

Switching to `TryGetValue` makes it spec-compliant and lets you delete the fake-`QueryAttribute` fallback in `BuildParameterInfoField`. That injection only exists to dodge this throw, and it has the side effect of stamping a bogus `[Query]` onto path params.

```suggestion
    public object[] GetCustomAttributes(Type attributeType, bool inherit) =>
        attributes.TryGetValue(attributeType, out var value) ? value : [];
```

src/InterfaceStubGenerator.Shared/Parser.Request.cs
```diff
@@ -233,7 +269,11 @@ private static ImmutableEquatableArray<RequestParameterModel> ParseRequestParame
 
         for (var i = 0; i < parameters.Length; i++)
         {
-            var parsedParameter = ParseRequestParameter(parameters[i]);
+            var parameter = parameters[i];
+            var aliasAttr = parameter.GetAttributes().FirstOrDefault(static a => a.AttributeClass?.Name == "Refit.AliasAsAttribute");
+            var name = aliasAttr?.ConstructorArguments.FirstOrDefault().Value?.ToString() ?? parameter.Name;
```
glennawatson: `AttributeClass?.Name` returns the simple name (`AliasAsAttribute`), so comparing it to the fully-qualified `Refit.AliasAsAttribute` is always false. `AliasAs` on a route param never matches and the param quietly falls back to reflection. Confirmed in a scratch Roslyn run: `.Name` is `AliasAsAttribute`, `.ToDisplayString()` is `Refit.AliasAsAttribute`.

The same file already does this correctly a few methods down in `BuildFormFieldModel` (around line 543) using `ToDisplayString()` plus the existing `GetFirstStringArgument` helper. Reusing both keeps it consistent and drops the hand-rolled constructor-arg parsing:

```suggestion
            var aliasAttr = parameter.GetAttributes().FirstOrDefault(static a => a.AttributeClass?.ToDisplayString() == "Refit.AliasAsAttribute");
            var name = aliasAttr is not null ? GetFirstStringArgument(aliasAttr) ?? parameter.Name : parameter.Name;
```

src/InterfaceStubGenerator.Shared/Parser.Request.cs
```diff
@@ -294,9 +336,63 @@ private static ParsedRequestParameter ParseRequestParameter(IParameterSymbol par
...
+
+            static ITypeSymbol GetUnderlyingType(ITypeSymbol type)
+            {
+                return type is INamedTypeSymbol s && s.IsGenericType &&
+                       string.Equals(s.ContainingNamespace.Name, "System", StringComparison.Ordinal) &&
+                       string.Equals(s.Name, "Nullable", StringComparison.Ordinal)
+                    ? s.TypeArguments.Single()
+                    : type;
+            }
+        }
```
glennawatson, line 390: Two things here. The lowercase comparisons (`int`, `bool`, `string`, `char`, `byte`) are dead code: a symbol's `Name` is the CLR type name (`Int32`, `Boolean`, ...), never the C# keyword, so those branches never hit. And matching purely on `Name` is namespace-blind: a user's own `MyApp.Int32` or `MyApp.String` struct would be misclassified as a framework primitive.

Roslyn gives you an exact, allocation-free signal via `SpecialType`, which the codebase already uses (for example `Parser.Request.Helpers.cs:186`). This collapses the whole block, including the nested helpers and `GetUnderlyingType`, and handles `Nullable<T>` unwrapping in one shot:

```suggestion
        static bool IsSimpleType(ITypeSymbol type)
        {
            var underlyingType = type is INamedTypeSymbol { OriginalDefinition.SpecialType: SpecialType.System_Nullable_T } nullable
                ? nullable.TypeArguments[0]
                : type;
            return underlyingType.SpecialType is SpecialType.System_String
                or SpecialType.System_Boolean or SpecialType.System_Char
                or SpecialType.System_SByte or SpecialType.System_Byte
                or SpecialType.System_Int16 or SpecialType.System_UInt16
                or SpecialType.System_Int32 or SpecialType.System_UInt32
                or SpecialType.System_Int64 or SpecialType.System_UInt64;
        }
```

If you want to close the parity gap with reflection, this is the natural spot to add `System_Decimal`/`System_Double`/`System_Single` (one-liners) plus `Guid`/`DateTime`/enum handling. Happy to leave that for a follow-up.

src/InterfaceStubGenerator.Shared/Emitter.Inline.cs
```diff
@@ -112,8 +116,28 @@ private static string BuildInlineRefitMethod(
...
+        var parameterInfoNames = GetParameterInfoUniqueNames(request, uniqueNames);
+        var parameters = GetParametersArg(request, parameterInfoNames);
+        var paramInfoSb = new StringBuilder();
+
+        foreach (var parameter in request.Parameters)
+        {
+            if (parameter.Kind is RequestParameterKind.CancellationToken)
+            {
+                continue;
+            }
```
glennawatson: This loop emits a `GeneratedParameterAttributeProvider` static field for every parameter except the cancellation token, but only `Path` params are ever referenced (via `GetParametersArg`). So `[Post("/users/{id}")] Task Create(int id, [Body] User user)` generates a dead provider field for `user`. Narrowing the guard to `Path` keeps the generated output tight:

```suggestion
            if (parameter.Kind is not RequestParameterKind.Path)
            {
                continue;
            }
```

glennawatson (commented):
Really nice progress, Caleb. Routing path values through `UrlParameterFormatter` is the right move for reflection parity, and the span-based scan in `BuildRequestPath` is clean. I traced single-param, multi-param, leading-placeholder, and trailing-segment cases and they all came out correct.

I left four inline suggestions below. Two are correctness blockers (the `GetCustomAttributes` throw and the `AliasAs` name check, both verified against a real build / Roslyn run) and two are small cleanups (`SpecialType` over name matching, and not emitting provider fields for non-path params). Each has a one-click suggestion attached.

A few things that are not inline:

Test coverage. The new tests assert generated text plus `CompilesWithoutErrors`, which is good for shape, but nothing executes a path-param request and checks the final `RequestUri`. An emit-load-invoke test (single param, multiple params, non-string, null/empty, and importantly a path param carrying a non-Query attribute) would have caught the `GetCustomAttributes` crash immediately. Highest-value addition before merge.

Type coverage (worth pulling in). `Guid`, `DateTime`/`DateTimeOffset`, `decimal`, `double`/`float`, and enums currently fall back to reflection. `Guid` and `DateTime` are probably the two most common route types, so accelerating them is a real win rather than an edge case. The `SpecialType` rewrite in the `IsSimpleType` suggestion makes the numeric ones (`System_Decimal`, `System_Double`, `System_Single`) a one-line add; `Guid`/`DateTime`/enum need a value-formatting parity check but are straightforward. Fine as a fast-follow if you would rather keep this PR tight.

Dotted and round-trip placeholders (`{request.Prop}`, `{**param}`). Reflection supports these and your code correctly falls back today, so there is no regression. The ask is just to add a test that pins that fallback so it cannot silently break later. If you decide to extend coverage to them, @TimothyMakkison models them in #2182 as a small `RouteFragmentModel` discriminated union (constant / standard-param / object-access / round-trip), which is a good reference.

Perf. We do care about allocations on this path, so this one is more than a nicety. `BuildRequestPath` re-parses the template and allocates a `Dictionary` plus a 1024-char pooled buffer on every call. For the common 1-2 param case, a linear scan over the tuple array (drop the dictionary) and a smaller `stackalloc` buffer would be allocation-free. The dictionary only pays off if the same placeholder repeats, which is rare. Bigger picture: #2182 sidesteps the runtime re-parse entirely by emitting straight-line `ValueStringBuilder.Append` calls at compile time. Not required for this PR, but worth seeing as a direction since perf is a focus here.

Heads-up on overlap: @TimothyMakkison is building the same feature in #2182. Rather than have you both work it in parallel, the plan is to land this PR as the base and pull in a few of his ideas as we go (all optional, your call on scope). Thanks for jumping on this, it is a useful chunk of the AOT story.

calebkiage:
Thanks for the review. I've addressed most of the changes and have also added some tests. I will update the PR with more tests later.

On the suggestion for inline path building, is it not better to share the logic than generate it on every request method? It would reduce the amount of generated code in the end. If I build the path inline, there will be a lot of duplicated code for parameter building. If perf is an issue, I can optimize it further by passing in the template and span ranges to replace instead of the parameter name. i.e. `BuildRequestPath("/user/{id}", allowUnmatchedParameter: false, ((6, 10), "234"))`. What do you think of that idea over using generated path building?

glennawatson:
We can run some benchmarks and see either way. I'm happy whit the approach of using shared approach as long as our reflection costs go down to close as 0 in the known cases as possible.

glennawatson:
btw @calebkiage if it helps @TimothyMakkison is on Slack if you want to talk to him about his approach. Had a bit of a chat to him on there and he agreed taking your PR was best, since you put a bit of effort in before. But if you want you can swap ideas between each other there. https://reactiveui.net/Slack/

<blockquote><div><strong><a href="https://reactiveui.net/Slack/">Just a moment...</a></strong></div></blockquote>

calebkiage:
@glennawatson, I can't join that workspace. It says my account is deactivated.

glennawatson:
> @glennawatson, I can't join that workspace. It says my account is deactivated.

Ah ok, we can use comments then that's fine.

calebkiage:
@glennawatson, I think I broke the incremental compilation tests on the latest commit and I'm not sure where the problem is. Could you take a look?

glennawatson:
Pushed a commit to get the build and tests green.

- Migrated the new path-parameter RestService tests to the Refit.Testing StubHttp API (dropped MockHttpMessageHandler, which is no longer referenced)
- Added a missing await and fixed the analyzer errors in the new tests; split the large runner test into a partial file
- Reworked how the generator model stores parameter locations and attributes

Why I moved it off ImmutableArray and the AttributeData field for the incremental engine:

- The incremental generator caches one model per method and only re-runs when that model changes by value equality
- ImmutableArray<T> compares by its backing-array reference, so a fresh parse never matched the cached model and it regenerated on every edit (this is what the 3 Incremental tests were catching)
- AttributeData is worse: it never compares equal across compilations, and holding it roots the whole Compilation in the generator cache
- So locations now use ImmutableEquatableArray, and attributes are flattened into a small value record at parse time - no Roslyn symbols in the cache

Generated output is byte-identical (Verify snapshots unchanged), and all tests pass locally.

src/Refit/GeneratedRequestRunner.cs
```diff
@@ -28,11 +28,63 @@ public static Uri BuildRelativeUri(HttpClient client, string relativePath, UrlRe
...
+
+        var pathSpan = relativePathTemplate.AsSpan();
+        using var sb = new ValueStringBuilder(stackalloc char[256]);
+        var pos = 0;
+        foreach (var ((startIdx, endIdx), value) in uriParams)
+        {
+            sb.Append(pathSpan[pos..startIdx]);
+            if (value is not null)
+            {
+                sb.Append(value);
```
glennawatson: Values are appended verbatim here, so they are never URL-encoded. The reflection path wraps every value in `StringHelpers.EscapeDataString(...)` (see `RequestBuilderImplementation.RequestBuilding.cs`), so today the generated and reflection pipelines produce different URLs for the same input, and a value containing a space, `/`, `?`, `&`, or `#` yields a wrong or injectable URL.

Suggest escaping to match the reflection path (and add a test with a value like `"a b/c"`):

```suggestion
                sb.Append(StringHelpers.EscapeDataString(value));
```

glennawatson (commented):
Nice work - CI is green now (build and all tests pass across every TFM). Below is a quick pass on what is likely still needed before the reflection-free path can be called "done" - the main one is URL-encoding parity, the rest are coverage/robustness follow-ups. Specifics are in the inline comments.

- Encoding (blocker): the generated path substitutes formatted values verbatim, while the reflection path URL-encodes every value, so the two pipelines produce different (and unescaped) URLs for the same input - a correctness and injection risk.
- Coverage: only route-template placeholders ({id}, and ?q={q}) generate inline; plain auto-appended query params (Task<T> Get(string q) -> ?q=...), [Query] and [AliasAs] still fall back to reflection, so the generated-only/AOT path still throws for typical query APIs.
- Robustness: BuildRequestPath's trailing unmatched-parameter scan has a couple of edge cases (inline).
- Note (not a regression): value formatting still goes through IUrlParameterFormatter, which uses reflection for enums/TypeConverter - fine for now, but it means "reflection-free" holds at the request-shaping layer, not value formatting.

src/Refit/GeneratedRequestRunner.cs
```diff
@@ -28,11 +28,63 @@ public static Uri BuildRelativeUri(HttpClient client, string relativePath, UrlRe
...
+        sb.Append(pathSpan[pos..]);
+
+        var path = sb.ToString();
+        var i = path.IndexOf('{');
+        if (i < 0)
+        {
+            return path;
+        }
+
+        var j = i + path.AsSpan(i).IndexOfAny('}', '/');
```
glennawatson: Two edge cases in this trailing unmatched-parameter check:

1. If a substituted value legitimately contains `{`, this scan treats it as an unmatched template placeholder and throws - a false positive on valid input.
2. If neither `}` nor `/` follows the `{`, `IndexOfAny` returns -1, so `j = i - 1` and `path[j]` reads the character before the brace.

Safer to validate unmatched placeholders against the original template before substitution, and guard the -1 case. At minimum:

```suggestion
        var rel = path.AsSpan(i).IndexOfAny('}', '/');
        if (rel < 0)
        {
            return path;
        }

        var j = i + rel;
```

src/InterfaceStubGenerator.Shared/Parser.Request.cs
```diff
@@ -294,9 +347,47 @@ private static ParsedRequestParameter ParseRequestParameter(IParameterSymbol par
                 1);
         }
 
-        return TryParsePropertyParameter(parameter, parameterType, out var propertyParameter)
-            ? new(propertyParameter, true, 0, 0, 0)
+        if (TryParsePropertyParameter(parameter, parameterType, out var propertyParameter))
+        {
+            return new(propertyParameter, true, 0, 0, 0);
+        }
+
```
glennawatson, line 354: This is where a parameter with no `{...}` placeholder in the route falls back to `UnsupportedRequestParameter` -> reflection. That covers the most common Refit shape - the auto-appended query parameter (`[Get("/x")] Task<T> Get(string q)` -> `/x?q=...`), plus `[Query]`/`[AliasAs]` - so the generated-only/AOT path still throws for typical query-based APIs.

Not blocking this PR, but extending inline generation to auto-appended query params (and documenting what still needs reflection) is what is needed to actually remove reflection from these pipelines end to end.

src/Refit/GeneratedRequestRunner.cs
calebkiage: The 1st case should be handled by the URL escaping of the value and should now be safe. I've add a test to confirm it.
I have fixed the 2nd case and added some tests. The path to trigger it is strange so I didn't consider that. lol.

src/InterfaceStubGenerator.Shared/Parser.Request.cs
calebkiage: I intentionally didn't handle query parameters that aren't in the URL template since I haven't taken the time to understand how the reflection path builds them up. Can we handle that in a separate PR? @TimothyMakkison can also try his strategy for handling unspecified query parameters if he's interested, otherwise I can also handle it.

calebkiage:
> value formatting still goes through IUrlParameterFormatter, which uses reflection for enums/TypeConverter

Yeah. I'll need some time to think about how to properly handle this situation.

src/InterfaceStubGenerator.Shared/Parser.Request.cs
glennawatson: Yeah separate PR is fine.

glennawatson:
> > value formatting still goes through IUrlParameterFormatter, which uses reflection for enums/TypeConverter
> 
> Yeah. I'll need some time to think about how to properly handle this situation.

Let's just handle this in the next PR from either yourself or @TimothyMakkison

glennawatson (approved):

glennawatson:
Merged this one in, we can now tackle the rest as smaller PRs.

PR #2177, opened 2026-06-27
fix: annotate HasRequestError/HasResponseError out-params for nullable flow

HulinCedric:
### Problem

`IApiResponse.HasRequestError` / `HasResponseError` (and the `ApiResponse<T>` implementations) declare their `out` parameter as nullable (`out ApiRequestException?` / `out ApiException?`) but annotate it with only `[MaybeNullWhen(false)]`. On an already-nullable `out`, `[MaybeNullWhen(false)]` says nothing about the `true` branch, so the compiler keeps treating the value as *maybe-null* even when the method returns `true`:

```csharp
if (response.HasResponseError(out var error))
{
    var status = error.StatusCode; // CS8602: dereference of a possibly null reference
}
```

This is a nullable-annotation regression: in v11 the parameter was non-nullable `out ApiException` + `[MaybeNullWhen(false)]`, which *did* flow non-null on the `true` branch. When the out-params were made nullable (#2159), `[NotNullWhen(true)]` was not added, so the positive narrowing was lost.

### Fix

Replace `[MaybeNullWhen(false)]` with `[NotNullWhen(true)]` on all four declarations (interface + implementation, both methods). This matches the actual runtime contract — each method returns `true` exactly when `error` is non-null:

```csharp
error = Error as ApiException;
return error is not null;
```

so callers now get correct non-null narrowing in the `true` branch.

### Notes

- Annotation-only; **no runtime behavior change**.
- **No `PublicAPI.Shipped.txt` change required** — those baselines record `out Refit.ApiException? error`, and nullability *flow* attributes are not part of that signature.
- Verified `Refit.csproj` builds clean on `net8.0` (0 warnings).

Follows up #2159.

### DIFF
--- a/src/Refit/ApiResponse{T}.cs
+++ b/src/Refit/ApiResponse{T}.cs
@@ -147,7 +147,7 @@ public sealed class ApiResponse<T>(
         "CA1716:Identifiers should not match keywords",
         Justification = "By Design")]
     public bool HasRequestError(
-        [MaybeNullWhen(false)] out ApiRequestException? error)
+        [NotNullWhen(true)] out ApiRequestException? error)
     {
         error = Error as ApiRequestException;
         return error is not null;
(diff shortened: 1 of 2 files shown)


glennawatson (approved):

glennawatson:
Looks good. Thanks for the PR. We will be doing another minor release today with some other fixes so will go out with those.

PR #2195, opened 2026-07-02
feat: add generic tests to `ReflectionTests`

TimothyMakkison:
Add test for overloads of the same method with different generic type overloads. Think I've found a possible bug with generic interfaces support

Both `GetGenericParameter<T1>(T1 @value1)` and `GetGenericParameter<T1>(T1 @value1, TInterface @value2)` result in the error: 
`No suitable Method found...
   at Refit.RequestBuilderImplementation.FindMatchingRestMethodInfo(String key, Type[] parameterTypes, Type[] genericArgumentTypes) in `

### DIFF
--- a/src/tests/Refit.Tests/ReflectionTests.cs
+++ b/src/tests/Refit.Tests/ReflectionTests.cs
@@ -18,6 +18,62 @@ public sealed class ReflectionTests
     /// <summary>The query value exercised by the query parameter tests.</summary>
     private const string QueryValue = "queryValue";
 
+    /// <summary>A Refit interface with many overloaded generic methods.</summary>
+    public interface IGenericMethod
+    {
+        /// <summary>Request with a generic route parameter.</summary>
+        /// <param name="value1">Generic route parameter.</param>
+        /// <typeparam name="T1">Generic type of route parameter.</typeparam>
+        /// <returns>A task that completes when the request finishes.</returns>
+        [Get("/{value1}")]
+        Task GetGenericParameter<T1>(T1 value1);
+
+        /// <summary>Request with a generic route parameter and additional parameter.</summary>
+        /// <param name="value1">Generic route parameter.</param>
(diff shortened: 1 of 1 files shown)


TimothyMakkison:
Okay, converted to `Refit.Tests`, it appears that refit doesn't support generic parameters is a generic argument on the class. I don't know if this was always the case or if I'm missing something here. If this is expected I can remove the failing tests.


Suggestion: `Refit.Tests` should have an "accept anything" function similar to `Fallback` which would blanket respond with the same thing, having to specify your exact route / header / query may not always be necessary 🤷 
```c#
_mockHandler
    .Fallback
    .Respond("application/json", nameof(IBasicApi.GetParam));
```

glennawatson:
Yeah, if its a good use case in our new Refit.Testing framework with the ability to fallback please add it. The refit mocking framework is more flat, so more you add in lists/actions rather than fluent style, so if you can keep that style when adding this new feature that'd be awesome.

glennawatson:
eg our goal with Testing is a decent mock framework that is perf/alloc focused but still easy to use, that anyone using refit can then use in their test framework (so it needs to be test framework agnostic also)

TimothyMakkison:
Thanks for making the issue for the fallback, seems like a good first issue TBH.

I'm fairly confident that the error I've been having is a bug, see #2197. Thankfully I think this issue will be unintentionally fixed once Refit is 100% source generated, as the language/compiler will handle disambiguating method overloads and not Refit. I suspect the bug has existed in Refit for a very long time, but no one has noticed up until now. 😅

TimothyMakkison:
Fixed the tests by renaming one of the method overloads, hopefully when #2197 is fixed we can changed it back.

glennawatson (approved):

glennawatson:
ready for a merge?

TimothyMakkison:
> ready for a merge?

Yup, hopefully haven't forgotten any test cases 👍

Issue #2188, opened 2026-07-01
[Bug]: Caller-requested cancellation is wrapped in ApiRequestException instead of propagating OperationCanceledException


glennawatson:
### Describe the bug 🐞

Since v11 (PR #2052), **all** exceptions surfaced from `HttpClient.SendAsync` — including an `OperationCanceledException`/`TaskCanceledException` raised by a caller-supplied `CancellationToken` — are wrapped in `ApiRequestException` (which derives from `ApiExceptionBase`).

Caller-requested cancellation is control flow, not a transport error. Wrapping it breaks the normal .NET contract that cancellation propagates as an `OperationCanceledException`, so it is no longer caught by:

- ASP.NET Core request-abort handling / middleware that special-cases `OperationCanceledException`,
- `Polly`/retry policies that treat cancellation distinctly from transient faults,
- ordinary `catch (OperationCanceledException)` blocks.

Consumers are forced into awkward guards such as:

```csharp
catch (ApiRequestException ex) when (cancellationToken.IsCancellationRequested
                                     && ex.InnerException is OperationCanceledException)
{
    // treat as cancellation
}
```

This is a behavior change from pre-v11 and is arguably a regression. See discussion [#2162](https://github.com/reactiveui/refit/discussions/2162); a maintainer agreed it is likely a regression and worth an issue.

### Steps to reproduce

1. Define a Refit interface method taking a `CancellationToken`, e.g. `Task<Foo> GetFooAsync(CancellationToken ct);`.
2. Call it with a token that is cancelled while the request is in flight (`cts.CancelAfter(...)` or cancel before the send completes).
3. Observe the thrown exception type.
4. Expected `OperationCanceledException`/`TaskCanceledException`; actual is `ApiRequestException` wrapping it.

### Expected behavior

When cancellation is **explicitly requested by the caller's token** (i.e. `cancellationToken.IsCancellationRequested` is true for the token passed into the call), Refit should let the original `OperationCanceledException`/`TaskCanceledException` propagate unwrapped.

Non–caller-initiated cancellations that actually represent a transport failure (e.g. `HttpClient.Timeout` firing, which surfaces as a `TaskCanceledException` with an inner `TimeoutException` and **without** the caller's token being signalled) should keep the current `ApiRequestException` wrapping so genuine timeouts are still reported as request errors.

### Implementation notes (for whoever picks this up)

[...]

PR #2203, opened 2026-07-07
test: cover TransportExceptionFactory cancellation and timeout behavior

arpitjain099:

- Tests only. No product code changes.

## What is the new behavior?

- Adds tests for the TransportExceptionFactory send path added in #2205:
  - Caller-requested cancellation is rethrown unwrapped as OperationCanceledException.
  - A cancellation the caller did not request (for example a timeout) is still wrapped in ApiRequestException.
  - A custom TransportExceptionFactory controls the exception that is surfaced.

- These branches were not covered by tests after #2205 merged.

## What might this PR break?

- None. Tests only.

## Checklist
- [ ] Docs have been added or updated (for bug fixes / features)
- [x] Changes target the main branch

## Additional information

- The original fix was superseded by #2205, which added TransportExceptionFactory. Its default already propagates caller cancellation unwrapped, so the production guard was reverted and this PR now only adds the missing test coverage.

### DIFF
--- a/src/Refit/RequestExecutionHelpers.cs
+++ b/src/Refit/RequestExecutionHelpers.cs
@@ -386,6 +386,13 @@ internal static class RequestExecutionHelpers
                 .ConfigureAwait(false);
             return SendResult<T>.FromResponse(response);
         }
+        catch (OperationCanceledException) when (cancellationToken.IsCancellationRequested)
+        {
+            // Caller-requested cancellation is control flow, not a transport error.
+            // Propagate it unwrapped so callers, ASP.NET Core request-abort handling and
+            // Polly cancellation policies still see an OperationCanceledException.
+            throw;
+        }
         catch (Exception ex)
         {
             if (!isApiResponse)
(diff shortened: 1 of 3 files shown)


glennawatson:
Hi @arpitjain099, thank you for this fix and for taking the time to add tests with it.

While this was open, #2205 was merged into main. It adds a TransportExceptionFactory whose default behaviour already lets a caller-requested cancellation propagate unwrapped, which is exactly the #2188 regression your change set out to solve. Because that case is now handled by the factory, the small guard in this PR would sit in front of the factory and stop it from ever seeing a cancellation, so I have reverted it.

Rather than close your PR, I have redirected it to the test coverage that was still missing after #2205, so your work still lands:

- Your caller-cancellation send test is kept. I only switched its cancel to the async overload to satisfy the analyzer.
- A new test confirms that a cancellation the caller did not request, such as a timeout, is still wrapped in ApiRequestException.
- A new test confirms that a custom TransportExceptionFactory controls the exception that is surfaced.

I pushed those changes to your branch as a maintainer edit. It builds cleanly and the send-path tests pass locally. The PR stays yours, so the credit is yours too. Thanks again for pushing on this.

glennawatson (approved):

PR #2205, opened 2026-07-07
feature: Create a TransportExceptionFactory to let the user decide the exception to be thrown

### LINKED ISSUE #2202: Avoid breaking change on 11.0.0
The issue I have is that I have a codebase that relies on Refit and with the addition of #1238 is making me difficult to upgrate Refit to the latest version.
In the API I built I am relying on the Exceptions thrown by any requestHandler, and depending on the exception I write a statusCode or other in the response.
I have basically an catch with the exceptions thrown, not with the InnerExceptions thrown and I do not want to change the codebase unwrapping the exception when there is an inner exception.

I believe that the breaking change is just on the PR that close the issue because previously it was just await client.SendAsync and now there is a catch that if it's not an ApiResponse it throws ApiRequestException with the inner exception.

I would expect a RefitSetting that if is on(By default could be the current behaviour) then just do throw else do what is doing right now

If you agree I can create a PR to avoid the BreakingChange for existing codebases

jgarciadelanoceda:
Bug fix. Fixes #2202 and #2188 

## What is the new behavior?
When the TransportExceptionFactory is specified avoid the exception to be thrown as ApiRequestException, and let the user decide the exception to be throw.
By default the OperationCanceledException is thrown but the user can overload the TransportExceptionFactory

The breaking change introduced in Refit 11, avoided an easy to migrate to new versions of Refit because the exceptions where shadowed as ApiRequestException

## What might this PR break?
N/A

## Checklist
- [X] I have read the [Contribute guide](https://www.reactiveui.net/contribute/index.html)
- [X] Tests have been added or updated (for bug fixes / features)
- [X] Docs have been added or updated (for bug fixes / features)
- [X] Changes target the `main` branch
- [X] PR title follows [Conventional Commits](https://www.conventionalcommits.org/)

### DIFF
--- a/src/Refit/RefitSettings.cs
+++ b/src/Refit/RefitSettings.cs
@@ -195,18 +195,20 @@ public class RefitSettings
     public Dictionary<string, object>? HttpRequestMessageOptions { get; init; }
 
     /// <summary>
-    /// Gets or sets a factory invoked when <see cref="HttpClient.SendAsync(HttpRequestMessage)"/> throws a transport-level
-    /// exception, giving callers full control over the exception that is ultimately thrown or captured.
+    /// Gets or sets a factory invoked when <see cref="HttpClient.SendAsync(HttpRequestMessage)"/> throws a
+    /// transport-level exception, giving callers full control over the exception that is ultimately surfaced.
     /// </summary>
     /// <remarks>
+    /// The factory receives the <see cref="HttpRequestMessage"/>, the raw transport exception, and the
+    /// <see cref="CancellationToken"/> that was passed to the call, and returns the exception to surface.
     /// <para>
-    /// The factory receives the <see cref="HttpRequestMessage"/> and the raw transport exception.
-    /// The returned exception is thrown directly (non-ApiResponse path) or stored as the error inside the
-    /// <see cref="ApiResponse{T}"/> (ApiResponse path); if it is not already an <see cref="ApiRequestException"/>,
-    /// it is automatically wrapped in one that carries the full request context.
+    /// By default, an <see cref="OperationCanceledException"/> is returned unchanged when
+    /// <see cref="CancellationToken.IsCancellationRequested"/> is <see langword="true"/> (so a cancelled
+    /// <see cref="ApiResponse{T}"/> call throws instead of populating <c>Error</c>); every other exception
+    /// is wrapped in an <see cref="ApiRequestException"/> that carries the full request context.
     /// </para>
     /// </remarks>
-    public Func<HttpRequestMessage, Exception, Exception> TransportExceptionFactory { get; set; }
+    public Func<HttpRequestMessage, Exception, CancellationToken, Exception> TransportExceptionFactory { get; set; }
 
 #if NET6_0_OR_GREATER
 
@@ -255,12 +257,13 @@ public class RefitSettings
         };
     }
 
-    /// <summary>The default <see cref="TransportExceptionFactory"/>.</summary>
+    /// <summary>Returns the default <see cref="TransportExceptionFactory"/> delegate for this instance.</summary>
     /// <returns>
-    /// A factory delegate that, given an <see cref="HttpRequestMessage"/> and an <see cref="Exception"/>,
-    /// returns a new <see cref="ApiRequestException"/> with the original exception as its inner exception.
+    /// A factory that passes an <see cref="OperationCanceledException"/> through unchanged when
+    /// <see cref="CancellationToken.IsCancellationRequested"/> is <see langword="true"/>, and wraps every
+    /// other exception in an <see cref="ApiRequestException"/> that captures the request and these settings.
     /// </returns>
-    private Func<HttpRequestMessage, Exception, Exception> DefaultTransportExceptionFactory()
-        => (req, ex) =>
-             ex is OperationCanceledException ? ex : new ApiRequestException(req, req.Method, this, ex);
+    private Func<HttpRequestMessage, Exception, CancellationToken, Exception> DefaultTransportExceptionFactory()
+        => (req, ex, ct) =>
+             ex is OperationCanceledException && ct.IsCancellationRequested ? ex : new ApiRequestException(req, req.Method, this, ex);
 }
--- a/src/Refit/RefitSettings.cs
+++ b/src/Refit/RefitSettings.cs
@@ -19,6 +19,7 @@ public class RefitSettings
         UrlParameterFormatter = new DefaultUrlParameterFormatter();
         FormUrlEncodedParameterFormatter = new DefaultFormUrlEncodedParameterFormatter();
         ExceptionFactory = new DefaultApiExceptionFactory(this).CreateAsync;
+        TransportExceptionFactory = DefaultTransportExceptionFactory();
     }
 
     /// <summary>Initializes a new instance of the <see cref="RefitSettings"/> class.</summary>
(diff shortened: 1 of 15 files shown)


src/Refit/RefitSettings.cs
```diff
@@ -192,6 +194,20 @@ public Func<
...
+    /// exception, giving callers full control over the exception that is ultimately thrown or captured.
+    /// </summary>
+    /// <remarks>
+    /// <para>
+    /// The factory receives the <see cref="HttpRequestMessage"/> and the raw transport exception.
+    /// The returned exception is thrown directly (non-ApiResponse path) or stored as the error inside the
+    /// <see cref="ApiResponse{T}"/> (ApiResponse path); if it is not already an <see cref="ApiRequestException"/>,
+    /// it is automatically wrapped in one that carries the full request context.
+    /// </para>
+    /// </remarks>
```
glennawatson, line 210: The doc does not match the behavior: it fires for any exception from the send (not only transport-level), and a factory result that is not an ApiExceptionBase is thrown as-is, not wrapped. Suggested rewrite:

```suggestion
    /// <summary>
    /// Gets or sets a factory invoked when the underlying send throws any exception, letting callers
    /// control the exception that is ultimately surfaced.
    /// </summary>
    /// <remarks>
    /// The factory receives the request and the raw exception and returns the exception to surface.
    /// If the returned exception derives from <see cref="ApiExceptionBase"/> it is thrown (non-ApiResponse
    /// path) or captured into <see cref="ApiResponse{T}"/>.Error (ApiResponse path). Otherwise it is thrown
    /// as-is on both paths and is not wrapped. With the default factory an <see cref="OperationCanceledException"/>
    /// is returned unchanged (so a cancelled ApiResponse call throws rather than populating Error) and every
    /// other exception is wrapped in an <see cref="ApiRequestException"/>.
    /// </remarks>
```

glennawatson (commented):
Thanks for tackling this - the opt-in factory is a clean approach and the tests cover both response shapes.

A few things to tighten up (inline suggestions below):

- Default timeout handling: the default treats every OperationCanceledException as caller cancellation, so a genuine HttpClient.Timeout (a TaskCanceledException raised while the caller token is not signalled) is no longer wrapped in ApiRequestException. #2188 suggested keeping real timeouts wrapped. If the simpler behavior is intended, please add a note to the README breaking-changes section so upgraders know timeouts now surface raw.
- The XML doc describes wrapping that the code does not actually do (inline).
- The factory never receives the CancellationToken, so a custom factory cannot make the timeout-vs-cancel distinction either (inline).
- The IApiResponse path now throws on cancellation instead of populating Error - also worth a README note (inline).
- Minor: the raw rethrow drops the original stack trace (inline), and the PR title has a typo ('excepton' -> 'exception').

src/Refit/RefitSettings.cs
```diff
@@ -192,6 +194,20 @@ public Func<
...
+    public Func<HttpRequestMessage, Exception, Exception> TransportExceptionFactory { get; set; }
```
glennawatson: The factory cannot tell caller cancellation from an HttpClient.Timeout because it never sees the token (see #2188). Passing the CancellationToken lets custom factories branch on it:

```suggestion
    public Func<HttpRequestMessage, Exception, CancellationToken, Exception> TransportExceptionFactory { get; set; }
```
This also needs the default factory lambda, the call site in RequestExecutionHelpers.cs, and the PublicAPI.Unshipped.txt entries (all TFMs) updated to the 3-arg form.

src/Refit/RefitSettings.cs
```diff
@@ -238,4 +254,13 @@ private static RefitSettings ForNamingConvention(
...
     }
+
+    /// <summary>The default <see cref="TransportExceptionFactory"/>.</summary>
+    /// <returns>
+    /// A factory delegate that, given an <see cref="HttpRequestMessage"/> and an <see cref="Exception"/>,
+    /// returns a new <see cref="ApiRequestException"/> with the original exception as its inner exception.
+    /// </returns>
+    private Func<HttpRequestMessage, Exception, Exception> DefaultTransportExceptionFactory()
+        => (req, ex) =>
+             ex is OperationCanceledException ? ex : new ApiRequestException(req, req.Method, this, ex);
```
glennawatson: Match the new signature (the default ignores the token):

```suggestion
    private Func<HttpRequestMessage, Exception, CancellationToken, Exception> DefaultTransportExceptionFactory()
        => (req, ex, _) =>
             ex is OperationCanceledException ? ex : new ApiRequestException(req, req.Method, this, ex);
```

src/Refit/RequestExecutionHelpers.cs
```diff
@@ -388,18 +388,24 @@ private static async Task<SendResult<T>> SendOrCaptureExceptionAsync<T, TBody>(
         }
         catch (Exception ex)
         {
-            if (!isApiResponse)
+            var transportException = settings.TransportExceptionFactory(request, ex);
```
glennawatson: Pass the token through so the factory can use it:

```suggestion
            var transportException = settings.TransportExceptionFactory(request, ex, cancellationToken);
```

src/Refit/RequestExecutionHelpers.cs
```diff
@@ -388,18 +388,24 @@ private static async Task<SendResult<T>> SendOrCaptureExceptionAsync<T, TBody>(
...
             }
 
-            var failure = ApiResponse.Create<T, TBody>(
-                request,
-                null,
-                default,
-                settings,
-                new ApiRequestException(request, request.Method, settings, ex));
-            return SendResult<T>.FromFailure(failure);
+            throw transportException;
```
glennawatson, line 409: `throw transportException;` resets the stack trace of the original exception (e.g. the cancellation). Re-throw via ExceptionDispatchInfo so the original stack and OperationCanceledException.CancellationToken are preserved:

```suggestion
            System.Runtime.ExceptionServices.ExceptionDispatchInfo.Capture(transportException).Throw();
            throw transportException; // unreachable; keeps the original stack trace intact
```

src/tests/Refit.Newtonsoft.Json.Tests/RestServiceIntegrationTests.GitHub.cs
```diff
@@ -458,6 +497,49 @@ public async Task RequestCanceledBeforeResponseReadWithIApiResponse()
 
         var fixture = handler.CreateClient<IGitHubApi>(GitHubBaseUrl);
 
+        await Assert.That(
+            () => (Task)fixture.GetUserWithMetadata(OrgName, cts.Token)).ThrowsExactly<TaskCanceledException>();
```
glennawatson, line 501: Behavior change worth calling out: with the default factory a cancelled IApiResponse<T> call now throws TaskCanceledException instead of populating ApiResponse.Error. This is reasonable per #2188, but it is breaking for IApiResponse consumers - please add a README breaking-changes note.

jgarciadelanoceda: Finally I decided to do not do a Breaking change.. Or just to adjust to #2188, you can tell me if still is breaking

glennawatson (approved):

glennawatson:
Thanks @jgarciadelanoceda LGTM. Merging it in now for you. 

Planning on doing another release over the weekend (maybe sooner), so will get your fixes in for then.

PR #2210, opened 2026-07-12
feat!: reflection-free source-generated request building

### LINKED ISSUE #1101: Refit code generation doesn't work with extern aliases
Hello!

I'm having trouble implementing Refit into a project I'm working on, because one of the interfaces used for Rest communication is using "extern alias" . It is needed because otherwise we are having conflicts with namespaces and classes from company nuggets we are referencing.

However, the extern alias seems to break code generation and I'm getting compile time errors from Refit.

Can detecting & resolving extern aliases be done in the scope of Refit library?

### LINKED ISSUE #2002: [Bug]: URL Segments getting encoded, with seemingly no way to toggle this
### Describe the bug 🐞

s there any way to stop Refit url encoding my URL segments, i'm using a Zoho API and it doesn't like it when it gets encoded.
Example:

```
[Post("/{prevCalendarId}/events/{**eventId}?destinationcaluid={destinationCalendarId}")]
    Task<CalendarEvent> MoveEventToNewCalendar(string prevCalendarId, 
       string eventId, string destinationCalendarId);
```

The eventId is passed as something like 0000121312312313@zoho.com, but the @ symbol is being encoded, to something like this:

`/api/v1/calendars/9000000001ea4a9287a548f28ca330b0/events/3bf00000000488fda0ec154ee%40zoho.com?destinationcaluid=10e03c34110f402dbf8368e500asdas0`

I have seen various options to stop it encoding Query strings, but not the actual URL segments like above.

### Reproduction repository

### Expected behavior

This should happen...

### Screenshots 🖼️

_No response_

### IDE

_No response_

### Operating system

_No response_

### Version

_No response_

### Device

_No response_

### Refit Version

_No response_

### Additional information ℹ️

_No response_

### LINKED ISSUE #2165: feature: Pluggable return-type adapters (Retrofit CallAdapter.Factory equivalent)
## Summary

Refit currently hardcodes the set of supported method return types in the request builder/source generator (`Task`, `Task<T>`, `Task<IApiResponse<T>>`, `ApiResponse<T>`, `Stream`, `HttpResponseMessage`, `HttpContent`, `IAsyncEnumerable<T>`, `JsonDocument`). Adding a new return shape today means editing core. This proposes a **pluggable return-type adapter seam** so users (and us) can map the underlying HTTP call to arbitrary wrapper types without touching core.

## Prior art: Retrofit `CallAdapter.Factory`

Retrofit (the Java library Refit is modelled on) solves this with [`CallAdapter<R, T>`](https://github.com/square/retrofit/blob/master/retrofit/src/main/java/retrofit2/CallAdapter.java) + `CallAdapter.Factory`. The factory is consulted by **declared return type**, and ships as separate modules: RxJava (`Observable<T>`/`Single<T>`/`Completable`), Guava `ListenableFuture<T>`, Java8 `CompletableFuture<T>`, Scala `Future`. The core `Call<T>` stays transport-only; the adapter decides the surfaced type.

## Motivation for Refit / ReactiveUI

- Refit **used to** support `IObservable<T>` returns and dropped them. A proper seam would let us re-introduce `IObservable<T>` / Rx returns as an **out-of-band package** (e.g. `Refit.Reactive`) instead of bloating core — directly valuable for the ReactiveUI ecosystem.
- Lets users adapt to their own wrappers (`Result<T>`, `Option<T>`, custom envelope types) without a PR to Refit.

## Design notes / the hard part

The seam must work with the **source generator**, not just the reflection path — this is what makes it a real design effort rather than a small add. Sketch:

- An interface roughly like `IReturnTypeAdapter` exposing `bool CanHandle(Type returnType)` and an adapt step from the internal call (the thing that produces `Task<HttpResponseMessage>` / deserialized body) to the target type.
- The generator needs to either (a) emit a call into a registered adapter when it sees an unknown return type, or (b) recognise a known set of adapter-provided shapes at generate time. Decide whether adapters are **compile-time discoverable** (best for AOT/trim) or **runtime-registered** (more flexible, weaker AOT story). Lean compile-time to preserve Refit's AOT guarantees.
- Registration surface should mirror `RefitSettings.ContentSerializer` (settings-level) and the `RegisterGeneratedFactory` pattern for the AOT path.

## Acceptance criteria

- A documented extension point that adds a new return type without modifying `RequestBuilderImplementation` / core.
- Works in both the reflection and generated paths, with AOT/trim behaviour unchanged for users who don't use it.
- A proof-of-concept adapter (e.g. `IObservable<T>` or `Result<T>`) living outside Refit core.

## Pointers

- Reflection return-type dispatch: `src/Refit/RequestBuilderImplementation.Execution.cs`, `RestMethodInfoInternal.cs`.
- Generator return-type handling: `src/InterfaceStubGenerator.Shared/*`.
- Retrofit reference: `~/square/retrofit/retrofit/src/main/java/retrofit2/CallAdapter.java` and `retrofit-adapters/`.

### LINKED ISSUE #2168: feature (minor): Valueless query flags via [QueryName] (Retrofit @QueryName equivalent)
## Summary

Refit's `[Query]` always emits `key=value`. There's no way to emit a **valueless query flag** — a bare `?filter` with no `=` — which some APIs use for presence-style switches.

## Prior art: Retrofit `@QueryName`

Retrofit has [`@QueryName`](https://github.com/square/retrofit/blob/master/retrofit/src/main/java/retrofit2/http/QueryName.java): the parameter value becomes the query parameter *name* with no value. A `List`/array produces multiple bare flags (`?a&b&c`). `encoded=true` disables encoding of the name.

## Proposal

Add an equivalent — most naturally a small attribute (e.g. `[QueryName]`) or an option on `[Query]` — that renders the parameter's string value as a valueless query segment. Collections render as repeated bare names. Honour the existing query encoding/URI-format settings.

## Considerations

- Interaction with `UrlResolutionMode` and `QueryUriFormatAttribute`.
- Null handling: omit on null (consistent with `[Query]`).
- Must be implemented in both the reflection and source-generator paths.
- Decide attribute-vs-`[Query]`-option; a separate attribute reads clearer and matches Retrofit.

## Acceptance criteria

- A parameter renders as a bare `?name` flag; collections render as repeated names.
- Works in reflection + generated paths.
- TUnit tests for scalar and collection cases; README/docs entry.

## Pointers

- Query building: `src/Refit/RequestBuilderImplementation.QueryAndHeaders.cs` and `.Helpers.cs`.
- Attribute examples to mirror: `src/Refit/QueryAttribute.cs`.
- Generator: `src/InterfaceStubGenerator.Shared/*`.
- Retrofit reference: `~/square/retrofit/retrofit/src/main/java/retrofit2/http/QueryName.java`.

### LINKED ISSUE #2169: feature (minor): Per-parameter URL encoding opt-out (Retrofit encoded=true equivalent)
## Summary

Refit encodes URL `[Query]` and path segment values automatically, and the only controls are method-level (`[QueryUriFormat]`) or coarse (`TreatAsString`). There's no **per-parameter** way to say "this value is already encoded, pass it through verbatim." This is the feature side of bug #2002 (URL segments getting encoded with no toggle).

## Prior art: Retrofit `encoded`

Retrofit puts an `encoded` flag on individual parameter annotations — [`@Query(value, encoded=true)`](https://github.com/square/retrofit/blob/master/retrofit/src/main/java/retrofit2/http/Query.java), `@Path(value, encoded=true)`, `@Field(..., encoded=true)`, `@QueryMap(encoded=true)`. When set, Retrofit skips URL-encoding for that parameter, letting callers pass pre-encoded values.

## Proposal

Add a per-parameter encoding opt-out, e.g. an `Encode = false` property on `[Query]` (and a path equivalent), so a single value can bypass automatic encoding while the rest of the request encodes normally. Resolves #2002.

## Considerations

- Interaction with `UrlResolutionMode` (RefitLegacy vs Rfc3986) and `QueryUriFormatAttribute` — define precedence (per-param should win).
- Path-segment opt-out is the case #2002 specifically hits; cover both query and path.
- Reflection + source-generator paths.
- Document the footgun: caller becomes responsible for producing valid encoded output.

## Acceptance criteria

- A per-parameter switch that passes a query/path value through without encoding.
- Closes #2002 (path segment case verified).
- TUnit tests for query and path; README/docs entry.

## Pointers

- Query/path building: `src/Refit/RequestBuilderImplementation.RequestBuilding.cs`, `RequestBuilderImplementation.QueryAndHeaders.cs`.
- Existing attribute: `src/Refit/QueryAttribute.cs` (`TreatAsString`, `Format`), `src/Refit/QueryUriFormatAttribute.cs`.
- Related: #2002, #2046.
- Retrofit reference: `~/square/retrofit/retrofit/src/main/java/retrofit2/http/{Query,Path,Field}.java`.

### LINKED ISSUE #2185: [Bug]: AddRefitGeneratedClient fails for methods with query parameters
### Describe the bug 🐞

`AddRefitGeneratedClient<T>()` throws `NotSupportedException` at runtime when a generated Refit method contains query parameters.

Minimal interface:

```csharp
public interface IAuthRestApi
{
    [Get("/signin")]
    Task<JsonNode> SignIn(
        [AliasAs("login")] string login,
        [AliasAs("password")] string password);
}
```

Registration:

```csharp
[JsonSerializable(typeof(JsonNode))]
public partial class ApiJsonSerializerContext : JsonSerializerContext
{
}

var settings = new RefitSettings
{
    ContentSerializer = new SystemTextJsonContentSerializer(
        ApiJsonSerializerContext.Default.Options)
};

services
    .AddRefitGeneratedClient<IAuthRestApi>(settings)
    .ConfigureHttpClient(client =>
        client.BaseAddress = new Uri("https://api.example.com"));
```

Calling:

```csharp
await api.SignIn("login", "password");
```

throws:

```text
System.NotSupportedException: This Refit client was created with the generated-only API, but the generated client needs the reflection request builder for 'SignIn'. Enable generated request building for this method or use RestService.For when reflection is acceptable.
   at Refit.GeneratedOnlyRequestBuilder.BuildRestResultFuncForMethod(...)
   at Refit.Implementation.Generated.IAuthRestApi.SignIn(...)
```

The source generator creates `IAuthRestApi.g.cs`, but the generated `SignIn` implementation still uses the reflection-based path:

```csharp
var refitFunc = refitRequestBuilder.BuildRestResultFuncForMethod(
    "SignIn",
    parameterTypes);
```

The same problem occurs when query parameters are explicitly included in the route:

```csharp
[Get("/signin?login={login}&password={password}")]
Task<JsonNode> SignIn(string login, string password);
```

`RefitGeneratedRequestBuilding` is `true`, `DisableRefitSourceGenerator` is not enabled, and source-generated `System.Text.Json` metadata is provided for `JsonNode`.

A parameterless method returning the same `JsonNode` type is generated correctly through `GeneratedRequestRunner.SendAsync`, so the problem does not appear to be caused by `JsonNode` or JSON serialization. Other methods containing ordinary query parameters also fall back to `BuildRestResultFuncForMethod`.

Expected behavior: the source generator should generate request construction for ordinary route and query parameters so the interface can be used through `AddRefitGeneratedClient<T>()` in NativeAOT applications without reflection. If this method shape is intentionally unsupported, the analyzer should report a compile-time diagnostic explaining that the method is incompatible with generated-only registration instead of allowing registration and failing only at runtime.

Tested with:

- Refit 12.1.0
- Refit.HttpClientFactory 12.1.0
- .NET 10
- NativeAOT

### Step to reproduce

Step to reproduce: See the complete reproduction steps in the bug description above.

### Reproduction repository

### Expected behavior

See the expected behavior in the bug description above.

### Screenshots 🖼️

_No response_

### IDE

_No response_

### Operating system

_No response_

### Version

_No response_

### Device

_No response_

### Refit Version

_No response_

### Additional information ℹ️

_No response_

### LINKED ISSUE #2190: [Bug]: Methods require explicit [Body] attribute for improved code-gen, implicit is insufficient
### Describe the bug 🐞

Follow on from #2183.  Methods without explicit attributes fail to use the optimised code-gen path.

### Step to reproduce

I have a very simple interface method which I've been using for years, note no explicit `[Body]` attribute:

```csharp
[Headers("Content-Type: application/json")]
public interface IMyApi
{
	[Post("/v1/somewhere")]
	Task<AuthenticationResponse> AuthMePlease(AuthenticationRequest request, CancellationToken token);
}
````

This generates the reflection based pipeline:

```csharp
public async global::System.Threading.Tasks.Task<global::Leapfrog.Common.AuthenticationResponse> AuthMePlease(global::Leapfrog.Common.AuthenticationRequest @request, global::System.Threading.CancellationToken @token)
            {
                var refitArguments = new object[] { @request, @token };
                var refitRequestBuilder = _requestBuilder ?? throw new global::System.InvalidOperationException("This generated Refit method requires a request builder.");
                var refitFunc = refitRequestBuilder.BuildRestResultFuncForMethod("AuthMePlease", ______typeParameters );

                return await ((global::System.Threading.Tasks.Task<global::Leapfrog.Common.AuthenticationResponse>)refitFunc(this.Client, refitArguments)).ConfigureAwait(false);
            }
```

Update the signature to explicitly include the `[Body]` (`Task<AuthenticationResponse> AuthMePlease([Body]AuthenticationRequest request, CancellationToken token);`) and we get a nice reflection-free impl:

```csharp
public global::System.Threading.Tasks.Task<global::Leapfrog.Common.AuthenticationResponse> AuthMePlease(global::Leapfrog.Common.AuthenticationRequest @request, global::System.Threading.CancellationToken @token)
            {
                var refitRequest = new global::System.Net.Http.HttpRequestMessage(global::System.Net.Http.HttpMethod.Post, global::Refit.GeneratedRequestRunner.BuildRelativeUri(this.Client, "/v1/somewhere", refitSettings.UrlResolution));
                #if NET6_0_OR_GREATER
                refitRequest.Version = refitSettings.Version;
                refitRequest.VersionPolicy = refitSettings.VersionPolicy;
                #endif
                refitRequest.Content = global::Refit.GeneratedRequestRunner.CreateBodyContent<global::Leapfrog.Common.AuthenticationRequest>(
                    refitSettings,
                    @request,
                    global::Refit.BodySerializationMethod.Default,
                    !refitSettings.Buffered);
                global::Refit.GeneratedRequestRunner.SetHeader(refitRequest, "Content-Type", "application/json");
                global::Refit.GeneratedRequestRunner.AddConfiguredRequestOptions(refitRequest, refitSettings, typeof(global::Leapfrog.WVI.SrsWebApi.IMyApi));
                return global::Refit.GeneratedRequestRunner.SendAsync<global::Leapfrog.Common.AuthenticationResponse, global::Leapfrog.Common.AuthenticationResponse>(
                    this.Client,
                    refitRequest,
                    refitSettings,
                    false,
                    true,
                    refitSettings.Buffered,
                    @token);
            }
```

### Reproduction repository

### Expected behavior

I was _hoping_ to not have to add explicit attributes to all my API functions, and that they'd 'just work' as currently written.

### Screenshots 🖼️

_No response_

### IDE

Rider Windows

### Operating system

Windows 11

### Version

_No response_

### Device

_No response_

### Refit Version

13

### Additional information ℹ️

_No response_

### LINKED ISSUE #2191: [Feature]: generate inline (reflection-free) request building for query parameters
Follow-up from #2174 (v13.0.0).

**Is your feature request related to a problem?**

#2174 made path/route-template parameters generate inline (reflection-free). Query parameters still fall back to the reflection request builder:

- auto-appended query params: `[Get("/users")] Task<T> Search(string q)` -> `/users?q=...`
- `[Query]` (including `[Query(Format=...)]`, nested/collection expansion, custom delimiters)
- `[AliasAs]` on a query parameter

A method with any of these is emitted against `BuildRestResultFuncForMethod` (the reflection tail), so it throws at runtime under `AddRefitGeneratedClient` (generated-only) or NativeAOT. Query params are the most common Refit shape, so the generated-only/AOT path is still unusable for typical query APIs. The runtime symptom is #2185.

Generate inline request building for methods whose only otherwise-unsupported parameters are query params. The generated query string must match the reflection path's escaping and formatting so the `RequestUri` is identical between the two.

- Leave it as a reflection fallback (status quo): keeps #2185 broken for the common case.
- #2182 models route/query fragments as a `RouteFragmentModel` union (constant / standard-param / object-access / round-trip), a useful reference for this and the dotted/round-trip work.

Generator (`src/InterfaceStubGenerator.Shared/`):
- `Parser.Request.cs`: `ParseRequestParameter` returns `UnsupportedRequestParameter` (~line 357) for a parameter with no `{...}` placeholder. Add a `RequestParameterKind.Query`, classify plain / `[Query]` / `[AliasAs]` params into it, and relax `canGenerateInline` in `ParseRequest`.
- Reuse the existing `[Query]` parsing (`ParseQueryAttribute*`, ~lines 580-625) and `IsSimpleType` (`Parser.Request.Helpers.cs`).

Runtime (`src/Refit/`):
- Add a query-append helper modeled on `GeneratedRequestRunner.BuildRequestPath`: linear scan, `ValueStringBuilder`, `StringHelpers.EscapeDataString` per value, handling collection expansion and delimiters. The reflection version is in `RequestBuilderImplementation.RequestBuilding.cs`.

Guardrails:
- Value formatting still goes through `IUrlParameterFormatter`; keep that boundary, the separate value-formatting issue covers it.
- Keep the RF006 fallback analyzer and `GeneratedRequestBuildingFallbackContractTests` in sync: as these shapes become inline, their `expectedFallback` cases flip.

Tests: emit-load-invoke asserting the final `RequestUri` (single param, multiple, `[AliasAs]`, `[Query(Format=...)]`, collection, null/empty, and a value that needs escaping such as `"a b/c"`) plus Verify snapshots.

**Additional context**

Deferred from the #2174 review. Related: #2185, #2168, #2169, #2182. Dotted (`{request.Prop}`) and round-trip (`{**param}`) placeholders stay on the reflection fallback (#2046) and are already covered by tests.

### LINKED ISSUE #2192: [Feature]: AOT-safe value formatting for generated request building
Follow-up from #2174 (v13.0.0).

**Is your feature request related to a problem?**

After #2174, request shaping is reflection-free, but value formatting for path params (and query values, once those go inline) still goes through `IUrlParameterFormatter` / `DefaultUrlParameterFormatter`, which uses reflection for enums and `TypeConverter`. So the reflection-free/AOT guarantee only holds at the request-shaping layer, not value formatting: enum, `TypeConverter`, and other non-primitive values can still hit trimming/AOT warnings or format wrong under `AddRefitGeneratedClient` / NativeAOT.

Format statically-known parameter types in generated code without runtime reflection, matching `DefaultUrlParameterFormatter` output and still honoring a custom `IUrlParameterFormatter` when one is set.

- Keep calling `IUrlParameterFormatter` for everything (status quo): simplest, but leaves reflection in the path and breaks the AOT promise for enum/`TypeConverter` values.
- Emit direct formatting only for provably-simple types and fall back to the formatter (and the reflection builder) for the rest: a reasonable first cut.

- Simple types (`IsSimpleType`, `Parser.Request.Helpers.cs`): emit direct `ToString(CultureInfo.InvariantCulture)` / `ToString("D")` matching `DefaultUrlParameterFormatter`.
- Enums: emit the resolved string at compile time from the enum symbol (plus any `[EnumMember]`/`Display`/`Description` the default formatter honors) instead of runtime enum reflection.
- Custom `IUrlParameterFormatter`: still call it, but pass the static `typeof(TParam)` and the generator-emitted `ICustomAttributeProvider` (`GeneratedParameterAttributeProvider`) instead of reflecting `ParameterInfo`.
- Anything not provably safe falls back to the reflection builder (RF006 then flags it at compile time).

**Additional context**

Deferred from the #2174 review. Best done after the extra path-parameter type coverage (Guid / DateTime / DateTimeOffset / decimal / double / float / enum), since that work is what surfaces types that need real formatting.

### LINKED ISSUE #2200: [Bug]: Generated files have RequiresUnreferencedCode attributes
### Describe the bug 🐞

So after updating to Refit 12 and when having the following properties in the project:

```xml
<TreatWarningsAsErrors>True</TreatWarningsAsErrors>
<EnableTrimAnalyzer>true</EnableTrimAnalyzer>
```

The App won't build as the generated files from the interfaces have `RequiresUnreferencedCodeAttribute` on them and there is no good way to mark consuming code with the same attributes.

So IL2026 errors will be generated for each interface like:

```
Using member 'Refit.IRequestBuilder.BuildRestResultFuncForMethod(String, Type[], Type[])' which has 'RequiresUnreferencedCodeAttribute' can break functionality when trimming application code. Building request delegates from reflected method metadata requires generic method metadata to be available at runtime. [/Users/tomaszcielecki/git/App/src/Data/Data.csproj]
```

So if I have interface `IMyApi` then the generated `IMyApi.g.cs` will have this error.

I am using `Refit.HttpClientFactory`, but doesn't seem to be related to this, as for the HttpClientFactory extension methods I can apply attributes.

### Step to reproduce

1. Enable `TreatWarningsAsErrors` and `EnableTrimAnalyzer`
2. Implement Refit interface
3. Build app

### Expected behavior

I would love for a way to handle the trimming warning. There is no way to add `[RequiresUnreferencedCode]` attribute on the interface, as it is only allowed on class, ctor and methods.

The only way I can handle the warning is to disable WarningsAsErrors on IL2026, which isn't ideal. I did try to add a rule to .editorconfig but it doesn't apply to these files as they are only spat out in obj folder.

### Screenshots 🖼️

_No response_

### IDE

_No response_

### Operating system

_No response_

### Version

_No response_

### Device

_No response_

### Refit Version

12.1.0+

### Additional information ℹ️

_No response_

glennawatson:

Feature, performance, and a breaking release (V14). It completes the move to source-generated, reflection-free request building.

## What is the new behavior?

Refit builds requests directly, so the common method shapes work in generated-only and Native AOT clients with no runtime reflection:

- Query strings: auto-appended parameters, `[AliasAs]`, `[Query]` (`Format`, `CollectionFormat`), scalar collections, and query-object flattening.
- Dictionary query parameters, plus `[QueryConverter]` for shapes only known at runtime.
- `[QueryName]` valueless flags and `[Encoded]` per-parameter encoding opt-out.
- Implicit `[Body]` on POST/PUT/PATCH, and an all-scalar form body serialized straight-line with no descriptor array and no boxing.
- Generic interface methods, and custom return types via `IReturnTypeAdapter`.
- Path parameters format straight into the path buffer: an integer with no intermediate string and no escaping (net6+), and other span-formattable values escaped span-to-string (net9+).

Generated code targets C# 7.3 and lights up newer syntax (nullable annotations and so on) when the consumer's language version allows.

Most of these shapes went through the reflection request builder, which is not trim- or AOT-safe. The RF006 analyzer now flags any method that still needs reflection, so a generated-only build fails at compile time instead of throwing at runtime.

Closes #2185
Closes #2190
Closes #2191
Closes #2192
Closes #2168
Closes #2169
Closes #2002
Closes #2165
Closes #1101
Closes #2200

## What might this PR break?

- Some hot-path async members return `ValueTask` instead of `Task` (`ApiResponse<T>.EnsureSuccess*Async`, the `IApiResponse<T>` extensions, `DefaultApiExceptionFactory.CreateAsync`, and the `RefitSettings` async delegates).
- The reflection request builder is now the opt-in `Refit.Reflection` package. Interfaces that all generate inline do not need it; RF006 names any method that does.
- A fully generated interface validates its route template on the first call, not in `RestService.For`.
- Query objects flatten by their declared type, not the runtime type, and honor the content serializer's property names.

Full detail is in `docs/breaking-changes.md`.

## Checklist

- [x] Tests have been added or updated
- [x] Docs have been added or updated

## Additional information

A lot of the path-parameter work here grew out of @TimothyMakkison's ideas. He was already exploring AOT path-parameter support in #2209 and said he planned to work on it. The `ISpanFormattable` fast-write for path values is his idea, and that commit credits him as co-author. Thanks, @TimothyMakkison.

Verified across net8 to net11: the full generator and runtime suites pass, the solution builds with no analyzer warnings, and generated code compiles down to the C# 7.3 baseline.


TimothyMakkison:
I see you are going the path of using one method for `BuildRequestPath` and creating serveral overloads. Instead consider using a single method 

```C#
 public static string BuildRequestPath(
        string relativePathTemplate,
        bool allowUnmatchedParameter,
        ReadOnlySpan<((int startIdx, int endIdx) range, string? value)> parameters)
```

Calling methods can use collection expressions when possible (c# version permitting) which will stack allocate the parameters.

I was going to eventually make this change if you didn't like #2209 

This won't work with generic parameters but it will work for params case.

ChrisPulman (approved):

PR #2246, opened 2026-07-15
feat: refactor `AppendFormattedValue` in `GeneratedQueryStringBuilder`

TimothyMakkison:
Refactor `AppendFormattedValue` to use `ValiueStringBuilder` and inline `FormatWithGrowth`

### DIFF
--- a/src/Refit/GeneratedQueryStringBuilder.cs
+++ b/src/Refit/GeneratedQueryStringBuilder.cs
@@ -255,34 +255,6 @@ public ref struct GeneratedQueryStringBuilder
         return _text.ToString();
     }
 
-#if NET6_0_OR_GREATER
-    /// <summary>Formats a span-formattable value into <paramref name="buffer"/>, growing a rented buffer until the value fits.</summary>
-    /// <typeparam name="T">The span-formattable value type.</typeparam>
-    /// <param name="value">The value to render.</param>
-    /// <param name="buffer">The target buffer, replaced with a larger rented buffer when the value overflows it.</param>
-    /// <param name="rented">The rented buffer to grow and return to the pool, or null while the stack buffer is in use.</param>
-    /// <param name="format">The compile-time format, or null for the default rendering.</param>
-    /// <returns>The number of characters written into <paramref name="buffer"/>.</returns>
-    [System.Diagnostics.CodeAnalysis.ExcludeFromCodeCoverage] // The grow back-edge only fires for a value larger than the stack buffer; the compiler's loop second-jump stays unreachable in practice.
-    private static int FormatWithGrowth<T>(T value, ref Span<char> buffer, ref char[]? rented, string? format)
-        where T : ISpanFormattable
-    {
-        int written;
-        while (!value.TryFormat(buffer, out written, format.AsSpan(), System.Globalization.CultureInfo.InvariantCulture))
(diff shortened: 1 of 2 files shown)


glennawatson:
@TimothyMakkison benchmarked this on .NET 10 - the `ValueStringBuilder` scratch idea is good, but the current form regresses the hot path and leaks the pool buffer. Suggesting a small "hybrid" tweak that keeps the clarity and fixes both.

## Hybrid `AppendFormattedValue` body

```csharp
using var buffer = new ValueStringBuilder(stackalloc char[FormatBufferLength]);
int written;
while (!value.TryFormat(buffer.RawChars, out written, format.AsSpan(), CultureInfo.InvariantCulture))
{
    buffer.EnsureCapacity(buffer.Capacity * BufferGrowthFactor);
}

var formatted = buffer.RawChars[..written];
if (escape)
{
    StringHelpers.AppendUriDataEscaped(ref target, formatted);
    return;
}

formatted.CopyTo(target.AppendSpan(written));
```

Two changes vs this PR:
- `using` - the PR never disposes the scratch, so a value longer than 128 chars leaks the rented array. `using` returns it.
- `RawChars[..written]` + `CopyTo(target.AppendSpan(written))` instead of `buffer.Length = written; target.Append(buffer.AsSpan())` - the `AsSpan`/`Append` path is what regressed the verbatim case.

## Numbers (net10, Ryzen 7 5800X, in-process; main = 1.00, >1 slower)

| value / mode      | this PR | hybrid |
|-------------------|---------|--------|
| int / verbatim    | 1.24    | 1.09   |
| int / escaped     | 1.18    | 1.21   |
| long / verbatim   | 1.10    | 1.02   |
| double / verbatim | 1.06    | 1.02   |
| Guid / verbatim   | 0.91    | 0.87   |

Allocations: zero in all three - identical, no change either way.

glennawatson:
The version in main still performs the best though.

glennawatson:
@TimothyMakkison you okay if we go for the hybrid approach documented above for this one. Yes it's a little less performant then the version in main, but, I think it combines the easier maintainability you introduced in this PR

glennawatson:
Thanks @TimothyMakkison! I've folded your feedback into `AppendFormattedValue` on main as part of the recent performance work (#2263) - the scratch buffer, the extracted `FormatWithGrowth`, the leak-free buffer return, and the `CopyTo(AppendSpan)` fast path all landed there, and it benchmarks as the fastest of the options. So I'll close this one out, but it genuinely shaped where things ended up - much appreciated.

TimothyMakkison:
Sorry, never got round to replying, happy you merged.

Had been meaning to write my own benchmarks like yours to compare the approaches, as I was surprised by the differences and wanted to know how many ns each operation took. I assume all of the slow down is caused by the instantiation of the `ValueStringBuilder` `struct`

I do wonder if it is worth decreasing the size of the initial `stackalloc` buffer as zeroing the stack does have a small cost (and most of the likely `ISpanFormattable` types are usually less that 32 characters).

I've been meaning to mention this along with the prospect of using `[SkipLocalsInit]` I do have a notepad with other ideas but haven't got round to proof reading it

image

https://www.meziantou.net/csharp-9-improve-performance-using-skiplocalsinit.htm

PR #2262, opened 2026-07-20
feat: add CollectionFormat.Indexed for indexed query collection expansion

### LINKED ISSUE #2256: Add a CollectionFormat DeepObject applicable to IEnumerable<object>
No

The addition of IQueryConverter is great but for some simple scenarios it might not worth it and might be simpler to have a solution to avoid using Query formatter always when is a IEnnumerable 

The generated source code should handle to build things as ?customQueryObject[0].Id=0&customQueryObject[0].Value=Value1&customQueryObject[1].Id=1&customQueryObject[1].Value=Value2

The Query formatter does well the job but this could work out of the box

**Additional context**
If you think it's valuable and you want to add this feature to the library I would be happy to contribute 

The DeepObject is an idea coming from OpenApi but I am open to other names too

jgarciadelanoceda:
Feature Closes #2256

## What is the new behavior?
Adds a new CollectionFormat.Indexed value that expands an IEnumerable<T> of complex objects into indexed key-prefix query parameters:
  ?items[0].Id=1&items[0].Value=a&items[1].Id=2

Both the source generator (inline, zero-reflection) and the reflection request builder paths are implemented. The generator parses an IEnumerable<T> parameter with CollectionFormat.Indexed and a complex element type into the new QueryParameterShape.DeepObjectCollection shape, emitting a loop with an incrementing index local and flattening each element's properties under the key[index].Property naming scheme.

It was possible with QueryFormatter

## What might this PR break?
n/a

## Checklist
- [X] I have read the [Contribute guide](https://www.reactiveui.net/contribute/index.html)
- [X] Tests have been added or updated (for bug fixes / features)
- [X] Docs have been added or updated (for bug fixes / features)
- [X] Changes target the `main` branch
- [X] PR title follows [Conventional Commits](https://www.conventionalcommits.org/)

### DIFF
--- a/src/InterfaceStubGenerator.Shared/Emitter.Inline.Query.Object.cs
+++ b/src/InterfaceStubGenerator.Shared/Emitter.Inline.Query.Object.cs
@@ -640,4 +640,98 @@ internal static partial class Emitter
         var prefix = ToNullableCSharpStringLiteral(property.PrefixSegment);
         return $"global::Refit.GeneratedRequestRunner.BuildQueryKey({emission.SettingsLocal}, {clrName}, null, {prefix})";
     }
+
+    /// <summary>Appends the statements that iterate a <c>[Query(CollectionFormat.Indexed)]</c> collection,
+    /// flattening each element's properties under an indexed key prefix: <c>key[0].Prop=val</c>.</summary>
+    /// <param name="sb">The statement builder.</param>
+    /// <param name="parameter">The parameter model.</param>
+    /// <param name="query">The query-binding metadata.</param>
+    /// <param name="providerField">The cached attribute-provider field name.</param>
+    /// <param name="emission">The shared emission locals and helper state.</param>
+    private static void AppendIndexedCollectionQueryStatements(
+        PooledStringBuilder sb,
+        RequestParameterModel parameter,
+        QueryParameterModel query,
+        string providerField,
+        in InlineValueEmission emission)
+    {
+        var bodyIndent = Indent(MethodBodyIndentation);
+        var guarded = parameter.CanBeNull;
+        var outerIndent = guarded ? bodyIndent + "    " : bodyIndent;
+
+        if (guarded)
+        {
+            _ = sb.Append(bodyIndent).Append(IfParameterPrefix).Append(parameter.Name).AppendLine(NotNullCheckSuffix)
+                .Append(bodyIndent).AppendLine("{");
+        }
+
+        // Wrap in a block so the index local is scoped to this parameter and cannot clash with other parameters.
+        var idxLocal = emission.QueryValueLocal + "_idx";
+        var blockIndent = outerIndent + "    ";
+        var foreachIndent = blockIndent + "    ";
+
+        _ = sb.Append(outerIndent).AppendLine("{")
+            .Append(blockIndent).Append("var ").Append(idxLocal).AppendLine(" = 0;")
+            .Append(blockIndent).Append(ForeachVarKeyword).Append(emission.QueryValueLocal)
+                .Append(" in @").Append(parameter.Name).AppendLine(")")
+            .Append(blockIndent).AppendLine("{");
+
+        AppendIndexedElement(sb, parameter, query, providerField, idxLocal, foreachIndent, emission);
+
+        _ = sb.Append(foreachIndent).Append(idxLocal).AppendLine("++;")
+            .Append(blockIndent).AppendLine("}") // close foreach body
+            .Append(outerIndent).AppendLine("}"); // close wrapper block
+
+        if (!guarded)
+        {
+            return;
+        }
+
+        _ = sb.Append(bodyIndent).AppendLine("}");
+    }
+
+    /// <summary>Appends one element's optional null guard, indexed key declaration, and flattened property statements.</summary>
+    /// <param name="sb">The statement builder.</param>
+    /// <param name="parameter">The enclosing parameter model.</param>
+    /// <param name="query">The query-binding metadata whose <c>ObjectProperties</c> describes each element's shape.</param>
+    /// <param name="providerField">The cached attribute-provider field name.</param>
+    /// <param name="idxLocal">The generated index counter local name.</param>
+    /// <param name="foreachIndent">The indentation inside the foreach body.</param>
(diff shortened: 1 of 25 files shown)


src/tests/.editorconfig
```diff
@@ -22,6 +22,7 @@ dotnet_diagnostic.S5332.severity = none # Using http protocol is insecure (test
 ###################
 dotnet_diagnostic.SST2016.severity = none # Prefer DateTimeOffset over DateTime on a visible signature — test models exercise DateTime handling deliberately
 dotnet_diagnostic.SST2410.severity = none # A created disposable is never disposed — test-local disposables are short-lived and GC-reclaimed
+dotnet_diagnostic.SST1522.severity = none # File length maximum — live test harnesses are intentionally consolidated in one file
```
glennawatson: Don't disable, use partial classes if needed. We can expand the length to 1000 from 500. It got changed recently with a change of analyzer platform, there's a configuration in the editorconfig.

glennawatson:
For `src/Refit.Reflection/RequestBuilderImplementation.Payload.cs`. Sorry for jumping in the same file was attempting to avoid it but see how you go merging, mostly that was for the allocation reduction work.

src/InterfaceStubGenerator.Shared/Emitter.Inline.Query.Object.cs
```diff
@@ -312,8 +317,10 @@ internal static string BuildNestedKeyExpression(
             return $"{prefixExpr} + {ToCSharpStringLiteral(alias)}";
         }
 
-        var formatterCall =
-            $"global::Refit.GeneratedRequestRunner.BuildQueryKey({emission.SettingsLocal}, {ToCSharpStringLiteral(property.ClrName)}, null, {prefixExpr})";
+        var propertyNameExpr = ToCSharpStringLiteral(property.ClrName);
+        var formatterCall = context.PreEscapedKeys
+            ? $"{prefixExpr} + {propertyNameExpr}"
```
glennawatson, line 323: This concatenates the key raw and emits via `AddPreEscapedKey`, so a base key or `[JsonPropertyName]` with a special char (e.g. `first name`) goes out unescaped, unlike every other query path. Escape the non-bracket segments at compile time - only the `[i]` should stay raw.

src/Refit.Reflection/RequestBuilderImplementation.Payload.cs
```diff
@@ -398,6 +405,78 @@ internal void AppendQueryParameter(
...
+        return true;
+    }
+
+    /// <summary>Expands a <c>[Query(CollectionFormat.Indexed)]</c> collection into indexed-prefix query pairs,
+    /// producing <c>key[0].Prop=val&amp;key[1].Prop=val</c> for each element's flattened properties.</summary>
+    /// <param name="queryParamsToAdd">The list receiving query parameters.</param>
+    /// <param name="collection">The enumerable parameter value.</param>
+    /// <param name="attr">The query attribute governing key naming.</param>
+    /// <param name="paramKey">The base query key for the parameter.</param>
+    internal void AppendIndexedCollectionParameters(
```
glennawatson, line 443: Since this ships as two independent paths, please add a test that runs one interface through both the generator and this reflection builder and asserts identical query strings - especially the 'nested collections fall back to the default format' rule.

src/Refit.Reflection/RequestBuilderImplementation.Payload.cs
```diff
@@ -43,41 +43,41 @@ internal void AddBodyToRequest(RestMethodInfoInternal restMethod, object param,
         switch (restMethod.BodyParameterInfo.Item1)
         {
             case BodySerializationMethod.UrlEncoded:
```
glennawatson, line 45: This `AddBodyToRequest` switch reindent is unrelated to the feature - please revert it so the diff stays scoped to `Indexed`.

src/InterfaceStubGenerator.Shared/Parser.Request.Query.cs
```diff
@@ -17,6 +17,9 @@ internal static partial class Parser
     /// <summary>The maximum nested-object depth flattened inline before the whole parameter falls back to reflection.</summary>
     private const int MaxNestingDepth = 32;
 
+    /// <summary>The underlying integer value of <c>CollectionFormat.Indexed</c>, used by the parser without a direct reference to the Refit assembly.</summary>
+    private const int IndexedCollectionFormatValue = 6;
```
glennawatson: Hardcoding the enum's int is fragile if `CollectionFormat` is ever reordered. Resolve `Indexed`'s value by name from the compilation symbol, or add a test pinning it to 6.

glennawatson (commented):
Solid approach - both the inline generator and reflection paths, docs, tests, and full public-API coverage. A few things before merge:

src/Refit.Reflection/RequestBuilderImplementation.Payload.cs
glennawatson: If in a analyzer reason it can remain though.

src/InterfaceStubGenerator.Shared/Emitter.Inline.Query.Object.cs
jgarciadelanoceda: Handle with a test.. It may be not needed?

src/Refit.Reflection/RequestBuilderImplementation.Payload.cs
jgarciadelanoceda: Fixed

src/Refit.Reflection/RequestBuilderImplementation.Payload.cs
jgarciadelanoceda: No, it was just Visual studio formatting options

src/InterfaceStubGenerator.Shared/Parser.Request.Query.cs
jgarciadelanoceda: Handled

glennawatson (approved):

glennawatson:
Thanks, will likely be another 14.0 beta release tomorrow with this change in it.

PR #2265, opened 2026-07-20
fix: generate inline query building for nullable value-type query obj.

### LINKED ISSUE #123: How to make Refit pass through proxy? (Charlie)

### LINKED ISSUE #2264: [Bug]: Source generator falls back to reflection for a nullable value-type (struct)
### Describe the bug 🐞

The source-generated request builder flattens a value-type (`struct`) `[Query]` object parameter inline, but the **nullable** form of the same struct (`MyStruct?`) is not recognized and the whole method silently falls back to the reflection request builder (and is flagged by `RF006`). A nullable reference-type query object (`MyClass?`) and a non-nullable struct query object (`MyStruct`) both generate inline; only `MyStruct?` does not.

### Step to reproduce

Using a simple struct that Refit flattens by its public readable properties (this is the shape from `Refit.GeneratedCode.TestModels`):

```csharp
public readonly struct GeneratedQueryParameter
{
    public GeneratedQueryParameter(string value) => Value = value;
    public string Value { get; }
    public override string ToString() => Value;
}

public interface IGeneratedUserApi
{
    [Get("/users/search")]
    Task<string> SearchUsersAsync(
        [Query] GeneratedQueryParameter hello,        // generates inline ✅
        [Query] GeneratedQueryParameter? maybeHello); // falls back to reflection ❌
}
```

### Reproduction repository

_No response_

### Expected behavior

`[Query] GeneratedQueryParameter? maybeHello` should generate inline just like the non-nullable parameter: when the parameter has a value, its underlying struct's properties are flattened into the query string; when it is `null`, the parameter contributes nothing — matching the reflection request builder exactly.

The actual behavior is that the presence of the nullable struct query object makes the whole method's request fall back to `BuildRestResultFuncForMethod` (reflection), and `RF006` reports that the method cannot be generated inline. The non-nullable variant is unaffected.

### Screenshots 🖼️

_No response_

### IDE

_No response_

### Operating system

_No response_

### Version

_No response_

### Device

_No response_

### Refit Version

14.0.0-beta.3

### Additional information ℹ️

For a `[Query]` parameter whose type is not a simple/`IFormattable` scalar, the generator tries to flatten it as a query object in `Parser.TryBuildFlattenedObjectQueryModel`, which calls `TryBuildQueryObjectProperties` → `IsInlineFlattenableQueryObject`. That predicate explicitly excludes `Nullable<T>`:

```csharp
internal static bool IsInlineFlattenableQueryObject(ITypeSymbol type) =>
    type.TypeKind is TypeKind.Class or TypeKind.Struct
    && type.SpecialType != SpecialType.System_Object
    && type.SpecialType != SpecialType.System_String
    && type is not INamedTypeSymbol { OriginalDefinition.SpecialType: SpecialType.System_Nullable_T } // <-- rejects Nullable<T>
    && !ImplementsEnumerable(type);
```

So `Nullable<GeneratedQueryParameter>` is never flattened, `TryBuildFlattenedObjectQueryModel` returns `null`, and the whole method falls back to reflection.

Notably, the generator **already** handles this exact shape when the nullable struct appears as a *nested property* of a query object (`Parser.BuildQueryObjectPropertyModel` unwraps `Nullable<T>` and sets `NestedThroughValue`, and the emitter dereferences through `.Value` inside the existing null guard). The gap is only at the **top-level parameter root**, which never unwraps the nullable before flattening.

rmja:
Adds live parity coverage (present value + null), an inline no-fallback test, and a nullable parameter in the generated-code compliance interface.

Fixes #2264

## What is the new behavior?

IsInlineFlattenableQueryObject excludes Nullable<T>, so the parameter root never unwrapped it. This mirrors the existing nested-property handling at the root: the parser unwraps to the underlying struct (keeping IsNullableValueType set), and the emitter dereferences through .Value inside the existing null guard (CanBeNull is already true for Nullable<T>).

A [Query] parameter whose type is a nullable value type (Nullable<T> over a struct Refit flattens by its public properties) fell back to the reflection request builder while the non-nullable form generated inline, and the method was flagged by RF006.

## What might this PR break?

None

## Checklist

## Additional information

The addition of the `NullableStructQueryObjectMatchesReflection` has increased the line count in that file too much, so that the I get "error SST1522: This file is 518 lines, over the maximum of 500". I really think that this test should be there, but I have been unable to suppress the error.

### DIFF
--- a/src/InterfaceStubGenerator.Shared/Emitter.Inline.Query.Object.cs
+++ b/src/InterfaceStubGenerator.Shared/Emitter.Inline.Query.Object.cs
@@ -42,7 +42,13 @@ internal static partial class Emitter
             providerField,
             query.CollectionFormatValue,
             ToLowerInvariantString(query.PreEncoded));
-        var scope = new ObjectFlattenScope("@" + parameter.Name, null, query.NestingDelimiter, string.Empty, indent);
+
+        // A Nullable<T> query object holds its underlying struct behind .Value; the != null guard above is the HasValue
+        // check, so the deref is always safe here. A reference-type object flattens directly off the parameter.
+        var accessExpr = query.ValueFormat.IsNullableValueType
+            ? "@" + parameter.Name + NullableValueAccess
+            : "@" + parameter.Name;
+        var scope = new ObjectFlattenScope(accessExpr, null, query.NestingDelimiter, string.Empty, indent);
         AppendObjectPropertyList(sb, context, query.ObjectProperties!.Value, scope, emission);
 
         if (!guarded)
--- a/src/InterfaceStubGenerator.Shared/Parser.Request.Query.cs
+++ b/src/InterfaceStubGenerator.Shared/Parser.Request.Query.cs
@@ -223,7 +223,16 @@ internal static partial class Parser
             return dictionaryQuery;
         }
 
-        if (TryBuildQueryObjectProperties(parameter.Type, parameterPrefixSegment, formattableSymbol, context) is not { } properties)
+        // A Nullable<T> query object flattens its underlying struct's properties. The object emitter reaches the value
+        // through .Value inside the HasValue guard it already emits for a nullable parameter (CanBeNull is true for a
+        // Nullable<T>), mirroring how a nested nullable-value-type property flattens. A Nullable<dictionary> is
+        // impossible, so only this object branch unwraps; BuildValueFormat keeps the Nullable<T> so the emitted access
+        // reads through .Value.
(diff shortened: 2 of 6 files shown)


glennawatson:
Use partial files is the general policy on focused tests. The line count should be 1000 now is that not being honoured in the latest code?

glennawatson:
8cc1c374c227f5cd05168e775d98741e06f5a485 is the commit I did earlier today increasing to 1000 lines.

glennawatson:
Let me fix the analyzer down stream so it honours the setting. 

Then we'll get this PR merged in.

Thanks for doing this one.

glennawatson (approved):

PR #2284, opened 2026-07-30
fix: Propagate cancellation exceptions during response buffering

### LINKED ISSUE #123: How to make Refit pass through proxy? (Charlie)

radixy0:

Bug fix

## What is the new behavior?

Caller cancellation during response buffering is propagated. Non-cancellation buffering failures remain best-effort.

If a request is cancelled while deserializing the response, the cancellation exception is swallowed, allowing deserialization of a partially consumed stream.

## What might this PR break?

None

## Checklist

## Additional information

Added a regression coverage test

### DIFF
--- a/docs/breaking-changes.md
+++ b/docs/breaking-changes.md
@@ -3,12 +3,21 @@
 Every breaking change and notable addition, newest first. Each major version links back to the feature documentation
 in the [main README](../README.md).
 
+* [V15.x.x](#v15xx)
 * [V14.x.x](#v14xx)
 * [V13.x.x](#v13xx)
 * [V12.x.x](#v12xx)
 * [V11.x.x](#v11xx)
 * [Updates in 8.0.x](#updates-in-80x)
 * [V6.x.x](#v6xx)
+
+## V15.x.x
+
+### Breaking changes in V15.x
+
+* **Response buffering now propagates caller cancellation.** If a caller cancels while Refit buffers response content,
+  Refit throws `OperationCanceledException` instead of continuing with partially consumed content. Other buffering
+  failures remain best-effort.
 
 ## V14.x.x
 
--- a/src/Refit/RequestExecutionHelpers.cs
+++ b/src/Refit/RequestExecutionHelpers.cs
@@ -637,10 +637,17 @@ internal static partial class RequestExecutionHelpers
             .ConfigureAwait(false);
     }
 
-    /// <summary>Attempts to buffer content into memory, ignoring buffering failures.</summary>
+    /// <summary>Attempts to buffer content into memory, ignoring buffering failures but honouring cancellation.</summary>
     /// <param name="content">The content to buffer.</param>
     /// <param name="cancellationToken">A token to cancel buffering.</param>
     /// <returns>A task that completes once buffering has been attempted.</returns>
+    /// <exception cref="OperationCanceledException">Thrown when the caller cancelled while buffering.</exception>
+    /// <remarks>
+    /// Buffering is best-effort: a failure just means the serializer reads the live stream instead. Cancellation is
+    /// not, because continuing would hand the serializer partially consumed content. The check sits after the catch
+    /// rather than in a filter on it, so it also covers the targets whose <c>LoadIntoBufferAsync</c> polyfill drops
+    /// the token and therefore never throws.
+    /// </remarks>
     internal static async Task TryBufferContentAsync(HttpContent content, CancellationToken cancellationToken)
     {
         try
@@ -651,6 +658,8 @@ internal static partial class RequestExecutionHelpers
         {
             _ = bufferingException;
         }
+
+        cancellationToken.ThrowIfCancellationRequested();
     }
 
     /// <summary>The outcome of attempting to send a request.</summary>
--- a/src/tests/Refit.Tests/GeneratedRequestRunnerTests.ResponseErrorHandling.cs
+++ b/src/tests/Refit.Tests/GeneratedRequestRunnerTests.ResponseErrorHandling.cs
@@ -317,6 +317,45 @@ public partial class GeneratedRequestRunnerTests
             .ThrowsExactly<OperationCanceledException>();
     }
 
+    /// <summary>Verifies cancellation while buffering the response stops before the serializer sees partially consumed content.</summary>
+    /// <returns>A task that represents the asynchronous operation.</returns>
+    [Test]
+    public async Task SendAsyncStopsBeforeDeserializationWhenCancelledDuringBuffering()
+    {
+        var serializer = new RecordingContentSerializer
+        {
+            DeserializedValue = new GeneratedResult(DeserializedResultValue)
+        };
+        using var tokenSource = new CancellationTokenSource();
+
+        // Cancel once the response exists, so cancellation lands on buffering rather than on the send itself.
+        var handler = new CapturingHandler(
+            async (_, _) =>
+            {
+                await tokenSource.CancelAsync();
+                return new HttpResponseMessage(HttpStatusCode.OK)
+                {
+                    Content = new StringContent("buffered")
+                };
+            });
+        using var client = CreateClient(handler);
+        using var request = new HttpRequestMessage(HttpMethod.Get, RelativeResourcePath);
+
+        await Assert
+            .That(
+                () => GeneratedRequestRunner.SendAsync<GeneratedResult, GeneratedResult>(
+                    client,
+                    request,
+                    CreateSettings(serializer),
+                    isApiResponse: false,
+                    shouldDisposeResponse: true,
+                    bufferBody: false,
+                    tokenSource.Token))
+            .ThrowsExactly<OperationCanceledException>();
+
+        await Assert.That(serializer.DeserializeCallCount).IsEqualTo(0);
+    }
+
     /// <summary>Verifies caller-requested cancellation during send is rethrown instead of being wrapped in <see cref="ApiRequestException"/>.</summary>
     /// <returns>A task that represents the asynchronous operation.</returns>
     [Test]


src/Refit/RequestExecutionHelpers.cs
```diff
@@ -647,6 +647,10 @@ internal static async Task TryBufferContentAsync(HttpContent content, Cancellati
         {
             await content.LoadIntoBufferAsync(cancellationToken).ConfigureAwait(false);
         }
+        catch (OperationCanceledException) when (cancellationToken.IsCancellationRequested)
+        {
+            throw;
+        }
```
glennawatson: Two problems here.

1. This is what turns CI red. With an already-cancelled token `LoadIntoBufferAsync` throws `TaskCanceledException`, and it now escapes ahead of the serializer's own cancellation, so the existing `SendAsyncRethrowsCancellationRequestedDuringDeserialization` fails its `ThrowsExactly<OperationCanceledException>` assertion on net9.0 and net11.0.

2. It is a no-op on net8.0 and below. The `LoadIntoBufferAsync(CancellationToken)` polyfill in `src/Shared/HttpContentExtensions.cs` discards the token, so buffering can never throw `OperationCanceledException` there. That is also why the new test needs `#if NET9_0_OR_GREATER`, and it means the bug still ships on net8.0, net4xx and netstandard2.0.

Throwing after the catch fixes both. It works on every TFM, and it throws a plain `OperationCanceledException` carrying the caller's token, so the existing test stays green:

```csharp
internal static async Task TryBufferContentAsync(HttpContent content, CancellationToken cancellationToken)
{
    try
    {
        await content.LoadIntoBufferAsync(cancellationToken).ConfigureAwait(false);
    }
    catch (Exception bufferingException)
    {
        _ = bufferingException;
    }

    cancellationToken.ThrowIfCancellationRequested();
}
```

The suggestion below just removes the new catch. The `ThrowIfCancellationRequested` line goes immediately after the existing catch block, which is outside the range GitHub lets me suggest on.

```suggestion
```

docs/breaking-changes.md
```diff
@@ -27,6 +27,9 @@ visible in three places.
   lambdas assigned to these delegates need no change; a delegate that returned `Task.FromResult(x)` should return
   `new ValueTask<T>(x)`, and a caller that stored an `EnsureSuccess...Async()` result as a `Task` should `await` it or
   call `.AsTask()`. `TransportExceptionFactory` is unaffected (it was already synchronous).
+* **Response buffering now propagates caller cancellation.** If a caller cancels while Refit buffers response content,
+  Refit throws `OperationCanceledException` instead of continuing with partially consumed content. Other buffering
+  failures remain best-effort.
```
glennawatson: This is a V15 change, but it is filed under "Breaking changes in V14.x". It wants its own section above V14 plus a TOC entry, rather than an append to the V14 list.

TOC, line 6:

```markdown
* [V15.x.x](#v15xx)
* [V14.x.x](#v14xx)
```

New section above `## V14.x.x`:

```markdown
## V15.x.x

### Breaking changes in V15.x

* **Response buffering now propagates caller cancellation.** If a caller cancels while Refit buffers response content,
  Refit throws `OperationCanceledException` instead of continuing with partially consumed content. Other buffering
  failures remain best-effort.
```

The suggestion below drops it from the V14 list.

```suggestion
```

src/tests/Refit.Tests/ResponseTests.cs
```diff
@@ -479,6 +485,37 @@ public async Task XmlDeserializationWithNonSeekableStream_DoesNotThrowStreamCons
...
+
+        var request = localFixture.GetTestObject(cancellationTokenSource.Token);
+        await contentStream.ReadStarted!;
+        await cancellationTokenSource.CancelAsync();
+
+        var exception = await Assert.That(() => (Task)request).Throws<OperationCanceledException>();
+
+        await Assert.That(exception!.CancellationToken).IsEqualTo(cancellationTokenSource.Token);
+    }
+#endif
```
glennawatson: This test passes without the product change, so it does not guard the regression. I checked by reverting `src/Refit/RequestExecutionHelpers.cs` to main, rebuilding and re-running it - still green.

The reason: once buffering swallows the cancellation, `FromHttpContentAsync` runs with the same already-cancelled token, and the stub stream's `ReadAsync` throws `OperationCanceledException` carrying that token anyway. Both assertions hold with or without the fix.

To make it real, the serializer must not be able to produce the same exception. A stub `IHttpContentSerializer` that records it was called and returns a value would do it - then assert it was never reached.

Once the fix uses `ThrowIfCancellationRequested`, the `#if NET9_0_OR_GREATER` guard can go too, so this covers net8.0 as well.

src/tests/Refit.Tests/ThrowOnGetLengthMemoryStream.cs
```diff
@@ -3,18 +3,55 @@
...
+using System.Threading;
+using System.Threading.Tasks;
 
 namespace Refit.Tests;
 
 /// <summary>A <see cref="MemoryStream"/> that can be toggled to behave like a non-seekable stream whose length is unavailable.</summary>
+[System.Diagnostics.CodeAnalysis.SuppressMessage(
+    "Style",
+    "SST1600",
+    Justification = "Blocking read support is test synchronization infrastructure.")]
```
glennawatson: This suppression only exists to silence two `SST1600` errors - on the private `_readStarted` field and on `WaitForCancellationAsync`. I confirmed that by removing it and rebuilding.

Doc comments on both fix the cause, and avoid muting the rule for every member added to this class later.

```suggestion
```

src/tests/Refit.Tests/ThrowOnGetLengthMemoryStream.cs
```diff
@@ -3,18 +3,55 @@
...
 public sealed class ThrowOnGetLengthMemoryStream : MemoryStream
 {
+    private readonly TaskCompletionSource _readStarted = new(TaskCreationOptions.RunContinuationsAsynchronously);
```
glennawatson:
```suggestion
    /// <summary>Signals that a blocked asynchronous read has begun.</summary>
    private readonly TaskCompletionSource _readStarted = new(TaskCreationOptions.RunContinuationsAsynchronously);
```

src/tests/Refit.Tests/ThrowOnGetLengthMemoryStream.cs
```diff
@@ -3,18 +3,55 @@
...
+
+    /// <summary>Gets or sets whether asynchronous reads wait for cancellation.</summary>
+    public bool BlockReads { get; set; }
+
     /// <summary>Gets or sets a value indicating whether the stream reports as seekable and exposes its length.</summary>
     public bool CanGetLength { get; set; }
 
+    /// <summary>Gets a task that completes when an asynchronous read starts.</summary>
+    public Task? ReadStarted => BlockReads ? _readStarted.Task : null;
```
glennawatson: Returning null when `BlockReads` is false forces `ReadStarted!` at the call site. The task is meaningful either way, so the nullable buys nothing.

Separately: the class is named for one behaviour (`ThrowOnGetLength`) but now does two. A second test double for the blocking read would read better than overloading this one.

```suggestion
    public Task ReadStarted => _readStarted.Task;
```

The `!` in `await contentStream.ReadStarted!` can go with it.

src/tests/Refit.Tests/ThrowOnGetLengthMemoryStream.cs
```diff
@@ -3,18 +3,55 @@
...
+            : base.ReadAsync(buffer, offset, count, cancellationToken);
+
+    /// <inheritdoc/>
+    public override ValueTask<int> ReadAsync(
+        Memory<byte> buffer,
+        CancellationToken cancellationToken = default) => BlockReads
+            ? new(WaitForCancellationAsync(cancellationToken))
+            : base.ReadAsync(buffer, cancellationToken);
+
+    private async Task<int> WaitForCancellationAsync(CancellationToken cancellationToken)
```
glennawatson:
```suggestion
    /// <summary>Blocks until the read is cancelled.</summary>
    /// <param name="cancellationToken">A token that releases the blocked read.</param>
    /// <returns>A task that never completes successfully.</returns>
    private async Task<int> WaitForCancellationAsync(CancellationToken cancellationToken)
```

glennawatson (commented):
Thanks for this - the bug is real and correctly located, and the product change is nicely minimal.

A few things to sort first:

- CI is red because of this change. `SendAsyncRethrowsCancellationRequestedDuringDeserialization` fails on net9.0 and net11.0. Reproduced locally.
- The new test passes without the product fix, so it does not guard the regression. Verified by reverting `RequestExecutionHelpers.cs` to main and re-running.
- The fix is a no-op on net8.0 and below, because the `LoadIntoBufferAsync` polyfill discards the token.
- The `SuppressMessage` should be two doc comments instead.
- The docs entry belongs in a new V15 section, not V14.

Suggestions inline for each.

glennawatson:
Thanks for this - the bug is real. I have rebased onto latest main and pushed the review points to your branch, keeping you as the author.

What changed from your version:

* The fix is now `cancellationToken.ThrowIfCancellationRequested()` after the existing catch, instead of a filtered `catch`/`throw`. Your version only worked on net9.0+, because the `LoadIntoBufferAsync(CancellationToken)` polyfill in `src/Shared/HttpContentExtensions.cs` drops the token on everything below, so buffering there never throws and the bug still shipped on net8.0, net4xx and netstandard2.0. It also broke `SendAsyncRethrowsCancellationRequestedDuringDeserialization`, which is what turned CI red.
* The regression test is replaced. I checked the original one against unpatched product code and it passed, so it was not guarding anything - once buffering swallowed the cancellation, the stub stream threw the same exception from the serializer instead. The new `SendAsyncStopsBeforeDeserializationWhenCancelledDuringBuffering` uses the existing `RecordingContentSerializer` and asserts `DeserializeCallCount == 0`, so it fails if the serializer is ever reached. Verified red against main's `RequestExecutionHelpers.cs` and green with the fix.
* Because the fix works on every target, the test needs no `#if NET9_0_OR_GREATER`, and the blocking-read machinery added to `ThrowOnGetLengthMemoryStream` is no longer needed, so that file and `ResponseTests.cs` are back to their original state. That also drops the `SuppressMessage` and the second responsibility on a class named for one behaviour.
* The breaking-changes note moved to its own V15.x.x section with a TOC entry, rather than being appended to the V14 list.

Result is 58 added lines across 3 files. Solution builds clean on all ten target frameworks, all 7505 tests pass, and `RequestExecutionHelpers.cs` keeps 100 percent line and branch coverage.

glennawatson (approved):

PR #2295, opened 2026-08-04
fix: differentiate explicitly empty string delimiter from "no delimiter"

### LINKED ISSUE #123: How to make Refit pass through proxy? (Charlie)

### LINKED ISSUE #2294: [Bug]: `[Query(delimiter: "")]` is ignored — nested query keys are always joined with `.`
### Describe the bug 🐞

When a complex query parameter is declared with an explicitly empty delimiter,
`[Query(delimiter: "")]`, Refit ignores the empty string and falls back to the default nesting
delimiter `.` when flattening nested properties.

An empty delimiter appears to be indistinguishable from "no delimiter specified" — most likely a
`string.IsNullOrEmpty(delimiter) ? DefaultNestingDelimiter : delimiter` (or equivalent) in the
query string builder.

This is a regression from Refit 10.1.6, where `[Query(delimiter: "")]` concatenated the parent and
child key with no separator.

## Why it matters

Some APIs delimit nested query keys with something other than `.`, and express range operators by
suffixing the field name, e.g.:

```
GET /members?created_at:ge=2025-01-01T00:00:00%2B01:00&created_at:lt=2025-02-01T00:00:00%2B01:00
```

The natural way to model that is a nested range object with `[AliasAs(":ge")]` on its members plus
an empty delimiter on the parent. That is impossible in 14.0.1: the injected `.` produces
`created_at.:ge`, which the server rejects.

### Step to reproduce

```csharp
public record Nested
{
    [AliasAs(":ge")]
    public string? Ge { get; set; }
}

public record Filter
{
    [AliasAs("created_at")]
    public Nested? CreatedAt { get; set; }
}

public interface IProbe
{
    [Get("/a")]
    Task A([Query(delimiter: "")] Filter filter);   // empty delimiter

    [Get("/b")]
    Task B(Filter filter);                          // no attribute

    [Get("/c")]
    Task C([Query(delimiter: ":")] Filter filter);  // non-empty delimiter
}

await api.A(new() { CreatedAt = new() { Ge = "X" } });
await api.B(new() { CreatedAt = new() { Ge = "X" } });
await api.C(new() { CreatedAt = new() { Ge = "X" } });
```

### Reproduction repository

_No response_

### Expected behavior

```
/a  ->  ?created_at:ge=X     (empty delimiter, nothing inserted)
/b  ->  ?created_at.:ge=X    (no attribute, default "." delimiter)
/c  ->  ?created_at:ge=X     (":" delimiter)
```

## Actual behavior

```
/a  ->  ?created_at.:ge=X    <-- delimiter "" ignored, "." used instead
/b  ->  ?created_at.:ge=X
/c  ->  ?created_at:ge=X     <-- non-empty delimiter is honoured correctly
```

`/a` and `/b` are identical, i.e. the explicitly empty delimiter has no effect.

The same happens for dictionary properties: with `[AliasAs("metadata.")]` and
`[Query(delimiter: "")]`, 10.1.6 produced `metadata.source=hello`, while 14.0.1 produces
`metadata..source=hello`.

### Screenshots 🖼️

_No response_

### IDE

_No response_

### Operating system

_No response_

### Version

_No response_

### Device

_No response_

### Refit Version

14.0.1

### Additional information ℹ️

The delimiter should only fall back to the default when it was not specified at all
(`delimiter is null`), not when it is an empty string. Something along the lines of:

```diff
- var delimiter = string.IsNullOrEmpty(queryAttribute?.Delimiter)
-     ? DefaultNestingDelimiter
-     : queryAttribute.Delimiter;
+ var delimiter = queryAttribute?.Delimiter ?? DefaultNestingDelimiter;
```

There is no way to repair the key downstream: `IUrlParameterKeyFormatter.Format` is only invoked
per-segment for non-aliased property names, never for the composed key, so the injected `.` is not
observable there. The only option is to avoid nesting altogether and flatten the object into
properties carrying the full key in `[AliasAs]`.

rmja:

Fixes #2294.

## What is the new behavior?

`QueryFormData` is `default` when no `[Query]` attribute is present, so `Delimiter is null` in that case — the `IsNullOrEmpty` guard was there for the absent case, but it swallows `""` along with it. Now `data.Delimiter ?? DefaultNestingDelimiter`, with `Delimiter` retyped to `string?` so the absent case is expressed as null rather than leaning on emptiness. Three call sites: object query, `CollectionFormat.Indexed` collections, and path-residual queries.

The source generated and reflection client generates different queries for empty string delimiters.

## What might this PR break?

There should be none, as this pr fixes a mismatch between the reflection based client generation and source generated client.

## Checklist

### DIFF
--- a/src/InterfaceStubGenerator.Shared/Parser.Request.Body.cs
+++ b/src/InterfaceStubGenerator.Shared/Parser.Request.Body.cs
@@ -300,14 +300,16 @@ internal static partial class Parser
         BodyBufferMode BufferMode);
 
     /// <summary>Form-relevant data parsed from a <c>[Query]</c> attribute on a parameter or body property.</summary>
-    /// <param name="Delimiter">The delimiter combined with the prefix.</param>
+    /// <param name="Delimiter">The delimiter combined with the prefix, or <see langword="null"/> when no
+    /// <c>[Query]</c> attribute was present. An explicitly supplied empty delimiter is preserved as
+    /// <see cref="string.Empty"/> and must not be confused with the absent case.</param>
     /// <param name="Prefix">The field name prefix, if any.</param>
     /// <param name="Format">The value format, if any.</param>
     /// <param name="CollectionFormatValue">The explicit collection format value, if any.</param>
     /// <param name="SerializeNull">Whether null values are serialized as empty fields.</param>
     /// <param name="TreatAsString">Whether the raw value is stringified via <c>ToString()</c> before formatting.</param>
     internal readonly record struct QueryFormData(
-        string Delimiter,
+        string? Delimiter,
         string? Prefix,
         string? Format,
         int? CollectionFormatValue,
--- a/src/InterfaceStubGenerator.Shared/Parser.Request.Path.cs
+++ b/src/InterfaceStubGenerator.Shared/Parser.Request.Path.cs
@@ -104,7 +104,7 @@ internal static partial class Parser
             ElementCanBeNull: false,
             BuildValueFormat(flattenType, null, formattableSymbol, context),
             residual.ToImmutableEquatableArray(),
-            NestingDelimiter: string.IsNullOrEmpty(data.Delimiter) ? "." : data.Delimiter);
+            NestingDelimiter: data.Delimiter ?? DefaultNestingDelimiter);
         return true;
     }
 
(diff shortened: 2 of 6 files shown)


rmja:
Worth flagging since it's a judgement call and not purely mechanical: this makes the fix touch a test that was written deliberately. The argument is parity — for the Indexed path the generator was the only one inserting the dot, and TryBuildIndexedCollectionModel requires data.CollectionFormatValue.HasValue, which requires a [Query] attribute, which always carries a non-null Delimiter. So the ?? DefaultNestingDelimiter fallback on that path is now purely defensive and unreachable in practice.

rmja:
Great. It seems CI is finally happy. Let me know what you think!

glennawatson (approved):

glennawatson:
Looks good. Thanks for the PR :)

